PR #13, opened 2022-10-14
Add support for Nobara Linux

K-Finlay:
Nobara is a Linux distribution derived from Fedora.
This PR adds support for updating Nobara packages via DNF.


DottoDev:
Sorry that it's not in the official Readme, but could you redo the PR so that it targets the dev branch?

PR #33, opened 2022-10-18
Fix broken fisher update command

tennox:
It's just `fisher` (no `update`)
```
fisher: unknown flag or command "update"
usage: fisher add <package...>     Add packages
       fisher rm  <package...>     Remove packages
       fisher                      Update all packages
       fisher ls  [<regex>]        List installed packages matching <regex>
       fisher --help               Show this help
       fisher --version            Show the current version
       fisher self-update          Update to the latest version
       fisher self-uninstall       Uninstall from your system
```

## Standards checklist:

- [x] The PR title is descriptive.
- [ ] *Optional:* I have tested the code myself
    - [ ] I also tested that Topgrade skips the step where needed

PR #37, opened 2022-10-19
Revert fix for older fisher version

tennox:
Reverts topgrade-rs/topgrade#33

Terribly sorry. I just had an old fisher version myself (v3.2.10) :face_with_spiral_eyes:


DottoDev:
Because of backward compatibility with old fisher versions, could you write a check for the version so that topgrade runs the right command?

tennox:
Yes, will do so :)

PR #70, opened 2022-10-25
Add(CI): For automatically publishing on new versions

kunzaatko:
## Standards checklist:

# Reason

In `Cargo.toml` the version is 10.0.0 but the version on `crates.io` is only 9. There is no reason
to manage publishes ourselves. No more need to think when a publish is necessary.

### DIFF
--- a/.github/workflows/crates-publish.yml
+++ b/.github/workflows/crates-publish.yml
@@ -1,3 +1,7 @@
+on:
+  push:
+    branches:
+      - main
 steps:
     - uses: actions/checkout@v2
     - uses: actions-rs/toolchain@v1
--- a/.github/workflows/crates-publish.yml
+++ b/.github/workflows/crates-publish.yml
@@ -1,21 +1,34 @@
 on:
-  push:
-    branches:
-      - main
+  release:
+    types: [published, edited]
 
 name: Publish on release
 
 jobs:
+  prepare:
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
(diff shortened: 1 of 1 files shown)


DottoDev:
Thank you for that.
But it would require an auto version bump or just a run at every release.

kunzaatko:
Well but that is the point, right? That it would no longer happen that there is an older version on crates... It should detect a version bump automatically and run the publish... Isn't that the desired outcome to have the platforms synchronized?

DottoDev:
Understand you.
Would it be possible for you to add a semver check to the CI Pipeline, maybe with automatic version bumps, for example with [rust-lang/rust-semverver](https://github.com/rust-lang/rust-semverver)

kunzaatko:
OK, I could actually add this with:
```yaml
on:
  push:
    paths:
      - 'Cargo.toml'
```
Which is not quite the same but would eliminate most pushes by itself...

kunzaatko:
Or even better maybe with
``` yaml
on:
  release:
    - published
    - edited
```

kunzaatko:
I am not sure if this will run... Whether the `cargo@v1` action makes cargo accessible through `run:` steps... I am not aware of a way to check it without releasing though... What do you think?

kunzaatko:
OK, I checked the and this is written in the `README` of `cargo@v1`. It should be OK I think

DottoDev:
Thank you for your work, looks good for me but I'm still trying to learn how github actions work so I just can review it for clearly visible errors xD.

DottoDev (approved):

PR #71, opened 2022-10-25
Don't panic on git repository with no remote

Osveron:
Current code panics when update of git repositories is run on repository with no remotes. This is a regression I noticed after recent update form crates.io.
Looking around I found that the issue was introduced in #60, simply reverting the offending line fixed the issue for me. I don't get any clippy warnings for this line after changing it back (maybe there was a false positive?).

## Standards checklist:

### DIFF
--- a/src/steps/git.rs
+++ b/src/steps/git.rs
@@ -231,8 +231,7 @@ impl Git {
         let results = basic_rt.block_on(async { stream_of_futures.collect::<Vec<Result<()>>>().await });
 
         let error = results.into_iter().find(|r| r.is_err());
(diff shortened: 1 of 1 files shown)


DottoDev:
This looks pretty interesting because of origin of this bug.

@underyx could you also try out this branch to test if this also solves your issue?

DottoDev:
Took me a while to figure out what this code does. Seem good by me but someone else should maybe try this out too. Maybe @underyx because this might also solve his error.

DottoDev (approved):

PR #76, opened 2022-10-26
feat: added distrobox

LeSnake04:
## Standards checklist:

- [x] The code passes clippy (`cargo clippy`) (The warning  has nothing to do with my additions )
- [x] *Optional:* I have tested the code myself (On fedora 36, only for linux so no mac or windows tester needed)

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -83,6 +83,7 @@ pub enum Step {
     Containers,
     CustomCommands,
     DebGet,
+    Distrobox,
     Deno,
     Dotnet,
     Emacs,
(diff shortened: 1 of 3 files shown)


LeSnake04:
@DottoDev please review

LeSnake04:
I dont see how tests could be implemented for this.

DottoDev (approved):

DottoDev (approved):
Looks good, but it would be nice if there is a possibility to add custom args, if there are any, through a config setting like fwe implemented for other package managers.

DottoDev:
Could you add that too in another PR?

LeSnake04:
Sure.
I would use a `Option<Vec<String>>` Instead since the command needs at least one container name (by default none, meaning -a).

Should I add them to the example config as well?

DottoDev:
> Should I add them to the example config as well?

Yes please

PR #82, opened 2022-10-27
Enhance the README

paescuj:
- Center logo and demo and make them a bit smaller so visitors have to scroll down a bit less
- Rename "Fork" section to "Introduction" and use note box to separate info from actual content
- Adjust links to packages
- Some formatting improvements

Also I think it would be great to be consistent in the casing of "topgrade" (capital letter vs lower case letter), for example:
- https://github.com/topgrade-rs/topgrade/blob/639c7fd422447342f81ae1d25bce647a99a29b48/README.md?plain=1#L12
- https://github.com/topgrade-rs/topgrade/blob/639c7fd422447342f81ae1d25bce647a99a29b48/README.md?plain=1#L34

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,58 +1,76 @@
-![Topgrade](doc/topgrade.png)
-<!---
-![GitHub release](https://img.shields.io/github/release/r-darwish/topgrade.svg)
-[![Crates.io](https://img.shields.io/crates/v/topgrade.svg)](https://crates.io/crates/topgrade)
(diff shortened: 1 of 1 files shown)


DottoDev:
Looks great, thanks

paescuj:
Thank you!

paescuj:
> Also I think it would be great to be consistent in the casing of "topgrade" (capital letter vs lower case letter), for example:

@DottoDev What do you think about this? 😃

DottoDev:
Consistency would be great, but I don't really know if topgrade or Topgrade should be used. But probably would be topgrade the better option.

PR #90, opened 2022-10-30
Add `fundle` step

9999years:
https://github.com/danhper/fundle

Tested on my machine :)

### DIFF
--- a/src/steps/os/unix.rs
+++ b/src/steps/os/unix.rs
@@ -169,7 +169,7 @@ pub fn run_fundle(ctx: &ExecutionContext) -> Result<()> {
     print_separator("fundle");
 
     ctx.run_type()
-        .execute(&fish)
+        .execute(fish)
         .args(["-c", "fundle self-update && fundle update"])
         .check_run()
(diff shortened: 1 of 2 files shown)


src/steps/os/unix.rs
```diff
@@ -157,6 +157,23 @@ pub fn run_fish_plug(ctx: &ExecutionContext) -> Result<()> {
...
+///
+/// See: <https://github.com/danhper/fundle>
+pub fn run_fundle(ctx: &ExecutionContext) -> Result<()> {
+    let fish = require("fish")?;
+    ctx.base_dirs().home_dir().join(".config/fish/fundle").require()?;
+
+    print_separator("fundle");
+
+    ctx.run_type()
+        .execute(&fish)
```
DottoDev, line 172: Use `.execute(fish)` instead of `.execute(&fish)`
Clippy should show that up/correct that normally.

DottoDev (changes requested):

DottoDev (approved):

PR #91, opened 2022-10-30
Add `rcm` step

9999years:
https://github.com/thoughtbot/rcm

Tested on my machine with `cargo run -- --only rcm`.

Should I add an `#[serde(alias = "rcup")]` on the `Step` item? (The tool suite is called `rcm`, and `rcup` is the one that actually updates the dotfiles.)

### DIFF
--- a/src/steps/os/unix.rs
+++ b/src/steps/os/unix.rs
@@ -464,7 +464,7 @@ pub fn run_rcm(ctx: &ExecutionContext) -> Result<()> {
     let rcup = require("rcup")?;
 
     print_separator("rcm");
-    ctx.run_type().execute(&rcup).arg("-v").check_run()
+    ctx.run_type().execute(rcup).arg("-v").check_run()
 }
 
 pub fn reboot() {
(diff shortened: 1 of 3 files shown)


DottoDev (approved):

src/steps/os/unix.rs
```diff
@@ -457,6 +457,16 @@ pub fn run_bun(ctx: &ExecutionContext) -> Result<()> {
...
 }
 
+/// Update dotfiles with `rcm(7)`.
+///
+/// See: <https://github.com/thoughtbot/rcm>
+pub fn run_rcm(ctx: &ExecutionContext) -> Result<()> {
+    let rcup = require("rcup")?;
+
+    print_separator("rcm");
+    ctx.run_type().execute(&rcup).arg("-v").check_run()
```
DottoDev, line 467: use `rcup` instead of `&rcup`

DottoDev (changes requested):

DottoDev (approved):

PR #124, opened 2022-11-03
fix: update fedora

0xMRTT:
Fedora 36 instead of Fedora 29


pan93412:
I found that such this change is unnecessary – this is for test, and won't be a part of the upgrade process.

0xMRTT:
So, should i upgrade all distro release or not ?

pan93412:
> So, should i upgrade all distro release or not ?

As this is for test, I think this part is not really important to waste time on updating them. However, I think the current PRs can leave as it is – we don't need to close them.

DottoDev:
> I found that such this change is unnecessary – this is for integration test, and won't be a part of the upgrade process.

I'm still wondering if our integration tests are even useful.

pan93412:
> > I found that such this change is unnecessary – this is for integration test, and won't be a part of the upgrade process.
> 
> I'm still wondering if our integration tests are even useful.

How about disabling `codecov` first? The notifications and warnings are really annoying, and the codebase has just a few of tests.

0xMRTT:
Yes, codecov is very annoying

DottoDev:
Is disabled
Will be enabled when their are tests that work

PR #127, opened 2022-11-03
Changes version requirements to latest path of minor version

DottoDev:
## Standards checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -225,16 +225,16 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "3.2.22"
+version = "3.1.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "86447ad904c7fb335a790c9d7fe3d0d971dc523b8ccd1561a520de9a85302750"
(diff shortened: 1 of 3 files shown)


GideonBear:
Hey @s34m, what was the reasoning for this? I want to either revert this, or enforce this consistently.

s34m:
The reason was simply to have the latest patch version of the used minor version while not having to manually update every time a new patch version is released.

GideonBear:
> The reason was simply to have the latest patch version of the used minor version while not having to manually update every time a new patch version is released.

For clarity: `1.2.3` does not main the same as `=1.2.3`, but [the same as](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#default-requirements) `^1.2.3`, which is semver-versioning (e.g. `0.1.0` -> `0.1.1`, `1.0.0` -> `1.1.0`), while tilde-versioning only allows patch updates (`0.1.0` -> `0.1.1`, `1.0.0` -> `1.0.1`)

My question is why you chose tilde-versioning over caret-versioning (the default)? Or were you unaware that caret-versioning was the default? (If you want to do the least manual updating, caret-versioning makes more sense)

s34m:
That's true but the intend was to still keep the same minor version, and only update the patch vetsion, to have the least amount of maintenance, which made sense at this point in time for me.

PR #142, opened 2022-11-04
Create SECURITY.md

0xMRTT:
# Support

What's our support policy, i thought about only supporting latest major version

# Security

What's our security policy ? To which guy, people should report vulnerability ?

---

cc @DottoDev


DottoDev:
Currently just last minor version is supported. But if we get a good CI/CD pipeline going through Git Flow we could make it possible to include bug fixes also into older versions.

0xMRTT:
And what about Security policy ?

DottoDev:
I would say they should mail someone in the Org, probably it will be me, if they might have found a security related bug/error. Could you write a short paragraph about that.
Possible footer: 
mailto: Thomas Schönauer <t.schoenauer@hgs-wt.at>

DottoDev (approved):

Issue #163, opened 2022-11-06
Paru issue and Distrobox again ! 


TechXero:
## What did you expect to happen?

For it to work....

## What actually happened?

Failed for Paru and Distrobox.. 

## Additional Details
- XeroLinux an Arch based disro 
- Installed from AUR
- Topgrade 10.1.2

<details>
<pre>
― 18:02:55 - Distrobox ――――――――――――――――――――――――――――――――――――――――――――――――――――――――
 2022-11-06T16:02:55.916Z DEBUG topgrade::executor   > Running "/usr/bin/distrobox" "upgrade" "--all"
Missing dependency: we need a container manager.
Please install one of podman or docker.
You can follow the documentation on:
        man distrobox-compatibility
or:
        https://github.com/89luca89/distrobox/blob/main/docs/compatibility.md

An error occurred
 2022-11-06T16:02:55.923Z DEBUG topgrade::runner     > Step "distrobox" failed: exit status: 127
 2022-11-06T16:02:55.923Z DEBUG topgrade::terminal   > Desktop notification: distrobox failed

―― 18:02:46 - System update ――――――――――――――――――――――――――――――――――――――――――――――――――――
 2022-11-06T16:02:46.641Z DEBUG topgrade::utils  > Cannot find "powerpill"
 2022-11-06T16:02:46.641Z DEBUG topgrade::utils  > Detected "/usr/bin/paru" as "paru"
/usr/bin/paru: error while loading shared libraries: libssl.so.1.1: cannot open shared object file: No such file or directory
 2022-11-06T16:02:46.641Z DEBUG topgrade::executor > Running "/usr/bin/paru" "--pacman" "pacman" "-Syu" "--nodevel" "--noconfirm"
/usr/bin/paru: error while loading shared libraries: libssl.so.1.1: cannot open shared object file: No such file or directory
 2022-11-06T16:02:46.641Z DEBUG topgrade::runner   > Step "System update" failed: exit status: 127
 2022-11-06T16:02:46.641Z DEBUG topgrade::terminal > Desktop notification: System update failed
</pre>
</details>

DottoDev:
Paru at least looks more like a dependency error of paru and not like a topgrade error. Distrobox on the other side start to annoy me because it seems as its impossible to just work.

TechXero:
Remove Distrobox it's becoming too annoying.. I agree. Pru just got an update will build now...

DottoDev:
[...]

PR #164, opened 2022-11-07
Add unstable feature with distrobox support

DottoDev:
## Standards checklist:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,3 +65,4 @@ lto = true
 [features]
 default = []
 self-update = ["self_update_crate"]
+unstable = []
--- a/src/config.rs
+++ b/src/config.rs
@@ -83,7 +83,10 @@ pub enum Step {
     Containers,
     CustomCommands,
     DebGet,
+
+    #[cfg(feature = "unstable")]
     Distrobox,
+
     Deno,
     Dotnet,
     Emacs,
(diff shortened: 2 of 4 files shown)


9999years:
Any reason to gate this behind a feature flag?

DottoDev:
Yes, distrobox current makes too many problems, I probably missed something in the implementation but I can't find it. So after discussion in #163 I put it behind a feature flag for now. Also I would love to put new "untested" features behind it just to not break the widely distributed versions.

pan93412 (commented):
It may be better to separate 'distrobox' as a feature, and makes 'unstable' the meta feature. For example:

```toml
distrobox = []
unstable = ["distrobox"]
```

Thus, users are able to customize which unstable features they want to enable.

DottoDev:
Good point, this makes more sense

LeSnake04:
Now that #163 is resolved, is there any other problem with distrobox? 

I will try my best to adress it.

LeSnake04:
Alternatively make a --unstable flag that activates untested features. 

Then ask users in the release notes to enable the flag for testing and to report issues when unstable steps added. 

When no issues reportet for a feature enable it globally

dylanmtaylor:
I don't think this is necessary - seems to have no issues on master.

DottoDev:
Because now it seems like that it works for everyone this is outdated.

Issue #167, opened 2022-11-10
missing separators for some steps


arctic-penguin:
For example:
* [npm](https://github.com/topgrade-rs/topgrade/blob/6a6a84b0c511158841a731dda79e4aa4fb0689cd/src/steps/node.rs#L218..L222)
* [pnpm](https://github.com/topgrade-rs/topgrade/blob/6a6a84b0c511158841a731dda79e4aa4fb0689cd/src/steps/node.rs#L232..L236)

When these steps are executed, there isn't necessarily an indication something gets executed. In my case the previous step was `pip`, which ran successfully, then topgrade asked me if I wanted to try again, and the desktop notification indicated that the step `pnpm` had failed.

## Additional Details
- Which operating system or Linux distribution are you using? Arch
- How did you install Topgrade? `git clone; cargo run`
- Which version are you running? HEAD

9999years:
Hmm, the npm and pnpm steps call `NPM::upgrade`, which does print a separator:

https://github.com/topgrade-rs/topgrade/blob/1b49b70833acf4f9d3d6e80d921eb90e6d2d1ae1/src/steps/node.rs#L99

<details>
<summary>Example output on my machine (on the <code>dev</code> branch), showing the separators</summary>

```ShellSession
$ cargo run -- --only node
   Compiling topgrade v10.1.2 (/Users/wiggles/topgrade)
    Finished dev [unoptimized + debuginfo] target(s) in 3.31s
     Running `target/debug/topgrade --only node`

―― 10:18:39 - Node Package Manager ―――――――――――――――――――――――――――――――――――――――――――――

removed 13 packages, changed 75 packages, and audited 224 packages in 2s

14 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities

―― 10:18:41 - Yarn Package Manager ―――――――――――――――――――――――――――――――――――――――――――――
yarn global v1.22.19
[1/4] 🔍  Resolving packages...
warning create-react-app > tar-pack > tar@2.2.2: This version of tar is no longer supported, and will not receive security updates. Please upgrade asap.
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
[4/4] 🔨  Rebuilding all packages...

success Saved lockfile.
success Saved 54 new dependencies.
info Direct dependencies
└─ create-react-app@5.0.1
info All dependencies
├─ ansi-styles@4.3.0
...
└─ yallist@4.0.0
[...]

PR #171, opened 2022-11-11
Print errors when steps fail

9999years:
Here's what the error encountered in #167 looks like with this change:

```
pnpm failed: Command failed: `/usr/local/bin/pnpm root -g`

Stdout:
ERROR  Unable to find the global bin directory
Run "pnpm setup" to create it automatically, or set the global-bin-dir setting, or the PNPM_HOME env variable. The global bin directory should be in the PATH.

Caused by:
    `/usr/local/bin/pnpm` failed: exit status: 1
```

Note that:
- It tells you what step failed (`pnpm`)
- It tells you what command was run and which arguments were passed to it
- It shows you the output (stdout and stderr) of the program called (unless the output was already displayed to the user)
- It shows the exit status of the command

### DIFF
--- a/src/runner.rs
+++ b/src/runner.rs
@@ -2,6 +2,7 @@ use crate::ctrlc;
 use crate::error::{DryRun, SkipStep};
 use crate::execution_context::ExecutionContext;
 use crate::report::{Report, StepResult};
+use crate::terminal::print_error;
 use crate::{config::Step, terminal::should_retry};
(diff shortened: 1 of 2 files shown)


9999years:
Also it's wild that `pnpm` is just by-default broken after install. Maybe we should detect and ignore this particular error?

DottoDev (approved):

PR #173, opened 2022-11-11
Use `color_eyre`

9999years:
This uses the `color_eyre` library, a fork of `anyhow` with improved console output: https://docs.rs/color-eyre/latest/color_eyre/index.html

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -4,9 +4,9 @@ use std::env;
 use std::io;
 use std::process::exit;
 
-use anyhow::Context;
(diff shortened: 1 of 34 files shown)


src/main.rs
```diff
@@ -36,9 +36,10 @@ mod terminal;
 mod utils;
 
 fn run() -> Result<()> {
+    color_eyre::install()?;
```
9999years, line 39: This sets a panic handler and makes `eyre::Report` values prettier

DottoDev (approved):

PR #174, opened 2022-11-11
Use `tracing`

9999years:
`tracing` is the standard logging library, used by rustc. Let's use it. In a future PR we can start leaning on `tracing`'s tools, annotating functions, and so on.

Might tweak the output format a bit more.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -442,7 +442,7 @@ pub struct CommandLineArgs {
     #[clap(long = "env", value_name = "NAME=VALUE", multiple_values = true)]
     env: Vec<String>,
 
-    /// Output debug logs. Alias for `--trace debug`.
+    /// Output debug logs. Alias for `--log-filter debug`.
     #[clap(short = 'v', long = "verbose")]
     pub verbose: bool,
 
@@ -485,7 +485,7 @@ pub struct CommandLineArgs {
     ///
     /// See: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/struct.EnvFilter.html
     #[clap(long, default_value = "info")]
-    pub trace: String,
+    pub log_filter: String,
 }
 
 impl CommandLineArgs {
@@ -505,7 +505,7 @@ impl CommandLineArgs {
         if self.verbose {
             "debug".into()
         } else {
-            self.trace.clone()
+            self.log_filter.clone()
         }
     }
 }
--- a/src/config.rs
+++ b/src/config.rs
@@ -10,11 +10,11 @@ use color_eyre::eyre;
 use color_eyre::eyre::Context;
 use color_eyre::eyre::Result;
 use directories::BaseDirs;
-use log::debug;
 use regex::Regex;
 use serde::Deserialize;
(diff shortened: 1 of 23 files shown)


src/config.rs
```diff
@@ -480,6 +480,12 @@ pub struct CommandLineArgs {
     /// Show the reason for skipped steps
     #[clap(long = "show-skipped")]
     show_skipped: bool,
+
+    /// Tracing filter directives.
+    ///
+    /// See: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/struct.EnvFilter.html
+    #[clap(long, default_value = "info")]
+    pub trace: String,
```
9999years, line 488: New CLI option. Should this be documented somewhere?

DottoDev:
I'll review it.

src/config.rs
Larandar: IMO rather than having a different option, it is more user friendly to change verbose to a count u8 (and maybe add a -q --quiet option) and implement tracing filter matching on that: -vv = trace -v = debug (-q = warning -qq error).

This is a rather common pattern in CLI design

src/terminal.rs
```diff
@@ -105,7 +105,7 @@ impl Terminal {
                     command.args(["-a", "Topgrade", "Topgrade"]);
                     command.arg(message.as_ref());
                     if let Err(err) = command.output_checked() {
-                        log::error!("{err:?}");
+                        tracing::error!("{err:?}");
```
Larandar, line 108: Couln't this one be `error!(` only since line 17 already use `tracing::error`?

This comment also apply to other times `log::` was called directly, it would be more consistent to have them all changed to unqualified call?

Larandar (commented):
Just some thoughts :)

src/config.rs
9999years: Those are not the same feature; while this can be used to limit logging based on level (`info`, `trace`, `debug`, `warn`, etc.), it can also do a lot more complex filtering; you can restrict log messages to various levels by module name, by extra data present, and more. See the linked docs:

https://docs.rs/tracing-subscriber/latest/tracing_subscriber/struct.EnvFilter.html

This sort of fine-grained filtering is super helpful when debugging, and I'm not sure I want to disable it in favor of only being able to select the level.

src/terminal.rs
9999years: A lot of these changes were find/replace'd, so they may not be super consistent.

src/config.rs
Larandar: I see your point, it's indeed useful for debugging, but should a debug option be exposed to the CLI? 

Since the EnvFilter support the common `RUST_LOG` variable, I feel like having the CLI having level only filtering via standard and self explanatory options, but document that `RUST_LOG` environment variable is available for more advanced filtering would be suffisant.

Else the name `--trace` is not really intuitive, if you want to keep it then `--log-filter` might be more explanatory?

DottoDev (approved):

PR #199, opened 2022-11-17
Npm rework

DottoDev:
## Standards checklist:

### DIFF
--- a/src/steps/node.rs
+++ b/src/steps/node.rs
@@ -4,6 +4,7 @@ use std::os::unix::fs::MetadataExt;
 use std::path::PathBuf;
 use std::process::Command;
 
+use crate::utils::require_option;
 use color_eyre::eyre::Result;
 #[cfg(target_os = "linux")]
 use nix::unistd::Uid;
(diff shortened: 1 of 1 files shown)


src/steps/node.rs
```diff
@@ -91,11 +93,9 @@ impl NPM {
     fn upgrade(&self, run_type: RunType, use_sudo: bool) -> Result<()> {
         let args = ["update", self.global_location_arg()];
         if use_sudo {
-            run_type
-                .execute("sudo")
-                .arg(&self.command)
-                .args(args)
-                .status_checked()?;
+            let sudo_option = sudo();
+            let sudo = require_option(sudo_option, String::from("sudo is not installed"))?;
```
9999years, line 97: We should see about making `require_option` take an `AsRef<str>` or `Display` or something.

9999years (approved):
lgtm

src/steps/node.rs
DottoDev: Good idea for the future.

PR #203, opened 2022-11-19
Add `gup`

### LINKED ISSUE #195: Golang manually installed binaries update
Update all golang binaries that are manually installed
https://github.com/nao1215/gup
Windows, Mac,Linux
It should be invoked every time like apt
gup update

9999years:
Fixes #195

I also want to keep splitting stuff out of the `generic` and `unix` modules into shorter, more specific modules (as I've done here with the `go` module).


DottoDev (approved):

DottoDev:
Lgtm, and also, I think the splitting up into multiple files should have been done way earlier. Appreciate the idea.

PR #256, opened 2022-12-08
Add `helm` (#255)

### LINKED ISSUE #255: Add helm update step

[Helm](https://helm.sh/) is a Kubernetes package manager.
It consists of charts and repositories, which should be regularly updated. 
Hence adding an update step would be beneficial.

Repository: https://github.com/helm/helm

Helm supports Linux, macOS, and Windows.

### What should Topgrade do to determine if the tool needs to be invoked?

Check if a Helm executable is in the current path.

```bash
helm repo update
```

ThomasK33:
Fixes #255 

Signed-off-by: Thomas Kosiewski <thoma471@googlemail.com>

## Standards checklist:


src/config.rs
```diff
@@ -110,6 +110,7 @@ pub enum Step {
     Juliaup,
     Kakoune,
     Krew,
+    Helm,
```
DottoDev, line 113: This should be higher up due to alphabetic ordering.

DottoDev (changes requested):

src/config.rs
ThomasK33: Sure, I'll fix that.
I just assumed to group it with krew/another Kubernetes tool.

ThomasK33: I've resorted the entire enum alphabetically.

DottoDev (approved):

PR #279, opened 2022-12-21
Use standard dashes instead of long dashes.

Crocmagnon:
Many fonts have ligatures for standard dashes, long dashes are less common.

### Fira Code
Before:
CleanShot 2022-12-21 at 20 39 55@2x

After:
CleanShot 2022-12-21 at 20 39 37@2x

### MonoLisa
Before:
CleanShot 2022-12-21 at 20 40 56@2x

After:
CleanShot 2022-12-21 at 20 41 10@2x

### Menlo
Before:
CleanShot 2022-12-21 at 20 43 29@2x

After:
CleanShot 2022-12-21 at 20 43 38@2x

### JetBrains Mono
Before:
CleanShot 2022-12-21 at 20 44 30@2x

After:
CleanShot 2022-12-21 at 20 44 38@2x

## Standards checklist:

### DIFF
--- a/src/terminal.rs
+++ b/src/terminal.rs
@@ -143,7 +143,7 @@ impl Terminal {
                     .write_fmt(format_args!(
                         "{}\n",
                         style(format_args!(
-                            "\n―― {} {:―^border$}",
+                            "\n-- {} {:-^border$}",
                             message,
                             "",
                             border = max(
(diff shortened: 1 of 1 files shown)


Crocmagnon:
Hm after seeing the examples with different fonts, I'm considering instead raising an issue with Fira Code as it looks like I was wrong in assuming that it would look better with standard dashes.

PR #316, opened 2023-01-17
Adds pip-review and pipupgrade support

### LINKED ISSUE #304: Pip modules update
Topgrade currently updates pip itself. My suggestion is to add an optional feature to update installed modules, through `pip-review`.
pip-review https://github.com/jgonggrijp/pip-review

All where python/pip works - both tools are pip modules

Same as pip.
`pipupgrade` should be explicited declared on command-line or `topgrade.toml`
**Maybe** add a message just like `winget` is "disabled by default" when running on Windows.

`pip-review -a -C`

```
  --auto, -a            Automatically install every update found
  --continue-on-fail, -C
                        Continue with other installs when one fails

```

DottoDev:
Closes #304 
## Standards checklist:

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -83,8 +83,6 @@
 #aura_aur_arguments = "-kx"
 #aura_pacman_arguments = ""
 #garuda_update_arguments = ""
-#enable_pip_review = true   ###disabled by default
-#enable_pipupgrade = true   ###disabled by default
 #show_arch_news = true
 #trizen_arguments = "--devel"
 #pikaur_arguments = ""
@@ -94,6 +92,10 @@
 #emerge_update_flags = "-uDNa --with-bdeps=y world"
 #redhat_distro_sync = false
 #rpm_ostree = false
+
+[python]
+#enable_pip_review = true   ###disabled by default
+#enable_pipupgrade = true   ###disabled by default
 
 [windows]
 # Manually select Windows updates
--- a/config.example.toml
+++ b/config.example.toml
@@ -83,6 +83,8 @@
 #aura_aur_arguments = "-kx"
 #aura_pacman_arguments = ""
 #garuda_update_arguments = ""
+#enable_pip_review = true   ###disabled by default
+#enable_pipupgrade = true   ###disabled by default
 #show_arch_news = true
 #trizen_arguments = "--devel"
 #pikaur_arguments = ""
--- a/src/config.rs
+++ b/src/config.rs
@@ -189,14 +189,9 @@ pub struct Windows {
 
 #[derive(Deserialize, Default, Debug)]
 #[serde(deny_unknown_fields)]
-pub struct PipReview {
-    enable: Option<bool>,
-}
-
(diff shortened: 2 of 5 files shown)


luciodaou (changes requested):
It's failing, when I try to run after building and uncommenting the enable_pip_review flag inside linux section:

```
lucio in topgrade-dev/target/debug
➜  ./topgrade --config ../../config.example.toml --only pip_review
ERROR Failed to deserialize ../../config.example.toml
ERROR failed to load configuration: unknown field `enable_pip_review`, expected one of `yay_arguments`, `aura_aur_arguments`, `aura_pacman_arguments`, `arch_package_manager`, `show_arch_news`, `garuda_update_arguments`, `trizen_arguments`, `pikaur_arguments`, `pamac_arguments`, `dnf_arguments`, `apt_arguments`, `enable_tlmgr`, `redhat_distro_sync`, `rpm_ostree`, `emerge_sync_flags`, `emerge_update_flags` for key `linux` at line 120 column 1
```
Update:
On the config file, I tried moving the flag outside of Linux section, as the only flag in the file, failed too:
```
lucio in topgrade-dev/target/debug
➜  ./topgrade --config ../../config.example.toml --only pip_review
ERROR Failed to deserialize ../../config.example.toml
ERROR failed to load configuration: invalid type: string "PipReview", expected struct PipReview for key `enable_pip_review` at line 1 column 20
```

Trying to use `"PipReview"` or `PipReview` instead of `true` didn't work either.

luciodaou:
I tried some things, like changing the flags to bool, but it gave me errors way too complex for me. Rust syntax ain't easy like Python... 🤣

luciodaou:
@DottoDev , managed to make it work after some trial and error. The problem is that the code is looking for this on the config file:
```
[enable_pip_review]
enable = true
```

The example file and error messages says different things about it. The example config.toml file says `enable_pip_review = true` should be inside `[Linux]` section, but that gives the errors I mentioned on my first reply (https://github.com/topgrade-rs/topgrade/pull/316#pullrequestreview-1252671696).

luciodaou:
Also, after making it work, I got a build error on one of the modules, but topgrade detected it as running OK:

image

DottoDev:
Damm it I hate it when tools use different exit codes 😂

DottoDev:
@luciodaou Sorry for the late response, could you check that out

luciodaou (approved):
Working!

image

luciodaou (approved):

luciodaou:
On one of my WSL distros, I have both pip-review and pipupgrade installed, so I don't need both on topgrade.

So I configured my `topgrade.toml` file like this:
```
[python]
enable_pip_review = true   ###disabled by default
enable_pipupgrade = false   ###disabled by default
```

Yet, it still enables pipupgrade:
image

If I comment out the pipupgrade line, it still runs:
image

Disabling both by comment:
image

Disabling both by the flags:
image

It seems that pipupgrade is piggybacking on pip-review flag.

PR #317, opened 2023-01-17
Fixes gem update --system requires sudo now

DottoDev:
## Standards checklist:

### DIFF
--- a/src/steps/generic.rs
+++ b/src/steps/generic.rs
@@ -16,7 +16,7 @@ use crate::command::{CommandExt, Utf8Output};
 use crate::execution_context::ExecutionContext;
 use crate::executor::{ExecutorOutput, RunType};
 use crate::terminal::{print_separator, shell};
-use crate::utils::{self, require_option, PathExt};
+use crate::utils::{self, require, require_option, which, PathExt};
 use crate::{
     error::{SkipStep, StepFailed, TopgradeError},
     terminal::print_warning,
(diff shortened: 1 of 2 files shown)


src/steps/generic.rs
```diff
@@ -83,17 +83,23 @@ pub fn run_gem(base_dirs: &BaseDirs, run_type: RunType) -> Result<()> {
...
+    ctx.base_dirs().home_dir().join(".gem").require()?;
 
     print_separator("RubyGems");
-
-    if !std::path::Path::new("/usr/lib/ruby/vendor_ruby/rubygems/defaults/operating_system.rb").exists() {
-        run_type.execute(gem).args(["update", "--system"]).status_checked()
+    if let Some(sudo) = &ctx.sudo() {
+        if !std::path::Path::new("/usr/lib/ruby/vendor_ruby/rubygems/defaults/operating_system.rb").exists() {
+            ctx.run_type()
+                .execute(sudo)
```
mouse07410, line 93: Insufficient for installations behind firewall. Please find a way to validate presence of `gem` executable elsewhere, and replace this with something like
```rust
  ctx.run_type()
     .execute(sudo)
     .args(["-EH", "gem", "update", "--system"])
     .status_checked()?;
```

mouse07410 (changes requested):
Please add `-EH` argument between `sudo` and `gem`.

src/steps/generic.rs
DottoDev: Why should that be insufficient. And also the check is just to get info about if it uses a modified version distributed by example by debian instead of the official version. Else it's just the normal gem command found by which like every other command.

mouse07410: > Why should that be insufficient

Because `sudo` ignores proxy settings in the env variables, unless it is given these flags. And unless internet requests are sent to the proxy, firewall blocks them.

> the check is just to get info about if it uses a modified version distributed by example by debian instead of the official version

I do not object to this check - I simply do not know how to put it there, as my familiarity with Rust is still limited.

Perhaps, something like can work
```rust
ctx.run_type()
     .execute(sudo)
     .arg("-EH")
     .arg(require("gem")?)
     .args(["update", "--system"])
     .status_checked()?;
```

mouse07410 (changes requested):
Please add `.arg("-EH")` as I showed above.

DottoDev:
Back again, sorry, working in it now

PR #347, opened 2023-02-01
run_custom_command: use interactive shell on unix

utkarshgupta137:
## Standards checklist:

Tested on macOS 13.2 (M1 Pro) with homebrew zsh.

### DIFF
--- a/src/steps/generic.rs
+++ b/src/steps/generic.rs
@@ -467,7 +467,10 @@ pub fn run_myrepos_update(base_dirs: &BaseDirs, run_type: RunType) -> Result<()>
 
 pub fn run_custom_command(name: &str, command: &str, ctx: &ExecutionContext) -> Result<()> {
     print_separator(name);
-    ctx.run_type().execute(shell()).arg("-c").arg(command).status_checked()
+    let mut exec = ctx.run_type().execute(shell());
+    #[cfg(unix)]
+    exec.arg("-i");
+    exec.arg("-c").arg(command).status_checked()
 }
 
 pub fn run_composer_update(ctx: &ExecutionContext) -> Result<()> {


DottoDev:
What would be the advantage of running the command in an interactive shell?

utkarshgupta137:
Aliases/functions defined in *rc file won't be available without an interactive shell

DottoDev (approved):

reini-1:
topgrade now stops working on the second custom command by sending it in the background.

```shell
Ubuntu-22.04> reini@ubuntu:~$ topgrade --only custom_commands
Due to a design issue with notify-send it could be that topgrade hangs when it's finished.
If this is the case on your system add the --skip-notify flag to the topgrade command or set skip_notify = true in the config file.
If you don't want this message to appear any longer set display_preamble = false in the config file.
For more information about this issue see https://askubuntu.com/questions/110969/notify-send-ignores-timeout and https://github.com/topgrade-rs/topgrade/issues/288.

── 10:58:13 - Self update ──────────────────────────────────────────────────────
Topgrade is up-to-date

── 10:58:14 - 01 Update downloaded apps ────────────────────────────────────────
2023-03-13 10:58:14: Update as-tree
2023-03-13 10:58:15: Update bat
...
── 10:58:19 - 12 Create/update ~/.ssh/config.known_hosts ───────────────────────

[7]+  Stopped                 topgrade --only custom_commands

Ubuntu-22.04> reini@ubuntu:~$ jobs
[1]   Stopped                 topgrade
[2]   Stopped                 topgrade --only custom_commands
[3]   Stopped                 topgrade --only custom_commands
[4]-  Stopped                 topgrade --only custom_commands
[5]+  Stopped                 topgrade --only custom_commands
```

How can I fix this?

As a workaround I downgraded to the previous version. That works fine.

utkarshgupta137:
I'm not sure this is entirely related to this PR. I also use multiple commands & don't face this issue. I'm using zsh on macOS btw.
Can you share your shell (`echo $SHELL`) & the full custom command? Does the custom command work from your normal shell?

utkarshgupta137:
@DottoDev In hindsight, this was a breaking change and should've been a minor version bump at least. Feel free to revert it, I'll work on a configurable version in the meanwhile.

reini-1:
See also https://github.com/topgrade-rs/topgrade/issues/380

DottoDev:
Will be reverted rn and 1 new feature will be added

PR #364, opened 2023-02-20
pacdef: support new version 1.x

arctic-penguin:
`pacdef` will release a new major version in the near future. Part of the release is a breaking change for the command line arguments.

This commit checks which version of pacdef is installed by parsing `pacdef version`, and uses either the old or the new syntax.

## Standards checklist:

### DIFF
--- a/src/steps/os/linux.rs
+++ b/src/steps/os/linux.rs
@@ -472,14 +472,14 @@ pub fn run_pacdef(ctx: &ExecutionContext) -> Result<()> {
         ctx.run_type()
             .execute(&pacdef)
             .args(["package", "review"])
-            .status_checked()
+            .status_checked()?;
     } else {
         ctx.run_type().execute(&pacdef).arg("sync").status_checked()?;
 
         println!();
-        ctx.run_type().execute(&pacdef).arg("review").status_checked()
+        ctx.run_type().execute(&pacdef).arg("review").status_checked()?;
     }
-
+    Ok(())
 }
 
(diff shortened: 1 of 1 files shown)


src/steps/os/linux.rs
```diff
@@ -458,10 +458,28 @@ pub fn run_pacdef(ctx: &ExecutionContext) -> Result<()> {
...
+        println!();
+        ctx.run_type()
+            .execute(&pacdef)
+            .args(["package", "review"])
+            .status_checked()
+    } else {
+        ctx.run_type().execute(&pacdef).arg("sync").status_checked()?;
+
+        println!();
+        ctx.run_type().execute(&pacdef).arg("review").status_checked()
```
DottoDev, line 480: This line should end with a `?;` and the return value of the function should be `Ok(())` but else it looks good.

DottoDev (changes requested):

DottoDev:
Thanks for bringing this up.

src/steps/os/linux.rs
arctic-penguin: I'm not quite following. Your requested changes will result in semantically identical code.  Both arms of the `if` block end with the implicit return of `Result<()>`. Since there is no semicolon in 481 this is also the return value of the function.

DottoDev: It will be semantically identical code but we use this style in the entire code base, which just makes it a bit easier to maintain. Just a style of code thingy.

DottoDev: Thx

DottoDev (approved):

PR #388, opened 2023-03-20
Add no-self-update config and flag

### LINKED ISSUE #381: Self-update configurable
It should be possible to disable self-update to be able to use a old version if the new version did not work.

## More information

When you install topgrade from Github releases, it will self-update on every run but if there is a problem with the new version, topgrade connot be used anymore until the bug is fixed. See e.g. https://github.com/topgrade-rs/topgrade/issues/380.

So it would be nice to have a option in the configuration file and optional a command line parameter to disable automatic self-update and use a old working version.

PolpOnline:
Closes #381

## Standards checklist:

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -50,6 +50,9 @@
 # Skip the preamble displayed when topgrade is run
 #display_preamble = false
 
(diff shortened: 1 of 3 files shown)


etrigan63:
What version has this? I am running 11.0.2 and it does not recognize `no-self-update`.

DottoDev:
You have to use `no_self_update=true` to disable it

PR #407, opened 2023-04-23
Replace `directories` with `home` & `etcetera`

utkarshgupta137:
In most places, the `BaseDirs` struct is just used for getting the user's home directory.
I've made it a global lazy variable & replaced all instances of `base_dirs.home_dir()` with `HOME_DIR` (commit 1).
After that, I removed all the places where `base_dirs` or `ctx` was no longer required (commit 2).
Finally, I replaced the `directories` crate with the `etcetera` crate (commit 3), since none of the crates use the `~/Library` paths returned for macOS. Also created a global variable for `XDG_DIRS` & `WINDOWS_DIRS` (if compiling for Windows) to remove unnecessary argument passing.

I only have a few applications installed, but it is working for me on macOS. In case you want to be extra cautious, you could release commit 1/2 & commit 3 separately.

## Standards checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2059,6 +2059,7 @@ dependencies = [
  "libc",
  "nix 0.24.3",
  "notify-rust",
+ "once_cell",
  "parselnk",
  "regex",
  "rust-ini",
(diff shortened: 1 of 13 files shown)


DottoDev (commented):
Nice work and cleanup

DottoDev (approved):

PR #414, opened 2023-05-01
feat(shell): add `packer.nu`

wizard-28:
## Standards checklist:


wizard-28:
`packer.nu`(https://github.com/Jan9103/packer.nu) needs to get updated before [`nu`](https://www.nushell.sh/) updates. Not sure how to enforce that.

DottoDev:
> > > > [`packer.nu`](https://github.com/Jan9103/packer.nu) needs to get updated before [`nu`](https://www.nushell.sh/) updates. Not sure how to enforce that.
> > > 
> > > 
> > > @DottoDev, any pointers on how to do that would be appreciated.
> > 
> > 
> > This should not be a problem because every package manager updates before `packer.nu` does.
> 
> That's exactly the issue. `packer.nu` will break and will be unable to update itself if `nu` gets updated by package managers before it. As `nu` updates often contain breaking changes.

Oh, misread that. The easiest way would be to put `packer.nu` in front of the other package managers and note in a comment that it has to be one of the first things to update.

DottoDev (approved):

PR #417, opened 2023-05-01
Make `zypper dist-upgrade` opt-in on SLE/Leap

signed-log:
## Standards checklist:

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -101,6 +101,7 @@
 #emerge_sync_flags = "-q"
 #emerge_update_flags = "-uDNa --with-bdeps=y world"
 #redhat_distro_sync = false
+#suse_dup = false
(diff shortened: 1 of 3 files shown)


DottoDev:
Could you change the flag to `disable_suse_dup` or change it to be enabled by default.

signed-log:
> Could you change the flag to `disable_suse_dup` or change it to be enabled by default.

I don't see why that would be a desired behavior for a SLE/Leap system 

A `distro-upgrade` should be opt-in. topgrade has the same behavior for Red Hat's `distro-sync`

DottoDev (approved):

signed-log:
Thanks ! I hope to also put topgrade on openSUSE/SLE repos as soon as possible

PR #421, opened 2023-05-04
Added ability to include directories as an extension of the config file

### LINKED ISSUE #405: Include directive in topgrade.toml

I want to suggest that topgrade add an include directive in its configuration file, so that configuration can be included from another named file. This feature is supported by other tools, including Bash (`source`), tmux (`source-file`), Vim (`source`), OpenSSH (`Include`), Git (`[include]`) and others.

I find this ability helpful when I synchronize dotfiles between computers, but want one small piece of configuration to be different on a particular computer for some reason. For example, my Git configuration file (_.config/git/config_) uses my personal email address, but also includes this section at the bottom:

```gitconfig
[include]
  # Include local configurations, if any.
  path = config.local
```

On my work computer, the _config.local_ file sets the email address to my work email address.

With topgrade, I would do something similar, using this ability to run certain custom commands on some machines, but not others.

I would be happy to help test this.

PolpOnline:
## Standards checklist:

I have tested it by myself, but I am not sure if I set all the merge behaviours correctly. 
@openjck you may want to test this.

closes #405


src/config.rs
```diff
@@ -390,11 +499,36 @@ impl ConfigFile {
...
+        let mut result: Self = Default::default();
+
+        if let Some(includes) = &config_file_include_only.include {
+            for include in includes.iter().rev() {
+                let include_path = shellexpand::tilde::<&str>(&include.as_ref()).into_owned();
+                let include_path = PathBuf::from(include_path);
+                let include_contents = fs::read_to_string(&include_path).map_err(|e| {
+                    tracing::error!("Unable to read {}", include_path.display());
+                    e
+                })?;
```
Dreaming-Codes, line 516: Wouldn't it be better to simply ignore the file instead of returning an error for the entire read config function?

src/config.rs
```diff
@@ -390,11 +499,36 @@ impl ConfigFile {
...
+
+                result.merge(toml::from_str::<Self>(&include_contents).map_err(|e| {
+                    tracing::error!("Failed to deserialize {}", include_path.display());
+                    e
+                })?);
```
Dreaming-Codes, line 521: Same as above

src/config.rs
```diff
@@ -390,11 +499,36 @@ impl ConfigFile {
...
+
+                debug!("Configuration include found: {}", include_path.display());
+            }
+        }
+
+        result.merge(toml::from_str::<Self>(&contents).map_err(|e| {
+            tracing::error!("Failed to deserialize {}", config_path.display());
+            e
+        })?);
```
Dreaming-Codes, line 530: Same as above

config.example.toml
```diff
@@ -1,3 +1,8 @@
+# Include any additional configuration file(s)
+# Note that this config file overrides the includes
+# and includes that are on the left override the ones on the right in this array
+#include = ["/etc/topgrade.toml"]
```
Dreaming-Codes, line 4: Couldn't you use the glob crate to match an entire directory and create a directory dedicated to topgrade configurations, such as `/etc/topgrade.d/`?

Dreaming-Codes (commented):
Overall, this is a generally good PR, but I would recommend a few minor improvements.

config.example.toml
PolpOnline: The linked issue to this pr asked to include files that are in this specific array, whether the directory they are located.

However I could match any file in `$XDG_CONFIG_HOME/topgrade/*` by default, but I would have to ask maintainers if that's a good idea

PolpOnline:
Blocking this from getting merged because the commands and pre_commands field don't have a merge strategy yet, will surely implement it later

config.example.toml
```diff
@@ -1,3 +1,8 @@
+# Include any additional configuration file(s)
+# Note that this config file overrides the includes
+# and includes that are on the left override the ones on the right in this array
```
openjck, line 3: Thanks for your interest in this issue.

This part caught my attention. If consistency with other tools is important, it might be good to look at how other tools handle this. It's my understanding that Git, Vim, Bash, and others don't have the main configuration file take precedence. Instead, settings at the bottom of those configuration files take precedence over settings at the tops of those files, and included files are treated as though their contents are inlined.

For example, with the following setup, the effective email address would be _included@example.com_:

```gitconfig
// .gitconfig
[user]
  email = main@example.com

[include]
  path = .gitconfig.included
```

```gitconfig
// .gitconfig.included
[user]
  email = included@example.com
```

However, if the include is moved up, the effective email address would be _main@example.com_:

```gitconfig
// .gitconfig
[include]
  path = .gitconfig.included

[user]
  email = main@example.com
```

```gitconfig
// .gitconfig.included
[user]
  email = included@example.com
```

As far as I know, the same principle is true for Bash, Vim, and others. Settings at the bottom override settings at the top, and included files are treated as though their contents are right there.

PolpOnline: This can't be done, not without a breaking change in the way the misc config entries are handled. 
This is because as of now these entries are handled in the global scope, and adding an `[include]` section in between these breaks the following ones, because there is currently no way to specify a return to global scope in TOML files.  

The simplest way to fix this is to add a section named `[misc]` and move all the currently-in-global-scope entries to it.
The problem with this approach is that it breaks all the previously-created config files.

PolpOnline: @DottoDev since you are the maintainer of this repository, I'm asking you to tell me how to proceed.

If we are going to migrate all currently-in-global-scope entries to the new `[misc]` section, I'm available to adapt `topgrade` to auto migrate the config file.

config.example.toml
DottoDev: Matching any file would be cool

config.example.toml
PolpOnline: I've just pushed a commit to resolve this. I've already tested it myself, but you might want to check it out too @openjck.

PolpOnline:
I think the pr might be ready to be merged if you approve it.

config.example.toml
PolpOnline: > Matching any file would be cool

I will push a commit to do that

PolpOnline: I think putting them at the beginning of the file (lowest priority) should be fine.

PolpOnline: Solved in [33d1fc6](https://github.com/topgrade-rs/topgrade/pull/421/commits/33d1fc6e6ea10bdd3ba53b06c9767720f8f95a33)

DottoDev (approved):
lgtm

DottoDev (changes requested):
Cargo fmt and clippy checks need to passbefore a merge can occure.

DottoDev:
Thanks for the work

niStee:
#581 @SteveLauC

PR #427, opened 2023-05-09
feat: support maza

### LINKED ISSUE #332: Add Maza support
Hi,

will it be possible to add support for Maza Ad Blocker?

Link: https://github.com/tanrax/maza-ad-blocking

Description:
Like Pi-hole but local and using your operating system. Simple, native and efficient local ad blocker. Bash script compatible with MacOS, Linux, BSD and Windows Subsystem for Linux (WSL).

Updating is done by simply typing: maza update

SteveLauC:
## Standards checklist:

Closes #332

### DIFF
--- a/src/steps/os/unix.rs
+++ b/src/steps/os/unix.rs
@@ -512,6 +512,13 @@ pub fn run_rcm(ctx: &ExecutionContext) -> Result<()> {
     ctx.run_type().execute(rcup).arg("-v").status_checked()
 }
 
+pub fn run_maza(ctx: &ExecutionContext) -> Result<()> {
+    let maza = require("maza")?;
+
+    print_separator("maza");
+    ctx.run_type().execute(maza).arg("update").status_checked()
+}
+
 pub fn reboot() -> Result<()> {
     print!("Rebooting...");
     Command::new("sudo").arg("reboot").status_checked()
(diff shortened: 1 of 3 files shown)


src/steps/os/unix.rs
```diff
@@ -512,6 +512,13 @@ pub fn run_rcm(ctx: &ExecutionContext) -> Result<()> {
     ctx.run_type().execute(rcup).arg("-v").status_checked()
 }
 
+pub fn run_maza(ctx: &ExecutionContext) -> Result<()> {
+    let maza = require("maza")?;
+
+    print_separator("maza");
+    ctx.run_type().execute(maza).arg("update").status_checked()
```
DottoDev, line 519: Return with ? at the eol and then the functions should return a Ok(())

DottoDev (changes requested):

src/steps/os/unix.rs
SteveLauC: You mean something like this?
```rust
    ctx.run_type().execute(maza).arg("update").status_checked()?;
    Ok(())
}
```
Any reason why we should write it in this way?

DottoDev: Could be that I'm mistaken but else if `require()` fails the `Err` doesn't get returned.

SteveLauC: I am a little bit confused, why would the error derived from `require()` won't be returned? It is followed by a question mark `?`:
```rust
let maza = require("maza")?;
```

PR #434, opened 2023-05-22
fix: use env ZSH to compose oh-my-zsh install dir

### LINKED ISSUE #426: [oh-my-zsh]`$ZSH` should be used to compose its install directory
When determining if `oh-my-zsh` is installed, we check the existence of its installation directory, and this directory [can be customized](https://github.com/ohmyzsh/ohmyzsh#custom-directory) through the `$ZSH` environment variable when installing, which means the directory is possibly not `~/.oh-my-zsh`. However, in the current impl of `topgrade`, it is hardcoded to `~/.oh-my-zsh`:
https://github.com/topgrade-rs/topgrade/blob/14753a14e70d2931ec71eebf23f93e7060288e58/src/steps/zsh.rs#L156-L157

We should do something like this:

```rust
    let oh_my_bash = var("ZSH")
        // default to `~/.oh-my-zsh`
        .unwrap_or(
            HOME_DIR
                .join(".oh-my-zsh")
                .to_str()
                .expect("should be UTF-8 encoded")
                .to_string(),
        )
        .require()?;
```

SteveLauC:
## Standards checklist:

Closes #426

cc @lime-desu who has already found this and posted [it](https://github.com/topgrade-rs/topgrade/discussions/361) in discussions

### DIFF
--- a/src/steps/zsh.rs
+++ b/src/steps/zsh.rs
@@ -154,7 +154,11 @@ pub fn run_zim(run_type: RunType) -> Result<()> {
 
 pub fn run_oh_my_zsh(ctx: &ExecutionContext) -> Result<()> {
     require("zsh")?;
-    let oh_my_zsh = HOME_DIR.join(".oh-my-zsh").require()?;
+    let oh_my_zsh = env::var("ZSH")
+        .map(PathBuf::from)
+        // default to `~/.oh-my-zsh`
+        .unwrap_or(HOME_DIR.join(".oh-my-zsh"))
+        .require()?;
 
     print_separator("oh-my-zsh");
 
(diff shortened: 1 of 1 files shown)


src/steps/zsh.rs
```diff
@@ -194,7 +198,10 @@ pub fn run_oh_my_zsh(ctx: &ExecutionContext) -> Result<()> {
 
     ctx.run_type()
         .execute("zsh")
-        .env("ZSH", &oh_my_zsh)
```
SteveLauC, line 197: This line of code was introduced in [this commit](https://github.com/topgrade-rs/topgrade/commit/44b68a048f06febfb14fc592f3271f7800dcdfa3), and I haven't figured out why do we need this. IMHO, it is useless so I removed it.

src/steps/zsh.rs
```diff
@@ -194,7 +198,10 @@ pub fn run_oh_my_zsh(ctx: &ExecutionContext) -> Result<()> {
...
         .arg(&oh_my_zsh.join("tools/upgrade.sh"))
+        // oh-my-zsh returns 80 when it is already updated and no changes pulled
+        // in this update.
+        // See this comment: https://github.com/r-darwish/topgrade/issues/569#issuecomment-736756731
+        // for more information.
         .status_checked_with_codes(&[80])
```
SteveLauC, line 206: This behavior of `oh-my-zsh` should really be documented in the code or nobody knows what this magic number means

DottoDev (approved):

PR #438, opened 2023-05-25
Merge code for desktop notification between MacOS and Linux

TheCactusVert:
Linux now use the crate [notify-rust](https://crates.io/crates/notify-rust). BSD could use that crate too.
For Windows we could use [winrt-notification](https://crates.io/crates/winrt-notification).

## Standards checklist:

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -785,7 +785,7 @@ pub struct CommandLineArgs {
     /// Tracing filter directives.
     ///
     /// See: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/struct.EnvFilter.html
-    #[clap(long, default_value = "info")]
+    #[clap(long, default_value = "warn")]
     pub log_filter: String,
 
     /// Print completion script for the given shell and exit
--- a/src/config.rs
+++ b/src/config.rs
@@ -397,8 +397,6 @@ pub struct Misc {
 
     display_time: Option<bool>,
 
-    display_preamble: Option<bool>,
-
     assume_yes: Option<bool>,
 
     #[merge(strategy = crate::utils::merge_strategies::string_append_opt)]
@@ -1486,14 +1484,6 @@ impl Config {
             .misc
             .as_ref()
             .and_then(|misc| misc.display_time)
-            .unwrap_or(true)
-    }
-
-    pub fn display_preamble(&self) -> bool {
-        self.config_file
-            .misc
-            .as_ref()
-            .and_then(|misc| misc.display_preamble)
             .unwrap_or(true)
     }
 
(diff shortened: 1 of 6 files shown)


SteveLauC:
Yes, I have been curious about why topgrade uses notify-send binary to send notifications on Linux:(

src/main.rs
```diff
@@ -94,16 +94,6 @@ fn run() -> Result<()> {
...
-    #[cfg(target_os = "linux")]
-    {
-        if config.display_preamble() && terminal::supports_notify_send() && !config.skip_notify() {
-            print_warning("Due to a design issue with notify-send it could be that topgrade hangs when it's finished.
-If this is the case on your system add the --skip-notify flag to the topgrade command or set skip_notify = true in the config file.
-If you don't want this message to appear any longer set display_preamble = false in the config file.
-For more information about this issue see https://askubuntu.com/questions/110969/notify-send-ignores-timeout and https://github.com/topgrade-rs/topgrade/issues/288.");
-        }
-    }
-
```
SteveLauC, line 106: We can get rid of that hang issue, so great

SteveLauC:
> Linux now use the crate [notify-rust](https://crates.io/crates/notify-rust). BSD could use that crate too.
For Windows we could use [winrt-notification](https://crates.io/crates/winrt-notification).

It seems that all these platforms can use `notify-rust`, including [Windows](https://github.com/hoodie/notify-rust#windows-support)(it uses  `winrt-notifycation` under the hood)

TheCactusVert:
Great if notify-rust support all OSes, there is no need to use conditional compilation.
I have tested this PR on Linux only (x86_64 and aarch64 with and without DE). If someone could compile and test on both Windows and BSD it would be perfect :).

DottoDev (approved):

src/config.rs
```diff
@@ -787,7 +785,7 @@ pub struct CommandLineArgs {
     /// Tracing filter directives.
     ///
     /// See: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/struct.EnvFilter.html
-    #[clap(long, default_value = "info")]
+    #[clap(long, default_value = "warn")]
```
SteveLauC, line 788: Why was the default log level changed

SteveLauC:
And before merging the PR, would u like to squash your commits? So that the commits history looks clean and self-explanatory:)

src/config.rs
DottoDev: IMO warn should have been their in the first place, that's why I have approved it, but still curious about the explanation.

DottoDev:
@SteveLauC no worries, I always squash and merge ;)

src/config.rs
TheCactusVert: notify-rust was a bit talkative so I thought it was better.

SteveLauC: Understand. Makes sense to me:)

PR #460, opened 2023-06-02
fix: handle no topgrade.toml but files in topgrade.d

slowsage:
When config file(s) exist in `topgrade.d` but no `topgrade.toml` exists, `ensure()`  does not create a default config file.  This causes an error as `read()` assumes the `topgrade.toml` file exists.

This PR assumes the desired behavior is to use just the contents of the files in `topgrade.d`.

Else, one could change `ensure()` to create `topgrade.toml`.
 
## Standards checklist:

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -602,6 +602,12 @@ impl ConfigFile {
             path
         };
 
+        if config_path == PathBuf::default() {
+            // Here we expect topgrade.d and consequently result is not empty.
+            // If empty, Self:: ensure() would have created the default config.
+            return Ok(result);
+        }
+
         let mut contents_non_split = fs::read_to_string(&config_path).map_err(|e| {
             tracing::error!("Unable to read {}", config_path.display());
             e


SteveLauC:
Good catch! I can reproduce the issue that this PR tries to fix with the latest git version of `topgrade`.

```shell
$ pwd
/home/steve/.config/topgrade.d


$ ls
Permissions Links Size User  Group Date Modified Name
.rw-r--r--@     1  191 steve steve  2 Jun 14:59  topgrade.toml

$ topgrade --dry-run 2>&1 | head -n 2
ERROR Unable to read
ERROR failed to load configuration: No such file or directory (os error 2)
[1]    155988 IOT instruction (core dumped)  topgrade --dry-run 2>&1 |
       155989 done                           head -n 2
```

I don't have a strong view on how we should handle it under such a case (using files under `config.d` or creating a default config file), but honestly, I don't like the idea of `topgrade.d`, IMHO, it really makes the whole thing complicated, `topgrade` does not have so many configuration entries that it needs to disperse the configuration files across the file system. And if users do want to separate their configuration files into different files, the `[include]` section is sufficient. 

Friendly ping @DottoDev , any thoughts on this?

slowsage:
Note on `topgrade.d`: Another complication (and possible future issue) is the the [misc] section. Currently, it is required in the main `topgrade.toml` file but not in the files in `topgrade.d`.

DottoDev:
IMHO there should be the possibility to use `topgrade.d` but by default it should automatically use and/or create the`topgrade.toml` file at `$HOME/.config/`

slowsage:
> IMHO there should be the possibility to use `topgrade.d` but by default it should automatically use and/or create the`topgrade.toml` file at `$HOME/.config/`

Should [misc] be required in the files in topgrade.d?

DottoDev:
Best case would be that [misc] isn't required.

slowsage:
@DottoDev , to confirm:

1. `[misc]` required in main file (as is currently)
2. Do not process/require `[misc]` for files in `topgrade.d` (as is currently)
3. Create `topgrade.toml` if it doesn't exist even if there are files in `topgrade.d`

This way the  `topgrade.toml` file can be the only one to have `[misc]`, `[include]`. Avoiding `[include]` in `topgrade.d` avoids some recursive [include] cases.

Please lmk if that makes sense, I'll change the diff to address 3. I'll leave 1. and 2. as as is.

DottoDev:
@slowsage seems good, but I don't know if topgrade.toml has to be required if topgrade.d exists, imo it doesn't have to be created in this case.

slowsage:
@DottoDev , this PR handles exactly that case. i.e no topgrade.toml file => use defaults and apply files in topgrade.d if they exist. No file created. Please accept if it makes sense.

DottoDev (approved):

DottoDev:
Lgtm

PR #473, opened 2023-06-15
fix: Run AstroUpdate before Lazy sync

slowsage:
AstroNvim switched to using `Lazy` to manage plugins some weeks back.
For users of AstroNvim:
- Without this PR, topgrade runs `Lazy sync` (since `:Lazy` exists via AstroNvim) and quits without updating AstroNvim.
- With this PR, AstroUpdate is called if it exists before checking for existence of `Lazy.nvim`. AstroUpdate will update the plugins (using Lazy currently) as needed. So the explicit call to `Lazy sync` in `upgrade.vim` becomes unnecessary.

I suggest that users of AstroNvim add
```
  updater = {
    skip_prompts = true,
  },
```
to their `init.lua` to automatically confirm and quit AstroNvim after `AstroUpdate` and skip the redundant call to `Lazy sync`.
Users not using AstroNvim should be unaffected.

## Standards checklist:

### DIFF
--- a/src/steps/upgrade.vim
+++ b/src/steps/upgrade.vim
@@ -1,3 +1,14 @@
+" AstroUpdate calls a plugin manager - Lazy as of this writing. So we check for it before
+" others. Add to init.lua:
+" updater = {
+"   auto_quit = true,
+"   skip_prompts = true,
+" },
+if exists(":AstroUpdate")
+   echo "AstroUpdate"
+   AstroUpdate
+endif
+
 if exists(":NeoBundleUpdate")
     echo "NeoBundle"
     NeoBundleUpdate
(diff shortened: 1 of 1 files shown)


src/steps/upgrade.vim
```diff
@@ -33,16 +33,16 @@ if exists(":PaqUpdate")
...
 if exists(":AstroUpdate")
    echo "AstroUpdate"
    AstroUpdate
 endif
 
+if exists(":Lazy")
+    echo "Lazy Update"
+    Lazy! sync | qa
+endif
+
```
SteveLauC, line 45: Would you like to add some comments stating the rationale for this change in case someone in the future accidentally "reorders" them:)

SteveLauC:
Except for that doc nitpick, this PR LGTM:)

SteveLauC (approved):

slowsage:
Please hold off on merging. Just realized I need to test one more case. Might need to change the comment/code. I'll comment back when its ready.

slowsage:
Turns out the auto_quit option only gets triggered when there are updates. So we need to call quitall to cover that case. It's ready to be merged. Thank you.

DottoDev (approved):

PR #495, opened 2023-07-12
Update only base conda env

marcelotrevisani:
Please update just the base env, it is common to have others environment activated and I don't want to update them. The important bit that should be updated is the base env, I modified the code to update just the base environment in conda and also in mamba.

## Standards checklist:

- [ ] The PR title is descriptive.
- [ ] The code compiles (`cargo build`)
- [ ] The code passes rustfmt (`cargo fmt`)
- [ ] The code passes tests (`cargo test`)

### DIFF
--- a/src/steps/generic.rs
+++ b/src/steps/generic.rs
@@ -339,7 +339,7 @@ pub fn run_conda_update(ctx: &ExecutionContext) -> Result<()> {
     print_separator("Conda");
 
     let mut command = ctx.run_type().execute(conda);
-    command.args(["update", "--all"]);
+    command.args(["update", "--all", "-n", "base"]);
     if ctx.config().yes(Step::Conda) {
         command.arg("--yes");
     }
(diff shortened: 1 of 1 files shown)


SteveLauC:
I am not familiar with these 2 tools, and since you are a member of `conda`, in your opinion, do you think updating only the base environment is suitable for **most users**?

If yes, do you think it is necessary to add a configuration entry to updating all environments, like this:

```
# Updating all the environments instead of just the base env
[conda]
update_all = true

[mamba]
update_all = true
```

marcelotrevisani:
> I am not familiar with these 2 tools, and since you are a member of `conda`, in your opinion, do you think updating only the base environment is suitable for **most users**?
> 
> If yes, do you think it is necessary to add a configuration entry to updating all environments, like this:
> 
> ```
> # Updating all the environments instead of just the base one
> [conda]
> update_all = true
> 
> [mamba]
> update_all = true
> ```

hey, sorry for my delay.

Usually people want to update conda/mamba and not all the environments, the current behaviour will update the environment that is activated. I don't think to update all environments should be a default behaviour or even if it should be a feature, it might be optional, but I doubt that much people will enable it.

Just to be clear, conda and also mamba are installed just in the base environment, that is the environment where resides just conda/mamba responsible to manage all the environments, that is the reason I am proposing to change it to be just the base env.

SteveLauC:
In https://github.com/r-darwish/topgrade/issues/829, the op said:

> #### Which exact commands should Topgrade run?
> I'd say conda update --all -y, but I'm **not sure it upgrades all the environments and not only the currently selected one.**

Considering that, this PR LGTM:)

PR #532, opened 2023-08-06
Prioritize sudo steps to prevent sudo timeout

sujay1844:
Steps that don't require `sudo` like `flatpak` and `brew` are executed before steps that require `sudo` like `snap` and `pacstall`. This times out the sudo privilege and topgrade waits for the password. Ideally, sending a notification, when sudo is required, is a better solution but reordering the steps would work for most cases.

## Standards checklist:

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -187,15 +187,10 @@ fn run() -> Result<()> {
         }
         runner.execute(Step::ConfigUpdate, "config-update", || linux::run_config_update(&ctx))?;
 
-        runner.execute(Step::BrewFormula, "Brew", || {
-            unix::run_brew_formula(&ctx, unix::BrewVariant::Path)
-        })?;
-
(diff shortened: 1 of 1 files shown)


SteveLauC:
There is a configuration entry that could cache sudo credentials before running any steps so that you don't need to input the password for sudo during execution.

But anyway,  this PR LGTM

sujay1844:
Yes, `pre_sudo`. But when steps like `brew` and `flatpak` take more than 5 minutes to run, sudo session times out and waits for password on the next steps. This isn't reported to the user until the sudo prompt times out.

SteveLauC:
> But when steps like `brew` and `flatpak` take more than 5 minutes to run, sudo session times out and waits for password on the next steps.

Get it, make sense to me. Thanks!

SteveLauC (approved):

PR #536, opened 2023-08-12
feat: support Vscode packages updates

samhanic:
## What does this PR do

Visual studio code IDE automatically updates extensions only when it is opened. It forces then to reload its windows, which is a (small but existing) annoyance.
I use Topgrade regularly with great success but observed it does not yet handle these extensions. I consequently made this PR to add this update.

Vscode extensions cli update is not trivial because no universal upgrade command exists ( a [coresponding issue](https://github.com/microsoft/vscode/issues/56578) has been opened but not implemented). The steps I use are:
- get list of installed extensions with `code --list-extensions` (obtain a line-return separated list of installed extensions)
- construct the upgrade command and execute it
  - `code --force --install-extension [ext0] --install-extension [ext1] ... --install-extension [extN]`

I tested myself on Ubuntu 20.04 (including the case when no extension is installed) and on Windows 10.

## Standards checklist:

### DIFF
--- a/src/steps/generic.rs
+++ b/src/steps/generic.rs
@@ -318,6 +318,26 @@ pub fn run_vcpkg_update(ctx: &ExecutionContext) -> Result<()> {
     command.args(["upgrade", "--no-dry-run"]).status_checked()
 }
 
+pub fn run_vscode_extensions_upgrade(ctx: &ExecutionContext) -> Result<()> {
+    let vscode = require("code")?;
+    print_separator("Visual Studio Code");
+
+    let extensions = Command::new(&vscode)
+        .arg("--list-extensions")
+        .output_checked_utf8()?
+        .stdout;
+
+    for extension in extensions.split_whitespace() {
+        ctx.run_type()
+            .execute(&vscode)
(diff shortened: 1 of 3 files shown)


src/steps/generic.rs
```diff
@@ -318,6 +318,27 @@ pub fn run_vcpkg_update(ctx: &ExecutionContext) -> Result<()> {
...
+pub fn run_vscode_extensions_upgrade(ctx: &ExecutionContext) -> Result<()> {
+    let vscode = require("code")?;
+    print_separator("Visual Studio Code extensions");
+
+    let extensions = Command::new(&vscode)
+        .arg("--list-extensions")
+        .output_checked_utf8()?
+        .stdout;
+
+    if !extensions.is_empty() {
```
SteveLauC, line 330: Since an update command is currently unavailable, would you like to add some comments here stating the approach we use and linking that VSCode issue:)

samhanic: Yes, for sure! I just pushed a commit to comment that ;)

SteveLauC (approved):

PR #550, opened 2023-09-16
Fix "Nix" step to use `nix upgrade-nix` in more situations

9999years:
`nix upgrade-nix` can be used on any platform except NixOS where `nix` is available.

Also use `nix profile upgrade --verbose` because the non-verbose mode doesn't print anything on stdout.

Also add `--extra-experimental-features nix-command` to `nix` invocations so that they don't error.

## Standards checklist:

### DIFF
--- a/src/steps/generic.rs
+++ b/src/steps/generic.rs
@@ -417,7 +417,7 @@ pub fn run_pip3_update(ctx: &ExecutionContext) -> Result<()> {
     Command::new(&python3)
         .args(["-m", "pip"])
         .output_checked_utf8()
-        .map_err(|_| SkipStep("pip does not exists".to_string()))?;
+        .map_err(|_| SkipStep("pip does not exist".to_string()))?;
 
     let check_externally_managed = "import sysconfig; from os import path; print('Y') if path.isfile(path.join(sysconfig.get_path('stdlib'), 'EXTERNALLY-MANAGED')) else print('N')";
     Command::new(&python3)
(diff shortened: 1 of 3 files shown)


src/steps/generic.rs
```diff
@@ -417,7 +417,7 @@ pub fn run_pip3_update(ctx: &ExecutionContext) -> Result<()> {
     Command::new(&python3)
         .args(["-m", "pip"])
         .output_checked_utf8()
-        .map_err(|_| SkipStep("pip does not exists".to_string()))?;
+        .map_err(|_| SkipStep("pip does not exist".to_string()))?;
```
9999years, line 420: Drive-by typo fixes

src/steps/os/unix.rs
```diff
@@ -363,24 +365,23 @@ pub fn run_nix(ctx: &ExecutionContext) -> Result<()> {
...
+    #[allow(unused_mut)]
+    let mut should_self_upgrade = cfg!(target_os = "macos");
 
     #[cfg(target_os = "linux")]
     {
-        use super::linux::Distribution;
-
+        // We can't use `nix upgrade-nix` on NixOS.
         if let Ok(Distribution::NixOS) = Distribution::detect() {
-            return Err(SkipStep(String::from("Nix on NixOS must be upgraded via nixos-rebuild switch")).into());
```
9999years, line 380: We shouldn't skip the whole step on NixOS, because we can still upgrade user-level packages with `nix-env --upgrade` or `nix profile upgrade '.*'`

SteveLauC (approved):

SteveLauC:
LGTM, thanks! :)

PR #555, opened 2023-09-20
Add option for nix-env arguments

PabloMarcendo:
Last week `nix-env --upgrade` started compiling chromium because [its build](https://hydra.nixos.org/job/nixos/trunk-combined/nixpkgs.ungoogled-chromium.x86_64-linux) had not succeeded. To avoid compiling packages myself, I would like to be able to add `--prebuilt-only` to the nix-env arguments.

I have little experience with Rust, but everything seems to work as intended.

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -112,6 +112,7 @@
 #suse_dup = false
 #rpm_ostree = false
 #nix_arguments = "--flake"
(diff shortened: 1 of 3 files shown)


SteveLauC (approved):

SteveLauC:
LGTM, thanks!

PR #569, opened 2023-10-07
Fix builds and runs on DragonFly BSD

strangelittlemonkey:
* Include a missing use for a referenced type

* Add a missing step to do as is done on the FreeBSD code, which should be the same on this step as for DragonFly BSD

* Go against what clippy says for the sake of not actually crashing if an audit of packages fails. Absent this change, status_checked() will actually cause the run to crash at that step rather than say that that specific step failed and continue, which is not the expected behavior.

## Standards checklist:

- [ X] The PR title is descriptive.
- [X ] The code compiles (`cargo build`)
- [X ] The code passes rustfmt (`cargo fmt`)
- [ X] The code passes tests (`cargo test`)
- [ X] *Optional:* I have tested the code myself
 
## For new steps
  the underlying command

### DIFF
--- a/src/steps/os/dragonfly.rs
+++ b/src/steps/os/dragonfly.rs
@@ -20,6 +20,7 @@ pub fn upgrade_packages(ctx: &ExecutionContext) -> Result<()> {
 pub fn audit_packages(ctx: &ExecutionContext) -> Result<()> {
     let sudo = require_option(ctx.sudo().as_ref(), REQUIRE_SUDO.to_string())?;
     println!();
+    #[allow(clippy::disallowed_methods)]
     if !Command::new(sudo)
         .args(["/usr/local/sbin/pkg", "audit", "-Fr"])
         .status()?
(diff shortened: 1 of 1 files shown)


SteveLauC:
Thanks for doing this!

> Go against what clippy says for the sake of not actually crashing if an audit of packages fails.

If we get a clippy warning, and it is ignorable, we can use `#[allow()]` to silence it :)

strangelittlemonkey:
> Thanks for doing this!
> 
> > Go against what clippy says for the sake of not actually crashing if an audit of packages fails.
> 
> If we get a clippy warning, and it is ignorable, we can use `#[allow()]` to silence it :)

Are you asking me to write an allow right above the line in my commit?

SteveLauC:
> write an allow right above the line in my commit

Yep, that would silence it :)

SteveLauC (approved):

SteveLauC:
Thanks!

PR #574, opened 2023-10-12
Add support for pushing custom git repositories

### LINKED ISSUE #562: Pushing git repositories
Topgrade already supports pulling git repositories, but I'd like the optionto also push found repositories. 

Source code version management. located here https://github.com/git/git

The same as the existing git step

Probably if there are branches in the repository with unpushed commits. 

git push <branch> for the appropriate branches.

## More information
I can imagine this might have to be an opt in feature, but I'd really appreciate it. Just pushing to the default remote, and target branches with the same name and just erroring out if either of them is not set is enough for me and probably the safest option. 

Thanks for all the hard work <3

savente93:
## Standards checklist

Fixes #562 

Not the most elegant solution I'll admit. I'm a little sad about the duplication but I was having some real issues with the async code and opaque types when I tried to deduplicate it. This implementation works at least (at least it did for my system, don't have a setup for more robust testing) so hopefully this is good enough for a first pass. I've added config options that seemed reasonable to me. Happy reviewing!

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -116,10 +116,21 @@
 
 [git]
 #max_concurrency = 5
-# Additional git repositories to pull
+# Additional git repositories to push and pull 
 #repos = [
 #    "~/src/*/",
 #    "~/.config/something"
+#]
+
+# Repositories that you only want to pull
+#pull_only_repos = [
+#    "~/.config/something_else"
+#]
+
+# Repositories that you only want to push
+#push_only_repos = [
+#    "~/src/*/",
+#    "~/.config/something_third"
 #]
 
 # Don't pull the predefined git repos
@@ -127,9 +138,6 @@
 
 # Arguments to pass Git when pulling repositories
 #pull_arguments = "--rebase --autostash"
-
-# Also push custom repositories
-#push_custom_repos = true
 
 # Arguments to pass Git when pushing repositories
 #push_arguments = "--all"
--- a/config.example.toml
+++ b/config.example.toml
@@ -116,7 +116,7 @@
 
 [git]
 #max_concurrency = 5
-# Additional git repositories to push and pull 
+# Git repositories that you want to pull and push
 #repos = [
 #    "~/src/*/",
 #    "~/.config/something"
--- a/config.example.toml
+++ b/config.example.toml
@@ -125,11 +125,8 @@
 # Don't pull the predefined git repos
 #pull_predefined = false
 
-# Also push custom repositories
-#push_custom = false
-
 # Arguments to pass Git when pulling repositories
-#arguments = "--rebase --autostash"
+#pull_arguments = "--rebase --autostash"
 
 # Also push custom repositories
 #push_custom_repos = true
--- a/config.example.toml
+++ b/config.example.toml
@@ -125,8 +125,17 @@
 # Don't pull the predefined git repos
 #pull_predefined = false
 
-# Arguments to pass Git when pulling Repositories
+# Also push custom repositories
+#push_custom = false
+
+# Arguments to pass Git when pulling repositories
 #arguments = "--rebase --autostash"
+
+# Also push custom repositories
+#push_custom_repos = true
+
+# Arguments to pass Git when pushing repositories
+#push_arguments = "--all"
 
 [windows]
 # Manually select Windows updates
--- a/src/config.rs
+++ b/src/config.rs
@@ -915,7 +915,7 @@ impl Config {
         get_deprecated_moved_opt!(&self.config_file.misc, git_repos, &self.config_file.git, repos)
     }
 
-    /// The list of additional git repositories to pull.
+    /// Whether custom repositories should also be pushed.
     pub fn git_should_push_custom_repos(&self) -> bool {
         match &self.config_file.git {
             Some(g) => g.push_custom_repos.unwrap_or(false),
@@ -1031,7 +1031,7 @@ impl Config {
             .and_then(|misc| misc.ssh_arguments.as_ref())
     }
 
-    /// Extra Git arguments
+    /// Extra Git arguments for when pushing
     pub fn push_git_arguments(&self) -> &Option<String> {
         get_deprecated_moved_opt!(
             &self.config_file.misc,
@@ -1040,7 +1040,7 @@ impl Config {
             push_arguments
         )
     }
-    /// Extra Git arguments
+    /// Extra Git arguments for when pulling
     pub fn pull_git_arguments(&self) -> &Option<String> {
         get_deprecated_moved_opt!(
             &self.config_file.misc,
--- a/src/config.rs
+++ b/src/config.rs
@@ -912,7 +912,7 @@ impl Config {
         &self.config_file.commands
     }
 
-    /// The list of additional git repositories to pull.
+    /// The list of git repositories to pull and pull.
     pub fn git_repos(&self) -> &Option<Vec<String>> {
         get_deprecated_moved_opt!(&self.config_file.misc, git_repos, &self.config_file.git, repos)
     }
@@ -923,7 +923,7 @@ impl Config {
             .as_ref()
             .and_then(|git| git.pull_only_repos.as_ref())
     }
-    /// The list of additional git repositories to pull.
+    /// The list of git repositories to push.
     pub fn git_push_only_repos(&self) -> Option<&Vec<String>> {
         self.config_file
             .git
--- a/src/config.rs
+++ b/src/config.rs
@@ -917,22 +917,18 @@ impl Config {
         get_deprecated_moved_opt!(&self.config_file.misc, git_repos, &self.config_file.git, repos)
     }
     /// The list of additional git repositories to pull.
-    pub fn git_pull_only_repos(&self) -> &Option<Vec<String>> {
-        get_deprecated_moved_opt!(
-            &self.config_file.misc,
-            git_repos,
-            &self.config_file.git,
-            pull_only_repos
-        )
+    pub fn git_pull_only_repos(&self) -> Option<&Vec<String>> {
+        self.config_file
+            .git
+            .as_ref()
+            .and_then(|git| git.pull_only_repos.as_ref())
     }
     /// The list of additional git repositories to pull.
-    pub fn git_push_only_repos(&self) -> &Option<Vec<String>> {
-        get_deprecated_moved_opt!(
-            &self.config_file.misc,
-            git_repos,
-            &self.config_file.git,
-            push_only_repos
-        )
+    pub fn git_push_only_repos(&self) -> Option<&Vec<String>> {
+        self.config_file
+            .git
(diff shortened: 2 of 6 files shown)


config.example.toml
```diff
@@ -125,9 +125,18 @@
 # Don't pull the predefined git repos
 #pull_predefined = false
 
-# Arguments to pass Git when pulling Repositories
+# Also push custom repositories
+#push_custom = false
```
SteveLauC, line 129: This seems to be unused?

src/steps/git.rs
```diff
@@ -192,28 +223,31 @@ impl Git {
             .iter()
             .for_each(|pattern| print_warning(format!("Path {pattern} did not contain any git repositories")));
 
-        if repositories.repositories.is_empty() {
-            return Err(SkipStep(String::from("No repositories to pull")).into());
+        if repositories.pull_repositories.is_empty() && repositories.push_repositories.is_empty() {
```
SteveLauC, line 226: `Repositories::is_empty()` has been updated to:

```rs
pub fn is_empty(&self) -> bool {
     self.pull_repositories.is_empty() && self.push_repositories.is_empty()
}
```

so we can use it here

src/steps/git.rs
```diff
@@ -192,28 +223,31 @@ impl Git {
...
     pub fn multi_pull(&self, repositories: &Repositories, ctx: &ExecutionContext) -> Result<()> {
         let git = self.git.as_ref().unwrap();
 
         if ctx.run_type().dry() {
             repositories
-                .repositories
+                .pull_repositories
                 .iter()
-                .for_each(|repo| println!("Would pull {}", &repo));
+                .for_each(|repo| println!("Would push {}", &repo));
```
SteveLauC, line 244: This seems to be changed by accident

src/steps/git.rs
```diff
@@ -44,6 +45,36 @@ fn output_checked_utf8(output: Output) -> Result<()> {
...
+        .stdin(Stdio::null())
+        .current_dir(&repo)
+        .args(["push", "--porcelain"]);
+    if let Some(extra_arguments) = ctx.config().push_git_arguments() {
+        command.args(extra_arguments.split_whitespace());
+    }
+
+    let output = command.output().await?;
+    let result = match output.status.success() {
+        true => Ok(output.stdout),
```
SteveLauC, line 65: This `output.stdout` seems to be unused?

SteveLauC (changes requested):

SteveLauC:
Thanks for doing this! The core part of this PR LGTM, I have left some comments on some small things that need to be changed.


I have some thoughts on the public interfaces (i.e., the config file/how users would use this feature), and I wanna hear your thoughts as well:

1. In this PR, a user would define the repos that he wish to pull or push in:
   
     ```toml
     [git]
     repos = []
     ```

      When
      
      ```toml
      [git]
      push_custom_repos = true
      ```
      
      is set to true, these repos would be pulled and pushed

      What about changing it to:

      ```toml
      [git]
      pull_repos = []
      push_repos = []
      ```

      so that users can separate them, IMHO, a user won't wanna pull or push a repo at the same time?

2. Actually, everything in Topgrade is an opt-in feature, and you can see this from our code `config.rs`, every configuration entry is optional, you want to use it, you write it, and those options that are not present will be silently ignored
   
    > I can imagine this might have to be an opt in feature

    So it is not that necessary to ask a user to manually turn this feature on or off

      ```toml
      [git]
      push_custom_repos = true
      ```
3. This `arguments` entry
    
     ```toml
     [git] 
     # Arguments to pass Git when pulling repositories
     arguments = "--rebase --autostash"
     ```
     should be renamed to `pull_arguments` since we now have a `push_arguments`

     ```toml
     [git] 
     # Arguments to pass Git when pulling repositories
     pull_arguments = "--rebase --autostash"

     # Arguments to pass Git when pushing repositories
     push_arguments = "--all"
     ```

    Don't worry about breaking changes to our user, it is fine to bump our major number in the next release, as long as I document these changes well in the release note, it shouldn't be a problem :)

savente93:
Thanks for the good comments, I've updated the ones I missed so that should be all good now. I'm also going to address your comments in reverse order:
3. Indeed, I think that's a fairly easy migration path. We could also alias `arguments` to `pull_arguments` and add a deprecation warning if you want, I'm fairly ambivalent so I'll defer to your decision on that. 
2. agreed, I guess what I meant when I said "opt in" was, "we shouldn't just start pushing any repo we find" i.e. "we should not do the new thing unless told so" so we are in agreement there.
1. I think this is a UX decision. Personally I have a slight preference for the current implementation because that works better with my workflow. I have two machienes I work across at no real defined schedule, and I have a bunch of repositories on both that need to both read and write. Think dotfiles, personal projects, prototypes, etc. Bascially I just have a projects folder and that folder is globbed into the repos so everything there I want as up to date as possible. Therefore the current impl works a bit nicer because that way I don't have to maintain essentially duplicate lists, however I'm not sure how common this workflow is, and I agree that your suggestions allows more flexibility. so tl;dr I slightly prefer my solution, but am happy to go with yours if you prefer. 

In any case, thanks for the guidance and thoughtful comments, contributing here has been a very pleasant experience :)

SteveLauC:
> Personally I have a slight preference for the current implementation because that works better with my workflow. 

So with the approach I suggested, your configuration file would be something like this, right?

```toml
[git]
pull_repos = [
    "project1", 
    "project2",
]
push_repos = [
    "project1", 
    "project2",
]
```

> I don't have to maintain essentially duplicate lists

Indeed, duplicate paths are bad :<

So, what about the following approach:

```toml
[git]
# Repos specified here will be pulled AND pushed
repos = []
# Repos to pull
pull_repos = []
# Repos to push
push_repos = []
```

For the implementation, we still only store `pull_repositories` and `push_repositories`:

```rs
pub struct Repositories<'a> {
    git: &'a Git,
    pull_repositories: HashSet<String>,
    push_repositories: HashSet<String>,
    glob_match_options: MatchOptions,
    bad_patterns: Vec<String>,
}
```

in `main.rs`, we add all the repos specified in `repos = []` to `pull_repositories` and `push_repositories`:

```rs
if config.should_run(Step::GitRepos) {
        if let Some(git_repos) = config.git_repos() {
            for git_repo in custom_git_repos {
                git_repos.glob_insert_to_pull(git_repo);
                git_repos.glob_insert_to_push(git_repo);
            }
        }

        if let Some(git_pull_repos) = config.git_pull_repos() {
             for git_repo in git_pull_repos {
                git_repos.glob_insert_to_pull(git_repo);
            }
        }

        if let Some(git_push_repos) = config.git_push_repos() {
             for git_repo in git_push_repos {
                git_repos.glob_insert_to_push(git_repo);
            }
        }

  
        runner.execute(Step::GitRepos, "Git repositories", || {
            git.multi_repo_step(&git_repos, &ctx)
        })?;
}
```

savente93:
Oeh I like that, I'll make the changes!

savente93:
Should be all good now!

SteveLauC (changes requested):

SteveLauC:
Left some comments, I think this PR is ready for merge once the comments are resolved :)

src/config.rs
```diff
@@ -909,6 +916,24 @@ impl Config {
...
+    /// The list of additional git repositories to pull.
+    pub fn git_pull_only_repos(&self) -> &Option<Vec<String>> {
+        get_deprecated_moved_opt!(
+            &self.config_file.misc,
+            git_repos,
+            &self.config_file.git,
+            pull_only_repos
+        )
+    }
+    /// The list of additional git repositories to pull.
```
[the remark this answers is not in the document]
SteveLauC: Ah, this should not be called `additional` as well since it is `push`

src/steps/git.rs
```diff
@@ -27,9 +27,16 @@ pub struct Git {
     git: Option<PathBuf>,
 }
 
+#[derive(Clone, Copy)]
+pub enum GitAction {
+    Push,
+    Pull,
+}
```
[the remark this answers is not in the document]
savente93: Thank you ^_^

src/config.rs
```diff
@@ -905,10 +912,24 @@ impl Config {
         &self.config_file.commands
     }
 
-    /// The list of additional git repositories to pull.
+    /// The list of git repositories to pull and pull.
```
SteveLauC, line 915: "to push and pull"

SteveLauC (approved):

SteveLauC:
The CI failure is for the reason that `is_empty()` is `#[cfg(unix)]`, we need to remove the attribute then:

```rs
 #[cfg(unix)]
    pub fn is_empty(&self) -> bool {
        self.repositories.is_empty()
        self.pull_repositories.is_empty() && self.push_repositories.is_empty()
    }
```

savente93:
Shouldn't we also remove it for `remove`? I'm not sure why those would need to be marked as `unix` anyway.

SteveLauC:
> Shouldn't we also remove it for `remove`?  I'm not sure why those would need to be marked as `unix` anyway.

It seems that `is_empty()` and `remove()` are only used in the `oh-my-zsh` step, which is UNIX-only. It might be better to just remove the attribute for `is_empty()` or there will be a clippy warning on Windows (unused method or something like this)

SteveLauC:
Yeah, clippy gave us a warning :<

```
Run cargo clippy --locked --target x86_64-pc-windows-msvc --all-features -- -D warnings
   Compiling topgrade v12.0.2 (D:\a\topgrade\topgrade)
error: method `remove` is never used
   --> src\steps\git.rs:394:[12](https://github.com/topgrade-rs/topgrade/actions/runs/6505914484/job/17670480656?pr=574#step:7:13)
    |
326 | impl<'a> Repositories<'a> {
    | ------------------------- method in this implementation
...
394 |     pub fn remove(&mut self, path: &str) {
    |            ^^^^^^
    |
    = note: `-D dead-code` implied by `-D warnings`

error: could not compile `topgrade` (bin "topgrade") due to previous error
Error: Process completed with exit code 1.
```

savente93:
my bad, added it back in!

SteveLauC:
> my bad, added it back in!

Removing it for `is_empty()` is fine as it is indeed used on Windows

savente93:
done!

SteveLauC:
> done!

Seems that you forgot to push your commit?

savente93:
I think we had a race condition, I pushed the commit just before you made the comment. As far as I can see it's all there now.

SteveLauC:
I am sorry, my fault, it is indeed there, I am kinda dizzy this afternoon (my timezone)

savente93:
No worries, You've been very patient with me, so I think you deserve the same :)

SteveLauC:
Thanks! And congratulations on your first contribution to Topgrade!

savente93:
Thank you very much, it was a pleasure ^^

PR #608, opened 2023-11-14
feat: add aurutils support (WIP)

cyrinux:
Hi,

Thanks for the project 😃 

This adds support for ArchLinux aurutils AUR wrapper.

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -110,11 +110,14 @@
 [linux]
 # Arch Package Manager to use.
 # Allowed values:
-#   autodetect, aura, garuda_update, pacman, pamac, paru, pikaur, trizen, yay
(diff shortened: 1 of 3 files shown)


cyrinux:
After careful consideration, this was a bad idea, aurutils is too much flexible to not end with a too much opinionated `aur` parameters.

A workaround I already use is:

```
[commands]
"AUR update" = "xargs -a <(aur vercmp-devel | cut -d' ' -f1) aur sync -Scu --rebuild"  
```

Which is very opinionated.

SteveLauC:
> After careful consideration, this was a bad idea, aurutils is too much flexible to not end with a too much opinionated aur parameters.

Ok

 Thanks for your interst in contributing to Topgrade:)

PR #615, opened 2023-11-21
revert git pushing functionalities

savente93:
some people seem very unhappy about this feature, I don't have time to develop it futher at the moment, and I'm getting a little sick of being painted as a bad guy just for contributing, so I'd like this functionality removed for now. The feature isn't worth the trouble to me.

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -153,34 +153,18 @@
 
 
 [git]
-# max_concurrency = 5
-
-# Git repositories that you want to pull and push
-# repos = [
-#     "~/src/*/",
-#     "~/.config/something"
-# ]
-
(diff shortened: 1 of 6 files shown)


SteveLauC (approved):

SteveLauC:
> and I'm getting a little sick of being painted as a bad guy just for contributing,

I am sorry about this trouble, I am the person who should be blamed

There are some CI failures that need to be taken care of, let's ignore them for now and I will clean them up

savente93:
@SteveLauC for what it's worth, I don't think you're to blame at all, and I'm also sorry for the trouble it's caused you. Hope we can both move on from this now.

PR #622, opened 2023-11-27
Skip `nix upgrade-nix` when Nix is installed in a `nix profile`

9999years:
`nix upgrade-nix` errors if `nix` is installed in a `nix profile`:

```
error: directory '/Users/wiggles/.nix-profile/bin' does not appear to be part of a Nix profile
```

Upstream issue: https://github.com/NixOS/nix/issues/5473

This patch duplicates the relevant Nix logic and skips running `nix upgrade-nix` if this error would occur: https://github.com/NixOS/nix/blob/f0180487a0e4c0091b46cb1469c44144f5400240/src/nix/upgrade-nix.cc#L102-L139

I've also split the `nix upgrade-nix` call into a separate step so we can use the `SkipStep` error messages appropriately.

## Standards checklist:

## For new steps
  the underlying command

### DIFF
--- a/src/steps/os/unix.rs
+++ b/src/steps/os/unix.rs
@@ -1,11 +1,15 @@
+use std::ffi::OsStr;
 use std::fs;
 use std::os::unix::fs::MetadataExt;
+use std::path::Component;
 use std::path::PathBuf;
 use std::process::Command;
 use std::{env::var, path::Path};
 
 use crate::command::CommandExt;
 use crate::{Step, HOME_DIR};
+use color_eyre::eyre::eyre;
+use color_eyre::eyre::Context;
 use color_eyre::eyre::Result;
 use home;
 use ini::Ini;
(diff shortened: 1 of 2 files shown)


src/steps/os/unix.rs
```diff
@@ -432,6 +403,123 @@ pub fn run_nix(ctx: &ExecutionContext) -> Result<()> {
...
+    #[allow(unused_mut)]
+    let mut should_self_upgrade = cfg!(target_os = "macos");
+
+    #[cfg(target_os = "linux")]
+    {
+        // We can't use `nix upgrade-nix` on NixOS.
+        if let Ok(Distribution::NixOS) = Distribution::detect() {
+            should_self_upgrade = false;
+        }
+    }
```
SteveLauC, line 419:
```suggestion
    // Should we attempt to upgrade Nix with `nix upgrade-nix`?
    let mut should_self_upgrade = if cfg!(target_os = "macos") {
        true
    } else if cfg!(target_os = "linux") {
        !matches!(Distribution::detect(), Distribution::NixOS)
    } else {
        unreachable!()
    };
```

Maybe we can do this here, it has better readability, though the generated code looks bad:

```rs
// On Linux
    let should = if false {
        true
    } else if true {
        match Distribution::detect() {
            Distribution::NixOS => true,
            _ => false,
        }
    } else {
        ::core::panicking::panic("internal error: entered unreachable code")
    };

```

SteveLauC: Ah, I am sorry, I didn't realize that this `Distribution` type is only defined on Linux, and since `cfg!(target_os = "macos")` will be simply translated to a bool value, it requires that this type is also available on macOS, so it won't compile:(

9999years: Ahh, I'll revert.

SteveLauC (approved):
Thanks

PR #634, opened 2023-12-30
refactor: update pip if extern managed and global.break-system-packages is true

### LINKED ISSUE #631: update pip3 when pip config has set `global.break-system-packages`
## Erroneous Behavior

I tend to install base environment for general scripts, and venv for specific projects. Therefore I set pip to `global.break-system-packages` default. Therefore I can install some common packages like numpy and pandas globally conveniently. But topgrade skips pip3 since it only checks if `EXTERNALLY-MANAGED` exists (#367). But it's not the only way to allow pip to install packages globally.

my ~/.pip/pip.conf looks like

```toml
[global]
break-system-packages = true
```

you can use `pip config get global.break-system-packages` to get boolean `true` to check.

## Expected Behavior

updates pip3 when `pip config get global.break-system-packages` is true

## Steps to reproduce

add `break-system-packages = true` in pip.conf

SteveLauC:
## Standards checklist:

## For new steps
  the underlying command

Closes #631 

From the information provided in #631, even though a pip3 is externally managed, we should still update it if `global.break-system-packages` is set to true.

### DIFF
--- a/src/steps/generic.rs
+++ b/src/steps/generic.rs
@@ -425,20 +425,53 @@ pub fn run_pip3_update(ctx: &ExecutionContext) -> Result<()> {
         .output_checked_utf8()
         .map_err(|_| SkipStep("pip does not exist".to_string()))?;
 
-    let check_externally_managed = "import sysconfig; from os import path; print('Y') if path.isfile(path.join(sysconfig.get_path('stdlib'), 'EXTERNALLY-MANAGED')) else print('N')";
-    Command::new(&python3)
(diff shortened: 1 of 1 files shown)


SteveLauC:
[Here](https://github.com/SteveLauC/topgrade/releases/tag/pr634) is a build for x86_64/Linux/musl, @ErnestDong you can give it a test.

ErnestDong:
It works. Thank you. Happy new year

PR #638, opened 2024-01-06
fix(os) consider Fedora Kinoite to be the Fedora Silverblue distribution

michaeljwood:
## Standards checklist:

## For new steps
  the underlying command


src/steps/os/linux.rs
```diff
@@ -54,7 +54,7 @@ impl Distribution {
             Some("clear-linux-os") => Distribution::ClearLinux,
             Some("fedora") | Some("nobara") => {
                 return if let Some(variant) = variant {
-                    if variant.contains(&"Silverblue") {
+                    if variant.contains(&"Silverblue") || variant.contains(&"Kinoite") {
                         Ok(Distribution::FedoraSilverblue)
```
SteveLauC, line 58: What about we rename `Distribution::FedoraSilverblue` to `Distribution::FedoraImmutable` since they only differ in their desktop environments

SteveLauC: If you want to add tests on parsing their `/etc/os-release` files, you can find them here: https://github.com/which-distro/os-release/tree/main/fedora

michaeljwood: I like it.  I'll have another commit up soon.

SteveLauC (approved):
Thanks!

Issue #639, opened 2024-01-07
FreeBSD pkg audit should be optional, or not fail if status code != 0


wolpert:
## Erroneous Behavior

Currently in FreeBSD packages, cmake-core has a vulnerability. When topgrade executes `pkg audit -Fr` it displays the vulnerability the errors out because the status return is 1 and not 0. However, topgrade should continue. Right now, audit cannot be disabled for FreeBSD (AFAIK) and topgrade just exits out even though it should not. We need to either make the audit optional or let it not fail if status code result is not zero. 

## Expected Behavior

Calling `pkg audit -Fr` on FreeBSD where a vulnerability is found should display to error but continue going. Note I would not want to make it part of the System step or package upgrade step... as I want those 'as is'. If I had to remove it by disabling System, I'd just manually have to add in the upgrade. But if that was the required technique, It would at least work.

Preferred option to me is to create an audit step (default on) given auditing is part of multiple systems.

## Steps to reproduce

1. Use FreeBSD 14.x
2. `sudo pkg install cmake`
3. run `topgrade`

## Possible Cause (Optional)

The audit command is required and status checked.

Here is where the audit command is [required](https://github.com/topgrade-rs/topgrade/blob/ab35cd7b106b0ba43e0c355edb640d7ebb5068ef/src/main.rs#L274), and the [implementation](https://github.com/topgrade-rs/topgrade/blob/ab35cd7b106b0ba43e0c355edb640d7ebb5068ef/src/steps/os/freebsd.rs#L31) with status checked

Sample fix in this PR: https://github.com/topgrade-rs/topgrade/pull/640

## Problem persists without calling from topgrade
- [X] Yes
- [ ] No

Sorta, you can see this by doing the following:
1. `sudo pkg install cmake`
2. `pkg audit -Fr`
3. `echo $?`

## Did you run topgrade through `Remote Execution`

- [ ] Yes
- [X] No

If yes, does the issue still occur when you run topgrade directlly in your
remote host

- [ ] Yes
- [ ] No

## Configuration file (Optional)

Used the default configuration file.

## Additional Details
- Operation System/Version

14.0-RELEASE-p3

- Installation

Cargo

- Topgrade version (`topgrade -V`)

Topgrade 13.0.0

## Verbose Output (`topgrade -v`)

<details>
<pre>
DEBUG Configuration at /home/wolpert/.config/topgrade.toml
DEBUG Version: 13.0.0
[...]

PR #640, opened 2024-01-07
Added an Audit step for FreeBSD and DragonFly packagees.

wolpert:
Allows for auditing the packages to be disabled since they are breaking steps. Current behaivor is the default, where if the audit fails topgrade stops. Can be disabled in the [misc] section independenly from other sections.

Probably we should have a document explaining all the Steps in the system. Maybe as a separate PR?

This is a potential fix for the bug I reported here: https://github.com/topgrade-rs/topgrade/issues/639

## Standards checklist:

## For new steps
  the underlying command


wolpert:
BTW, tested in FreeBSD.

With default configuration file, there is no change in behavior.

When the following is added to the configuration file, the Audit step is removed.
```toml
[misc]
disable = ["audit"]
```

SteveLauC (changes requested):
Hi, thanks for this patch!

If you want to make audit a step, you should also call `print_separator()` in that function:

```diff
pub fn audit_packages(ctx: &ExecutionContext) -> Result<()> {
    let sudo = require_option(ctx.sudo().as_ref(), REQUIRE_SUDO.to_string())?;
    
    print_separator("FreeBSD Audit");
    
    Command::new(sudo)
        .args(["/usr/sbin/pkg", "audit", "-Fr"])
        .status_checked()
}
```

src/steps/os/dragonfly.rs
```diff
@@ -19,7 +19,9 @@ pub fn upgrade_packages(ctx: &ExecutionContext) -> Result<()> {
 
 pub fn audit_packages(ctx: &ExecutionContext) -> Result<()> {
     let sudo = require_option(ctx.sudo().as_ref(), REQUIRE_SUDO.to_string())?;
-    println!();
+
+    print_separator("DragonFly BSD Packages");
```
SteveLauC, line 23:
```suggestion
    print_separator("DragonFly BSD Audit");
```

wolpert:
> If you want to make audit a step, you should also call `print_separator()` in that function:

Good idea. Just updated with the change to both FreeBSD and DragonFly. Thanks!

SteveLauC (approved):
Thanks!

Issue #630, opened 2023-12-23
Support for `xcodes`


schrobingus:
### Which tool is this about? Where is its repository?
A tool that manages Xcode versions, which can be found [here](https://github.com/XcodesOrg/xcodes). 

I can work on this, just wanted to discuss prior to make sure this can be implemented without any possible glaring issues, since this tool installs individual Xcode versions as opposed to upgrading at once.

### Which operating systems are supported by this tool?
macOS only.

### What should Topgrade do to figure out if the tool needs to be invoked?
It should install the latest version, in correlation to if the user is running a stable, beta, or golden master release. Since this wouldn't be removing the old version automatically, it might be best to prompt the user before invoking (like with macOS system updates).

### Which exact commands should Topgrade run?
`xcodes update`
`xcodes list`
(sort output to find installed and latest, detect the keywords Beta and GM and mind them when finding the latest version)
`xcodes install (release)`

SteveLauC:
Hi, this sounds like a nice step to add, I have some questions:

1. For macOS/iOS developers, what will be included in an Xcode update if there is no macOS major version bump?

    > If there is a macOS update, IIRC, there will be a new SDK available.

2. Does `xcodes install (release)` need a confirmation? i.e., users need to input `yes/y` to approve the installation.
3. Does this command support the `--dry-run` option, i.e., show users what will be done without actually doing it.
4. Since this tool itself is not an Xcode updater but a manager, should we inform (print to stdout) users of what will be done?

BrentBoyMeBob:
1. You'd be correct that there are new SDKs. When it comes to minor updates:
    - Standard releases generally include bug fixes and sometimes minor feature additions.
    - Beta releases sometimes include the features initially announced. Other times, they will be added gradually through development in large stages (for example, the VisionOS toolkit and simulator). A lot heavier on bugs and their fixes here as well.
2. Not sure, I will test right now and return the results.
3. I don't think there is a dry run option bundled in if I remember correctly.
4. Absolutely, as for example, Xcode is large in terms of file size, and the old version won't be getting removed, so there should definitely be a brief overview of what will be run.

SteveLauC:
> 2. Does xcodes install (release) need a confirmation? i.e., users need to input yes/y to approve the installation.


> Not within the command, no, but there are some imperative prompts.

Besides that Apple ID confirmation, we can explicitly ask users if they want to update when detecting new releases (xcodes list) before we actually do it with `xcodes install xxx`.

[...]

Issue #543, opened 2023-08-25
macOS system update still runs with --dry-run


drrlvn:
## Erroneous Behavior
Running `topgrade` with `--dry-run` still runs the macOS system update instead of showing which commands would be run.

```
$ topgrade --only system --dry-run

── 11:00:21 - macOS system update ──────────────────────────────────────────────
Finding available software
No new software available.

── 11:00:31 - Summary ──────────────────────────────────────────────────────────
System upgrade: OK
```

## Expected Behavior
System update should not run, instead `topgrade` should print what it would do.

## Steps to reproduce
Use `--dry-run` flag.

## Problem persists without calling from topgrade
N/A
- [ ] Yes
- [ ] No

## Did you run topgrade through `Remote Execution`
- [ ] Yes
- [x] No

## Additional Details
- Operation System/Version
macOS 13.5.1

- Installation
Homebrew

- Topgrade version (`topgrade -V`)
Topgrade 12.0.2

## Verbose Output (`topgrade -v`)
<details>
<pre>
$ topgrade -v --only system --dry-run
DEBUG Configuration at /Users/drrlvn/.config/topgrade.toml
DEBUG Path ~/config-files expanded to /Users/drrlvn/config-files
DEBUG Loaded configuration: ConfigFile { include: None, misc: Some(Misc { pre_sudo: None, sudo_command: None, git_repos: None, predefined_git_repos: None, disable: None, ignore_failures: None, remote_topgrades: None, remote_topgrade_path: None, ssh_arguments: None, git_arguments: None, tmux_arguments: None, set_title: None, display_time: None, assume_yes: None, yay_arguments: None, aura_aur_arguments: None, aura_pacman_arguments: None, no_retry: None, run_in_tmux: None, cleanup: None, notify_each_step: None, accept_all_windows_updates: None, skip_notify: None, bashit_branch: None, only: None, no_self_update: None }), pre_commands: None, post_commands: None, commands: None, python: None, composer: None, brew: Some(Brew { greedy_cask: None, autoremove: Some(true) }), linux: None, git: Some(Git { max_concurrency: None, arguments: None, repos: Some(["/Users/drrlvn/config-files"]), pull_predefined: None }), windows: None, npm: None, yarn: None, vim: None, firmware: None, vagrant: None, flatpak: None, distrobox: None }
DEBUG Version: 12.0.2
DEBUG OS: aarch64-apple-darwin
[...]

PR #643, opened 2024-01-17
feat: adds `xcodes` step

schrobingus:
## Standards checklist:

This pull request is respondent to #630, by adding support for the tool `xcodes`, which manages Xcode versions. The step upgrades the tool directly in this order:

- The releases to upgrade will be filtered based on if the releases installed are regular, beta, or golden master.
- With those conditions, if a new release is detected, it'll prompt if the user would like to install, given that prompts are enabled.
- It'll follow through and install the new release.
- If there are only two releases installed, being the new release and old release, it'll prompt if the user would like to move the old release to the trash. This is ignored if prompts are disabled.

## For new steps
- [x] *Optional:* The `--yes` option works with this step ~~if it is supported by the underlying command~~, the command might require an Apple ID 2FA verification to install the new release regardless


schrobingus:
Huh, I'll check Rustfmt again.

src/steps/os/macos.rs
```diff
@@ -93,3 +94,79 @@ pub fn run_sparkle(ctx: &ExecutionContext) -> Result<()> {
     }
     Ok(())
 }
+
+pub fn update_xcodes(ctx: &ExecutionContext) -> Result<()> {
+    let xcodes = require("xcodes")?;
+    print_separator("Xcodes");
+
+    let should_ask = !(ctx.config().yes(Step::System)) || (ctx.config().dry_run());
```
SteveLauC, line 102: Seems that it should be the `Xcodes` step, and the condition should be something like:

```suggestion
    let should_ask = !(ctx.config().yes(Step::Xcodes) || ctx.config().dry_run());
```

src/steps/os/macos.rs
```diff
@@ -93,3 +94,79 @@ pub fn run_sparkle(ctx: &ExecutionContext) -> Result<()> {
...
+
+    let releases = ctx.run_type()
+        .execute(&xcodes)
+        .args(["update"])
+        .output_checked_utf8()?.stdout;
```
SteveLauC, line 107: Would you like to give me a output of `xcodes update` so that I can understand the code better?

src/steps/os/macos.rs
SteveLauC: It seems that you referred to the implementation of the System step on macOS while implementing this feature, and given that you are a macOS user, would you like to take a look at #543, [changing the definition of `should_ask` should fix it](https://github.com/topgrade-rs/topgrade/issues/543#issuecomment-1897676440), I would like to file a PR for it but I don't have a macOS machine at hand so I cannot give it a test

schrobingus: Can do, should I open a separate PR or resolve the issue here?

src/steps/os/macos.rs
schrobingus:
```
1.0 (7B85)
1.5 (7K571)
2.2.1 (8G1165)
2.3 (8M1780)
2.4 (8K1079)
...
15.1 Beta 2 (15C5042i)
15.1 Beta 3 (15C5059c)
15.1 (15C65)
15.2 Beta (15C5500c) (Installed)
15.2 (15C500b)
```

It just updates the list of releases available. As shown above, it also outputs the list at the same time. I can also run this on it's own and run `xcodes list` separately if it's more visibly sensible. I'll put a comment explaining this on the next commit.

src/steps/os/macos.rs
SteveLauC: Yep, it should be done in another PR, thanks for your help:)

SteveLauC: Seems that the condition is till not correct? `should_ask` should be false if either `ctx.config().yes(Step::Xcodes)` or `ctx.config().dry_run()` is `true`

src/steps/os/macos.rs
```diff
@@ -93,3 +94,104 @@ pub fn run_sparkle(ctx: &ExecutionContext) -> Result<()> {
...
+        .filter(|release| {
+            (allow_gm && release.contains("GM"))
+                || (allow_beta && release.contains("Beta"))
+                || (allow_regular && !release.contains("GM") && !release.contains("Beta"))
+        })
+        .map(String::from)
+        .collect();
+
+    if !releases_filtered
+        .last()
```
SteveLauC, line 145: It seems that I am missing something? Why are we only doing this to the `.last()` one?

schrobingus: The last one is the most recent release on the list. If it contains the "(Installed)" substring, then Xcode is fully up to date. On the contrary, it'll then ask if one would like to upgrade to the newest release.

src/steps/os/macos.rs
schrobingus: Exactly, I'll implement the change.

schrobingus: #668 :^)

src/steps/os/macos.rs
SteveLauC: Assume `xcodes update` gives the following output:

```
1.0 GM (0) (Installed)
1.1 GM (1)
1.4 Beta (0) (Installed)
1.5 Beta (1)
2.2.1 (0) (Installed)
2.2.2 (1)
```
`allow_gm`, `allow_beta` and `allow_regular` will all be `true`, `releases_filtered` will contain all the listed versions, in this case, if I understand correctly, we should install `1.1 GM (1)`, `1.5 Beta (1)` and `2.2.2 (1)` all rather than only the last one?

schrobingus: So update each release? I can do that, will commit that soon :^)

schrobingus:
Sorry for all the commits, I can clean that up if need be.

src/steps/os/macos.rs
```diff
@@ -93,3 +94,121 @@ pub fn run_sparkle(ctx: &ExecutionContext) -> Result<()> {
...
+    Ok(())
+}
+
+pub fn process_xcodes_releases(releases_filtered: Vec<String>, should_ask: bool, ctx: &ExecutionContext) -> Result<()> {
+    let xcodes = require("xcodes")?;
+
+    if releases_filtered
+        .last()
+        .map(|s| !s.contains("(Installed)"))
+        .unwrap_or(true)
```
SteveLauC, line 194: If `releases_filtered` is empty, then `.last()` will return `None`, and we treat `None` as `true`, then we will run:

```sh
$ xcodes install <an empty string>
```

I believe this won't happen in reality as there are so many Xcode releases, but I believe we should check if `releases_filtered` is empty or not before actually involving this function, if it is empty, we should directly return.

src/steps/os/macos.rs
```diff
@@ -93,3 +94,121 @@ pub fn run_sparkle(ctx: &ExecutionContext) -> Result<()> {
...
+        .args(["list"])
+        .output_checked_utf8()?
+        .stdout;
+
+    let releases_new_installed: HashSet<_> = releases_new
+        .lines()
+        .filter(|release| release.contains("(Installed)"))
+        .collect();
+
+    if should_ask && releases_new_installed.len() == 2 {
```
SteveLauC, line 171: If we update every kind of Xcode, should we also do this respectively?

src/steps/os/macos.rs
```diff
@@ -93,3 +94,121 @@ pub fn run_sparkle(ctx: &ExecutionContext) -> Result<()> {
...
+        .collect();
+
+    if installed_gm {
+        let _ = process_xcodes_releases(releases_gm, should_ask, ctx);
+    }
+    if installed_beta {
+        let _ = process_xcodes_releases(releases_beta, should_ask, ctx);
+    }
+    if installed_regular {
+        let _ = process_xcodes_releases(releases_regular, should_ask, ctx);
```
SteveLauC, line 156: Why are we ignoring the result of `process_xcodes_releases`, this could hide the possible errors

SteveLauC (changes requested):

src/steps/os/macos.rs
SteveLauC: > So update each release? 

Emm, I am sorry, I thought this is the original implementation plan :<

>  I can do that, will commit that soon :^)

Thanks for doing that!

src/steps/os/macos.rs
schrobingus: This is checked in `update_xcodes`, but I do agree that it is a good idea to directly return in the scenario that it is empty.

SteveLauC:
No need to worry about the codecov CI failures:)

src/steps/os/macos.rs
schrobingus: That makes sense, I'll add an error checker. Should `process_xcodes_releases` return a boolean indicating success?

schrobingus:
Applied those changes, but I just realized that I might want to also separate release candidates, so I'll quickly add a commit for those. :^)

SteveLauC (approved):
Thanks!

PR #661, opened 2024-01-23
Fix system updates for Nobara

thecatcore:
## Standards checklist:

## For new steps
  the underlying command

## Description

Nobara has some repositories that aren't updated by the default `dnf distro-sync --refresh`.
I've implemented the workaround indicated at [Nobara Project](https://nobaraproject.org/docs/upgrade-troubleshooting/how-do-i-update-the-system/).


SteveLauC:
Hi, thanks for this patch!

What will happen if you simply run `nobara-sync` in your terminal? Will it only launch the Update System app, and thus you still need to click the update button yourself?

src/steps/os/linux.rs
```diff
@@ -234,6 +237,42 @@ fn upgrade_redhat(ctx: &ExecutionContext) -> Result<()> {
...
+    let mut update_command = ctx.run_type().execute(sudo);
+    update_command.arg(&pkg_manager).arg("update");
+    // See https://nobaraproject.org/docs/upgrade-troubleshooting/how-do-i-update-the-system/
+    update_command.args([
+        "rpmfusion-nonfree-release",
+        "rpmfusion-free-release",
+        "fedora-repos",
+        "nobara-repos",
+    ]);
+    update_command.arg("--refresh").status_checked()?;
```
SteveLauC, line 253: This needs a `-y` option if `assume_yes` is set

src/steps/os/linux.rs
```diff
@@ -234,6 +237,42 @@ fn upgrade_redhat(ctx: &ExecutionContext) -> Result<()> {
     Ok(())
 }
 
+fn upgrade_nobara(ctx: &ExecutionContext) -> Result<()> {
+    let sudo = require_option(ctx.sudo().as_ref(), REQUIRE_SUDO.to_string())?;
+    let pkg_manager = which("dnf").unwrap_or_else(|| Path::new("yum").to_path_buf());
```
SteveLauC, line 242: If Nobara wants its users to use the `dnf` command, should we still fall back to `yum`?

src/steps/os/linux.rs
```diff
@@ -234,6 +237,42 @@ fn upgrade_redhat(ctx: &ExecutionContext) -> Result<()> {
...
+    update_command.arg("--refresh").status_checked()?;
+
+    let mut upgrade_command = ctx.run_type().execute(sudo);
+    upgrade_command
+        .arg(&pkg_manager)
+        .arg(if ctx.config().redhat_distro_sync() {
+            "distro-sync"
+        } else {
+            "upgrade"
+        });
```
SteveLauC, line 262: If the Nobara suggests that we should use:

```sh
$ sudo dnf distro-sync
```
it might be better to just follow it and use the `distro-sync` option

src/steps/os/linux.rs
```diff
@@ -234,6 +237,42 @@ fn upgrade_redhat(ctx: &ExecutionContext) -> Result<()> {
...
+
+    if let Some(args) = ctx.config().dnf_arguments() {
+        upgrade_command.args(args.split_whitespace());
+    }
```
SteveLauC, line 266: Kinda think we should no longer expose this

SteveLauC (changes requested):
Thanks for the PR!

Generally, I think we should follow the document, and use the suggested commands to do the update:

```sh
$ sudo dnf update rpmfusion-nonfree-release rpmfusion-free-release fedora-repos nobara-repos --refresh 
$ sudo dnf distro-sync --refresh 
$ sudo dnf update --refresh
```

though a `-y` option is needed is `assume_yes` is set to true.

thecatcore:
> Hi, thanks for this patch!
> 
> What will happen if you simply run `nobara-sync` in your terminal? Will it only launch the Update System app, and thus you still need to click the update button yourself?

Yes that is exactly what is happening, it opens the graphical app.

thecatcore:
> Thanks for the PR!
> 
> Generally, I think we should follow the document, and use the suggested commands to do the update:
> 
> ```shell
> $ sudo dnf update rpmfusion-nonfree-release rpmfusion-free-release fedora-repos nobara-repos --refresh 
> $ sudo dnf distro-sync --refresh 
> $ sudo dnf update --refresh
> ```
> 
> though a `-y` option is needed is `assume_yes` is set to true.

I agree with you, but using `--refresh` 3 times and `update` 2 times seems redundant, I've tested my implementation and it seems to be enough on my end so far.

SteveLauC (approved):
Thanks!

PR #666, opened 2024-01-26
feat: support wildcard in ignored_containers

### LINKED ISSUE #653: Support wildcards in ignored_containers
When ignoring containers to pull, it would be useful to be able to use wildcards.

When you use docker for development, you often end up with loads of locally built images with various tags.
These will never be possible to pull, and you wouldn't want to either, because they represent your local state.

While the new feature in v14 allows ignoring specific image tags, often it would be more effective to ignore large sets using wildcards. That way the image would be ignored even if I decide to tag it with a slug describing the current branch or time of day.

## More information
I'm not sure how I would test a side branch of topgrade, but I might be able to work on a PR.

If I were to tackle it, I think I would base the approach on [Globset](https://docs.rs/globset/latest/globset/), where the values in the config are interpreted as globs instead of doing equality matching. It would be backwards compatible with any currently configured values.

SteveLauC:
## Standards checklist:

## For new steps
  the underlying command

Closes #653

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2417,6 +2417,7 @@ dependencies = [
  "tracing-subscriber",
  "walkdir",
  "which",
+ "wildmatch",
  "winapi",
 ]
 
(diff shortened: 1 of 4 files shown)


SteveLauC:
Hi @mortenlj, I cannot give you a pre-built binary unless you are a Linux user, in case of so, would you like to build Topgrade against my branch and give it a test?

mortenlj:
I cloned your branch and built it. Seems to work wonders :+1:

PR #667, opened 2024-01-26
feat(container): add optional ingored_prefixes to ignore multiple tags

markkovari:
## Standards checklist:

## For new steps
  the underlying command

My use case - sorry for not creating an issue for that - is that I am using rancher desktop as k8s/docker solution. And I wanted to ignore multiple paths with a prefix.

I am not really sure if this is a huge impact, in my case it was 8-10 tags to ignore, but might be helpful for others. Tagging @JakobFels, since he did the previous container related one; what is your take here? Should we rather ignore things one by one or should we be able to skip with prefix?

At the same time the config options merging strategy might be totally incorrect, can anyone help me out with that part?

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -230,3 +230,4 @@
 # containers = ["archlinux-latest"]
 [containers]
 # ignored_containers = ["ghcr.io/rancher-sandbox/rancher-desktop/rdx-proxy:latest"]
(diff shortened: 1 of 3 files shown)


markkovari:
It was literally parallelly developed in #666 today. Great work @SteveLauC .

PR #670, opened 2024-01-27
Fix/winget

niStee:
## Standards checklist:

## For new steps
  the underlying command

### DIFF
--- a/BREAKINGCHANGES_dev.md
+++ b/BREAKINGCHANGES_dev.md
@@ -1,0 +1,7 @@
+# Configuration
+
+## Windows
+
+### Winget
+
+The `enable_winget = true` configuration was removed as winget is now enabled by default.
--- a/BREAKINGCHANGES_dev.md
+++ b/BREAKINGCHANGES_dev.md
@@ -1,7 +1,3 @@
-# Configuration
+# Configutation
 
-## Windows
-
-### Winget
-
-The `enable_winget = true` configuration was removed as winget is now enabled by default.
+1. The `enable_winget` configuration entry in the `windows` section has been removed because it will not cause any issues and will be enabled by default.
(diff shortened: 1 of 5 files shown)


src/steps/os/windows.rs
```diff
@@ -42,11 +42,6 @@ pub fn run_winget(ctx: &ExecutionContext) -> Result<()> {
 
     print_separator("winget");
 
-    if !ctx.config().enable_winget() {
-        print_warning("Winget is disabled by default. Enable it by setting enable_winget=true in the [windows] section in the configuration.");
-        return Err(SkipStep(String::from("Winget is disabled by default")).into());
-    }
-
```
SteveLauC, line 49: > Remove the check for 'winget_enable' set to 'true'. On my Windows 10 and 11 machines, there are no issues with Winget anymore. As far as I remember, it was disabled by default because it was buggy back then.

If this tool will no longer have issues, you should also consider removing its [configuration entry](https://github.com/topgrade-rs/topgrade/blob/2e70d132d05e3bdd5f5b7dd41f741aef0e4b3a50/config.example.toml#L187-L188), and NOTE, remove an existing config entry is a breaking change, you should document it in [BREAKINGCHANGES_dev.md](https://github.com/topgrade-rs/topgrade/blob/main/CONTRIBUTING.md#breaking-changes)

BREAKINGCHANGES_dev.md
```diff
@@ -0,0 +1,7 @@
+# Configuration
```
SteveLauC, line 1: It might be better to simply say:

```md
1. The `enable_winget` configuration entry in the `windows` section has been removed 
   as it won't have issues and thus will be enabled by default.
```

and keep the format simple:)

BREAKINGCHANGES_dev.md
```diff
@@ -0,0 +1,3 @@
+# Configutation
+
+1. The `enable_winget` configuration entry in the `windows` section has been removed because it will not cause any issues and will be enabled by default.
```
SteveLauC, line 3: Please [wrap in 80](https://github.com/topgrade-rs/topgrade/blob/main/CONTRIBUTING.md#breaking-changes)

niStee: like this

SteveLauC (approved):
Thanks!

PR #700, opened 2024-02-18
Create devskim.yml to enable GitHub code scanning for this repository

niStee:
## Standards checklist:

- [ ] The code compiles (`cargo build`)
- [ ] The code passes rustfmt (`cargo fmt`)
- [ ] The code passes tests (`cargo test`)
 
## For new steps
  the underlying command


SteveLauC:
This seems to be a nice feature to add, but reports warning even to `TODO`s:

image


Is there a rule that we can use to stop it from doing this?

.github/workflows/devskim.yml
```diff
@@ -0,0 +1,34 @@
...
+# separate terms of service, privacy policy, and support
+# documentation.
+
+name: DevSkim
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
```
SteveLauC, line 12: if we want to run this tool on PRs, then we should not limit the branch here

niStee:
> This seems to be a nice feature to add, but reports warning even to `TODO`s:
> 
> image
> 
> Is there a rule that we can use to stop it from doing this?

I have not yet found a way to not check for TODOs, but you can either [dismiss](https://docs.github.com/en/code-security/code-scanning/managing-code-scanning-alerts/managing-code-scanning-alerts-for-your-repository) them or [define](https://docs.github.com/en/code-security/code-scanning/managing-code-scanning-alerts/managing-code-scanning-alerts-for-your-repository) them as not severe enough to cause a pull-request to fail a review.

.github/workflows/devskim.yml
niStee: Consider to use [Configuring default setup for code scanning](https://docs.github.com/en/code-security/code-scanning/enabling-code-scanning/configuring-default-setup-for-code-scanning-at-scale#configuring-default-setup-for-all-eligible-repositories-in-an-organization) instead of this PR.

SteveLauC: Sadly, I don't have write permission so I cannot do this:<

SteveLauC:
> but you can either [dismiss](https://docs.github.com/en/code-security/code-scanning/managing-code-scanning-alerts/managing-code-scanning-alerts-for-your-repository) them or [define](https://docs.github.com/en/code-security/code-scanning/managing-code-scanning-alerts/managing-code-scanning-alerts-for-your-repository) them as not severe enough to cause a pull-request to fail a review.

I have dismissed those alerts.


Would you like to enable this check for PRs by not limiting the branches to `main`?

.github/workflows/devskim.yml
niStee: I think it is currently triggered on push or pull request events on the "main" branch.
On which events exactly do you want it to fire?
```suggestion
    on:
        push:
        pull_request:
            branches: [ "main" ]
```
In this configuration, the push event doesn't specify any branches, which means that the workflow will be triggered on a push to any branch in the repository.

niStee: Would you like a weekly check?

SteveLauC:
```yaml
on:
  pull_request:
  push:
    branches:
      - main
```

Maybe we can do something like this? Run it for every PR and pushes to the main branch? Is the `schedule` key compulsory? I am wondering if we can remove it and do the check at the pace of PR and pushes.

SteveLauC (approved):
Thanks! :)

PR #717, opened 2024-02-25
Add GitHub Actions Workflow for Build and Test

niStee:
## Standards checklist:

## For new steps
  the underlying command


niStee:
I plan to write some unit tests to increase the code coverage. This is just the first step to actually see what is already covered.
I think I prefer LLVM-tools over cargo2junit.
But I am trying to update the former code coverage workflow as well.

src/steps/os/windows.rs
```diff
@@ -1,4 +1,4 @@
-use std::convert::TryFrom;
+// use std::convert::TryFrom;
```
SteveLauC, line 1: Are you on a nightly toolchain and got some redundant import warnings?

niStee: exactly

SteveLauC: Can we remove it rather than commenting it if the stable toolchain is ok with this

SteveLauC:
> But I am trying to update the former code coverage workflow as well.

For these 2 workflows for code coverage, are they duplicates? I see both are using the `grocv` tool.

> I think I prefer LLVM-tools over cargo2junit.

I don't have a preference over these 2 tools, it will be ok as long as it works.

> I plan to write some unit tests to increase the code coverage. This is just the first step to actually see what is already covered.

Actually, even though we already had a workflow for tests coverage, it has been disabled manually since it failed every pull requests, but I am willing to enable it if we can make it work:)

niStee:
> > But I am trying to update the former code coverage workflow as well.
> 
> For these 2 workflows for code coverage, are they duplicates? I see both are using the `grocv` tool.
> 
> > I think I prefer LLVM-tools over cargo2junit.
> 
> I don't have a preference over these 2 tools, it will be ok as long as it works.
> 
> > I plan to write some unit tests to increase the code coverage. This is just the first step to actually see what is already covered.
> 
> Actually, even though we already had a workflow for tests coverage, it has been disabled manually since it failed every pull requests, but I am willing to enable it if we can make it work:)

Nothing like a solid 5% code coverage to ensure our code is bulletproof! 🙃

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,57 @@
...
+  push:
+    branches:
+      - main
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
+        rust: [stable, beta, nightly]
```
SteveLauC, line 13: Leaving `stable` and `nightly` will be sufficient for us, `beta` is too bleeding-edge :D

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,57 @@
...
+        uses: actions/cache@v4
+        with:
+          path: target
+          key: ${{ runner.os }}-cargo-build-target-${{ hashFiles('**/Cargo.lock') }}
+      - name: Set up Rust
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: ${{ matrix.rust }}
+          override: true
```
SteveLauC, line 37:
```suggestion
      - name: Set up Rust
        uses: dtolnay/rust-toolchain@v1
        with:
          toolchain: ${{ matrix.rust }}
```

Let's use [dtolnay/rust-toolchain](https://github.com/dtolnay/rust-toolchain) as `action-rs/toolchain` has been archived.

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,57 @@
...
+      - name: Install grcov
+        run: cargo install grcov
```
SteveLauC, line 39:
```suggestion
      - name: Install grcov
        uses: taiki-e/install-action@v2
        with:
          tool: grcov
```

Let's use the pre-built binaries.

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,57 @@
...
+        env:
+          RUSTFLAGS: -Cinstrument-coverage
+          LLVM_PROFILE_FILE: "%p-%m.profraw"
+      - name: Generate lcov.info
+        run: |
+          grcov . --binary-path ./target/debug/ -s . -t lcov --branch --ignore-not-existing --ignore "/*" -o lcov.info
+      - name: Upload coverage reports to Codecov
+        uses: codecov/codecov-action@v4.0.1
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
```
SteveLauC, line 55: I have set this token in the repo settings.

.github/workflows/code-coverage.yml
SteveLauC, line 1: Is this workflow still necessary if we have that `build-and-test.yml`?

SteveLauC (changes requested):

.github/workflows/code-coverage.yml
niStee: No

SteveLauC: Then we can remove it

.github/workflows/build-and-test.yml
niStee: If I understand [this](https://rust-lang.github.io/rustup/concepts/channels.html
) correctly nightly is the bleeding-edge, isn't it?

.github/workflows/build-and-test.yml
niStee: Maybe it works without a token because it's an open source repository. But I'm not sure.

.github/workflows/build-and-test.yml
niStee: sure

.github/workflows/build-and-test.yml
SteveLauC: Oh, that's right, I am sorry.

Anyway, let's keep `stable` and `nightly`, `beta` is barely used in CI :>

SteveLauC (approved):
Thanks!

SteveLauC:
I forgot that `beta` toolchain thing, anyway, let's do it in another PR:D

PR #749, opened 2024-03-23
feat: allow to configure the apt upgrade method on debian

### LINKED ISSUE #2294: Allow `apt upgrade` instead of `apt dist-upgrade`
### Checklist

- [x] I have searched the issue tracker for [existing feature requests](https://github.com/topgrade-rs/topgrade/issues?q=type%3AFeature).

### I want to suggest a feature

The `system` step on Debian/Ubuntu runs `apt dist-upgrade`. That can remove packages, **which is potentially destructive!**

Most users **never** use that during normal operation, and use `apt upgrade` instead. In the very rare cases where it's needed, it should be done by hand rather than by an automated tool.

Please add a config option to choose between the two, e.g.:
```toml
[linux]
apt_upgrade_type = "upgrade"  # or "dist-upgrade"
```

Thanks!

### Contribution

- [ ] I am able and willing to implement this feature myself.
- [ ] I am willing to test if someone else implements it. [Guide](https://github.com/topgrade-rs/topgrade?tab=contributing-ov-file#testing-a-pull-request)

thvitt:
Closes #2294

dist-upgrade, which was the hard-coded option, is allowed to remove packages in order to resolve dependency conflicts. This might cause quite a mess ... it is left as the default, but can be modified by the option apt_command.

## Standards checklist:

## For new steps
  the underlying command

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -167,6 +167,8 @@
 # Extra Home Manager arguments
 # home_manager_arguments = ["--flake", "file"]
 
+# subcommand to use for apt variants != nala (dist-upgrade (default) or upgrade)
+# apt_command = "dist-upgrade"
 
 [git]
 # How many repos to pull at max in parallel
--- a/src/config.rs
+++ b/src/config.rs
@@ -313,6 +313,7 @@ pub struct Linux {
     nix_env_arguments: Option<String>,
 
     #[merge(strategy = crate::utils::merge_strategies::string_append_opt)]
+    apt_command: Option<String>,
     apt_arguments: Option<String>,
 
     enable_tlmgr: Option<bool>,
@@ -1226,6 +1227,15 @@ impl Config {
             .as_ref()
             .and_then(|s| s.aura_pacman_arguments.as_deref())
             .unwrap_or("")
+    }
+
+    /// apt command: upgrade or dist-upgrade
+    pub fn apt_command(&self) -> &str {
+        self.config_file
+            .linux
+            .as_ref()
+            .and_then(|linux| linux.apt_command.as_deref())
+            .unwrap_or("dist-upgrade")
     }
 
     /// Extra apt arguments
--- a/src/steps/os/linux.rs
+++ b/src/steps/os/linux.rs
@@ -518,7 +518,8 @@ fn upgrade_debian(ctx: &ExecutionContext) -> Result<()> {
     if is_nala {
         command.arg("upgrade");
     } else {
-        command.arg("dist-upgrade");
+        let apt_command = ctx.config().apt_command();
+        command.arg(apt_command);
     };
     if ctx.config().yes(Step::System) {
         command.arg("-y");


SteveLauC:

The CI failures are not related, please rebase your branch (against main) to get them resolved:)

> dist-upgrade, which was the hard-coded option, is allowed to remove packages in order to resolve dependency conflicts. This might cause quite a mess

I can see this disaster, in your opinion, what is the appropriate default command to use? And, I am thinking, it is suitable for most users, then we can actually replace `dist-upgrade` with it

thvitt:
> I can see this disaster, in your opinion, what is the appropriate default command to use? And, I am thinking, it is suitable for most users, then we can actually replace `dist-upgrade` with it

`upgrade` certainly is the safer variant, OTOH I am using topgrade --yes on Debian testing, and they are just [migrating to a 64bit datetime type](https://wiki.debian.org/ReleaseGoals/64bit-time), so I shouldn’t be too surprised :)

My use case for topgrade is 'keep the system relatively up-to-date without too much hassle', so for that, `upgrade` would be the better default. BTW it looks like it’s the default for nala anyway (nala’s equivalent to `apt-get dist-upgrade` is `nala full-upgrade`)

SteveLauC:
> BTW it looks like it’s the default for nala anyway (nala’s equivalent to apt-get dist-upgrade is nala full-upgrade)

Thanks for the info!

-----

I am thinking about:

1. use `apt upgrade` as the default option (for `apt-fast` and `apt-get`)
2. Make the subcommand for `nala` and `apt-fast/apt-get` configurable
   
    available options that can be specified in the config file:

   1. upgrade (can be used with `nala/apt-fast/apt-get`)
   2. dist-upgrade (ONLY available for `apt-fast/apt-get`)
   3. full-upgrade (ONLY available for `nala`)

3. If an invalid option is specified in the config file, print an error message and skip the step during step execution (i.e., we don't do check if the config is valid while loading configuration)
4. For the config entry name, we might want to use something like this, specifying:
    
    1. It is for Debian-like distros
    2. This is the "upgrade" sub-command
   
    ```toml
    # The upgrade subcommand to use on Debian-like distros
    # For `apt-get` and `apt-fast`, one can use:
    # 1. "upgrade"
    # 2. "dist-upgrade"
    #
    # For `nala`, available options:
    # 1. "upgrade"
    # 2. "full-upgrade"
    debian_like_apt_upgrade_cmd = "value"
    ```
   

I would like to hear your thoughts on this since I am not a Debian user:)

thvitt:
The longer I think about it:
* I agree that upgrade should be the default as safest option.
* `nala full-upgrade`≈ `apt-get dist-upgrade --autoremove`
* You can, in theory, have apt, apt-fast, and nala installed in parallel and use a different one of them each day, since they all work on the same package store. 

I think it is a little confusing to configure the subcommand by hand but let the package manager be autodetected. So maybe the most sensible variant is to have the auto-detected tool with `upgrade` as default and just offer one configuration option that can be set to the whole command line, like `debian_like_upgrade_cmdline = "apt-fast full-upgrade --autoremove"`? The update command could just take the first word of this and run it with update, since that is all the same. 

Alternatively, maybe a boolean option `debian_like_dist_upgrade` that, when true, selects dist-upgrade or full-upgrade depending on the detected package manager.

SteveLauC:
> I think it is a little confusing to configure the subcommand by hand but let the package manager be autodetected. So maybe the most sensible variant is to have the auto-detected tool with upgrade as default and just offer one configuration option that can be set to the whole command line,

Yeah, I agree that it feels weird

>  like debian_like_upgrade_cmdline = "apt-fast full-upgrade --autoremove"? 

Though I feel Topgrade should not do such things either

GideonBear:
We could switch the default in v18 if we want.

lonix1:
My [issue](https://github.com/topgrade-rs/topgrade/issues/2294) was linked to this PR, but I'm unsure they deal with the same underlying problem? (There's discussion of internal implementation details which I didn't understand.)

Just to be sure, I think the solution is something simple like:

```toml
[linux]
apt_upgrade_type = "upgrade"  # or "dist-upgrade"
```

GideonBear:
> Just to be sure, I think the solution is something simple like:

If you look at the `config.example.toml` in this PR, you'll see that's what they do.

PR #755, opened 2024-03-26
Fix local host detection for remotes with user 

raindev:
Strip the user when comparing the remote with the hostname of the local machine to prevent repeatedly running topgrade over SSH on localhost.

### Set-up

topgrade is run on a machine with a hostname equal to `host`, which has `user@host` present in the `remote_upgrades` list in `topgrade.toml`.

### Current behavior

topgrade will trigger remote execution for `user@host` over SSH connection to `host`. This will repeat indefinitely due to the host recursively connecting to itself during each remote execution.

### New behavior

`user@host` remote will be recognized as being local to `host` and skipped by topgrade.

## Standards checklist:

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -22,7 +22,7 @@ use which_crate::which;
 use super::utils::editor;
 use crate::command::CommandExt;
 use crate::sudo::SudoKind;
-use crate::utils::{hostname, string_prepend_str};
+use crate::utils::string_prepend_str;
 use tracing::{debug, error};
 
 pub static EXAMPLE_CONFIG: &str = include_str!("../config.example.toml");
@@ -1456,16 +1456,16 @@ impl Config {
     #[cfg(target_os = "linux")]
     str_value!(linux, emerge_update_flags);
 
-    pub fn should_execute_remote(&self, remote: &str) -> bool {
-        let remote_host = remote.split_once('@').map_or(remote, |(_,host)| host);
+    pub fn should_execute_remote(&self, hostname: Option<&str>, remote: &str) -> bool {
+        let remote_host = remote.split_once('@').map_or(remote, |(_, host)| host);
 
-        if let Ok(hostname) = hostname() {
+        if let Some(hostname) = hostname {
             if remote_host == hostname {
                 return false;
             }
         }
 
-        if let Some(limit) = self.opt.remote_host_limit.as_ref() {
+        if let Some(limit) = &self.opt.remote_host_limit.as_ref() {
             return limit.is_match(remote_host);
         }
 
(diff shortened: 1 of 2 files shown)


SteveLauC:

From what I understand, this seems to be a new feature rather than a bug-fix as [it is explicitly documented that the remote host list should only contain hostnames](https://github.com/topgrade-rs/topgrade/blob/ab21600ca6cd160f8257557e8e119358cdf3b9a3/config.example.toml#L23)

>  to repeatedly running topgrade over SSH on localhost.

Could you please show me the benefit of this new feature?

raindev:
Hi,

Sorry for the confusion, the description was missing a word:

> to **prevent** repeatedly running topgrade over SSH on localhost

The ability to specify a user is handy in situations when it differs between the local and the remote machine.

Despite the example config file only describes specifying hostnames of remotes, `user@host` form already works (except that `host` will attempt to connect to itself when the remote is present in its own configuration, which the PR resolves). Given remote execution is over SSH, this would likely be what the users expect (an example from [a random config](https://github.com/doot/dotfiles/blob/cc45e2123d80b48f6df6ce205cc6fea0074cc3de/config/topgrade.toml#L24) found on GitHub) - other tools that leverage SSH would usually also allow to specify a user.

raindev:
I see two approaches for improving how topgrade handles `user@` included into a remote: 1) detect this and let the user know that this is not supported 2) take user into account when determining if topgrade should be run remotely. The option 2 is what the pull request implements. Right now including users into a remote will work as expected in some cases but not in the others (see the screenshot below). @SteveLauC let me know what do you think.

topgrade-remote-user

SteveLauC:
Hi, I didn't reply, it is not because I am opposed to this change, it is simply because I am busy with my work and couldn't have time to handle these OSS things, I am sorry about this.

SteveLauC:
Ok, I just gave this PR a closer look, and the issue described in the PR description indeed exists, the approach proposed in this PR looks good to me!

src/config.rs
```diff
@@ -1456,15 +1456,17 @@ impl Config {
     #[cfg(target_os = "linux")]
     str_value!(linux, emerge_update_flags);
 
-    pub fn should_execute_remote(&self, remote: &str) -> bool {
-        if let Ok(hostname) = hostname() {
-            if remote == hostname {
+    pub fn should_execute_remote(&self, hostname: Result<String>, remote: &str) -> bool {
```
SteveLauC, line 1459: I am curious why we added this new argument (`hostname`),  looks like the previous way how we handled it still works

SteveLauC: Ahh, I get it, it is for tests

SteveLauC (approved):

raindev:
No need to apologise, I just wasn't sure if I explained the problem clearly enough previously. 

Thanks for your work on topgrade!

PR #763, opened 2024-04-02
Add config flag to toggle verbose Git repository output.

dsully:
If `true`: the default, no change.

If `false`: Only show repositories that have been updated or have an error.

Minor tweak to output (removed colon) so that copy and paste for 'cd' is nicer.

## Standards checklist:

## For new steps
  the underlying command

### DIFF
--- a/src/steps/git.rs
+++ b/src/steps/git.rs
@@ -292,10 +292,10 @@ impl RepoStep {
     }
 
     /// Try to pull a repo.
-    async fn pull_repo<P: AsRef<Path>>(&self, ctx: &ExecutionContext<'_>, repo: P, verbose: bool) -> Result<()> {
+    async fn pull_repo<P: AsRef<Path>>(&self, ctx: &ExecutionContext<'_>, repo: P) -> Result<()> {
         let before_revision = get_head_revision(&self.git, &repo);
 
-        if verbose {
+        if ctx.config().verbose() {
             println!("{} {}", style("Pulling").cyan().bold(), repo.as_ref().display());
         }
 
@@ -344,7 +344,7 @@ impl RepoStep {
                     println!();
                 }
                 _ => {
-                    if verbose {
+                    if ctx.config().verbose() {
                         println!("{} {}", style("Up-to-date").green().bold(), repo.as_ref().display());
                     }
                 }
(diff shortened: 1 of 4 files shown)


SteveLauC (commented):
Hi!

I wonder can we do this with logs? Then we can toggle them via log level, Topgrade already has a `--verbose` option, IMHO, it would match this well

dsully:
> Hi!
> 
> I wonder can we do this with logs? Then we can toggle them via log level, Topgrade already has a `--verbose` option, IMHO, it would match this well

Easy enough! Just pushed a new commit.

src/steps/git.rs
```diff
@@ -292,10 +292,12 @@ impl RepoStep {
     }
 
     /// Try to pull a repo.
-    async fn pull_repo<P: AsRef<Path>>(&self, ctx: &ExecutionContext<'_>, repo: P) -> Result<()> {
+    async fn pull_repo<P: AsRef<Path>>(&self, ctx: &ExecutionContext<'_>, repo: P, verbose: bool) -> Result<()> {
```
SteveLauC, line 295: We don't need an extra argument here, we can access the `--verbose` option through `ctx.config().verbose()`

SteveLauC (changes requested):

src/steps/git.rs
```diff
@@ -292,10 +292,12 @@ impl RepoStep {
...
         let before_revision = get_head_revision(&self.git, &repo);
 
-        println!("{} {}", style("Pulling").cyan().bold(), repo.as_ref().display());
+        if verbose {
+            println!("{} {}", style("Pulling").cyan().bold(), repo.as_ref().display());
+        }
```
SteveLauC, line 300: I kinda think it is unnecessary to inform users of this on every run, thoughts?

src/steps/git.rs
```diff
@@ -364,6 +368,12 @@ impl RepoStep {
             return Ok(());
         }
 
+        let verbose = ctx.config().verbose();
+
+        if !verbose {
+            println!("\n{} updated repositories will be shown...\n", style("Only").green().bold());
+        }
```
SteveLauC, line 375: I kinda think it is unnecessary to inform users of this on every run, thoughts?

dsully: I could go either way. It's somewhat helpful if you have a lot of repositories to know that something is happening.

SteveLauC: > It's somewhat helpful if you have a lot of repositories to know that something is happening.

i agree

SteveLauC (approved):
Thx!

PR #770, opened 2024-04-13
sudo

olivertzeng:
## Standards checklist:

## For new steps
  the underlying command


SteveLauC:
Could you explain a bit what this is for? i.e., the problem this patch would solve

olivertzeng:
> Could you explain a bit what this is for? i.e., the problem this patch would solve

Certainly!
In the example config .toml file there's an option of `sudo -v` preventing users from having to retype the sudo password in the middle of the update proccess
however this didn't work for me. then i found out that due to the timeout settings, it'll revoke the validation to sudo within 15 minutes by default meaning that you'll eventually need to retype the password again.

the `-T` flag stands for "time" meaning the time it takes to revoke the certificate. if set to 0 it'll not revoke for this session

SteveLauC:
> then i found out that due to the timeout settings, it'll revoke the validation to sudo within 15 minutes by default meaning that you'll eventually need to retype the password again.

We do have users complain about this session timeout issue

> the -T flag stands for "time" meaning the time it takes to revoke the certificate. if set to 0 it'll not revoke for this session

I am not aware of that this option can adjust the session timeout value, and a Google search didn't tell me that either, would you like to show me some documents on this thing?

From the man page (also the doc added in this PR), it looks like this option is for the command after `sudo`

> Used to set a timeout for the command.

olivertzeng:
> > then i found out that due to the timeout settings, it'll revoke the validation to sudo within 15 minutes by default meaning that you'll eventually need to retype the password again.
> 
> We do have users complain about this session timeout issue
> 
> > the -T flag stands for "time" meaning the time it takes to revoke the certificate. if set to 0 it'll not revoke for this session
> 
> I am not aware of that this option can adjust the session timeout value, and a Google search didn't tell me that either, would you like to show me some documents on this thing?
> 
> From the man page (also the doc added in this PR), it looks like this option is for the command after `sudo`
> 
> > Used to set a timeout for the command.

IDK for me with the `sudo -v` flag on it still asks me the password when upgrading with `yay`

SteveLauC:
>  for me with the sudo -v flag on it still asks me the password when upgrading with yay

With the new `-T 0` option, you won't be asked for password? 

Let's me give this a test on my Fedora machine

olivertzeng:
Probably  Sent from Proton Mail for iOS 
於 週六, 4月 13, 2024 時間 17:22, SteveLauC ***@***.***&gt; 寫道：  
for me with the sudo -v flag on it still asks me the password when upgrading with yay

With the new -T 0 option, you won't be asked for password?
Let's me give this a test on my Fedora machine

SteveLauC:
On my Fedora 39 machine, this option does not work, [the default timeout on Fedora is 5 mins](https://docs.fedoraproject.org/en-US/quick-docs/performing-administration-tasks-using-sudo/#_sudo_timeout):

> By default, sudo stores the password for a five minute timeout period.

```sh
$ sudo -vT 0

# 9 mins passed
$ sudo ls
[sudo] password for steve:  # Password needed
```

olivertzeng:
@SteveLauC 
bro im stupid

> In this case, we configure the session length to 666 minutes via the timestamp_timeout option. Using 0 forces sudo to always require a password, while a negative value means infinite session lengths for the lifetime of the [shell](https://www.baeldung.com/linux/sh-vs-bash#what-is-a-shell) and [terminal](https://www.baeldung.com/linux/pty-vs-tty#terminals-terminals-terminals).

it should be -1

SteveLauC:
> Using 0 forces sudo to always require a password,

```sh
$ sudo -vT 0
[sudo] password for steve:

$ sudo ls # type this command immediately
xx xx xx
```

Emm, still weird, 0 does not always require a password on my machine

olivertzeng:
DamnThese information are the things that I’ve search from the internet
Btw does sudo -v work for u
 Sent from Proton Mail for iOS 
於 週六, 4月 13, 2024 時間 19:22, SteveLauC ***@***.***&gt; 寫道：  
Using 0 forces sudo to always require a password,

$ sudo -vT 0
[sudo] password for steve:

$ sudo ls # type this command immediately
xx xx xx
Emm, still weird, 0 does not always require a password on my machine

SteveLauC:
> Btw does sudo -v work for u

I think yes, it starts a new sudo session, which is 5 mins on Fedora

olivertzeng:
> > Btw does sudo -v work for u
> 
> I think yes, it starts a new sudo session, which is 5 mins on Fedora

well shouldn't it been set to make it never expire in this session?

btw ive asked https://deep.ai about this

> The sudo -vT 0 and sudo -vT -1 commands have similar functionalities, but with one key difference in how long the cached credentials are valid for:

    sudo -vT 0:

    The 0 value specifies that the cached credentials will not be stored for any amount of time. This means that the user's credentials will be immediately cleared after running this command. As a result, if a user tries to execute a sudo command after running sudo -vT 0, they will be prompted to enter their password again for authentication.

    sudo -vT -1:

    The -1 value indicates that the cached credentials will never expire unless manually updated or revoked. This means that the user's credentials will remain valid indefinitely until the user explicitly refreshes them or logs out. Therefore, users will not be prompted to enter their password for subsequent sudo commands until they decide to update or revoke the cached credentials.

In summary, sudo -vT 0 clears the user's cached credentials immediately, while sudo -vT -1 sets the cached credentials to never expire until explicitly updated or revoked.

olivertzeng:
update ive tried both `sudo -vT 0` and `sudo -vT -1` just now
and neither of them worked

Issue #572, opened 2023-10-11
Dotnet may fail when trying to update tools just after it was updated from package manager


thecatcore:
## Erroneous Behavior

Dotnet step tried to update a tool "-------------------------------------" which is the separator, just after updating it from package manager.
I don't have any tools installed so it simply should not have tried running.

## Expected Behavior

Dotnet step doesn't run because I don't have any tools installed.

## Steps to reproduce

Run topgrade when a dotnet update is available through the package manager.

## Possible Cause (Optional)

When installing or updating dotnet through the package manager, it warns the user about a command running in the background to finish setuping it after the package manager is done running.
My guess is that in this case, the background command wasn't done when topgrade arrived to the dotnet step, and because of this there might have been more lines logged before the list of tools than usual (2, which are skipped in topgrade).

## Problem persists without calling from topgrade

- [ ] Yes
- [ ] No

## Did you run topgrade through `Remote Execution`

- [ ] Yes
- [x] No

If yes, does the issue still occur when you run topgrade directlly in your
remote host

- [ ] Yes
- [ ] No

## Configuration file (Optional)

```toml

```

## Additional Details
- Operation System/Version
  
Pop!_OS 22.04

- Installation
  
cargo/crates.io

- Topgrade version (`topgrade -V`)
12.0.2

## Verbose Output (`topgrade -v`)

<details>
<pre>

</pre>
</details>

SteveLauC:
> Dotnet step tried to update a tool "-------------------------------------" which is the separator

Can you still reproduce it, if so, could you please give me the output of:

```sh
$ dotnet tool list --global
```

SteveLauC:
Ah, I just installed dotnet-SDK on my host, and reproduce it:

```sh
$ topgrade --only dotnet

── 09:03:23 - Sudo ─────────────────────────────────────────────────────────────

── 09:03:23 - .NET ─────────────────────────────────────────────────────────────
/tmp/4ffdd387-40aa-4171-a469-56e9e636ccb0/restore.csproj : error NU1101: Unable to find package ---------------------. No packages exist with this id in source(s): nuget.org
Tool '---------------------' failed to update due to the following:
The tool package could not be restored.
Tool '---------------------' failed to install. This failure may have been caused by:

* You are attempting to install a preview release and did not use the --version option to specify the version.
* A package by this name was found, but it was not a .NET tool.
* The required NuGet feed cannot be accessed, perhaps because of an Internet connection problem.
[...]

PR #772, opened 2024-04-13
fix: better `dotnet tool list` header parsing

AThePeanut4:
I've just run into the same error as #572, in the same scenario (right after updating .NET). I'm also fairly sure I've had it happen before.

If I remove the `.dotnet` folder in my home directory, I get this output from `dotnet tool list`:
```
> mv .dotnet .dotnet2
> DOTNET_NOLOGO=true dotnet tool list --global 2> /dev/null
An issue was encountered verifying workloads. For more information, run "dotnet workload update".
Package Id      Version      Commands
-------------------------------------
<list of packages>
```
I think this is the the cause of the error, although I'm not 100% sure. Either way, topgrade should parse out the header in a more reliable way than just skipping the first two lines of output. 

## Standards checklist:

- [x] *Optional:* I have tested the code myself.
  Well, it no longer fails after removing the `~/.dotnet` folder. Hopefully this means it should no longer error right after a .NET update.

### DIFF
--- a/src/steps/generic.rs
+++ b/src/steps/generic.rs
@@ -773,6 +773,7 @@ pub fn run_dotnet_upgrade(ctx: &ExecutionContext) -> Result<()> {
         }
     };
 
+    let mut in_header = true;
     let mut packages = output
         .stdout
         .lines()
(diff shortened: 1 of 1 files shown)


SteveLauC (approved):
Thanks!

PR #796, opened 2024-05-11
feat(cargo): remove dependence on install-update

InnocentZero:
Fixes the dependency of cargo update on cargo-install-update binary and also fixes the issue where packages are installed without the relevant features. 

## Standards checklist:

## For new steps
  the underlying command

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2133,9 +2133,9 @@ dependencies = [
 
 [[package]]
 name = "serde_json"
-version = "1.0.114"
+version = "1.0.117"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c5f09b1bd632ef549eaa9f60a1f8de742bdbc698e6cee2095fc84dde5f549ae0"
+checksum = "455182ea6142b14f93f4bc5320a2b31c1f266b66a4a5c858b013302a5d8cbfc3"
 dependencies = [
  "itoa",
  "ryu",
@@ -2610,6 +2610,7 @@ dependencies = [
  "self_update",
  "semver",
  "serde",
+ "serde_json",
  "shell-words",
  "shellexpand",
  "strum",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,6 +51,7 @@ merge = "~0.1"
 regex-split = "~0.1"
 notify-rust = "~4.10"
 wildmatch = "2.3.0"
+serde_json = "1.0.117"
 
 [package.metadata.generate-rpm]
 assets = [{ source = "target/release/topgrade", dest = "/usr/bin/topgrade" }]
--- a/src/steps/generic.rs
+++ b/src/steps/generic.rs
@@ -12,9 +12,7 @@ use color_eyre::eyre::Result;
 use semver::{Version, VersionReq};
 use serde_json::Value;
 use tempfile::tempfile_in;
-use tracing::instrument::WithSubscriber;
 use tracing::{debug, error};
-use tracing_subscriber::fmt::writer::MakeWriterExt;
 
 use crate::command::{CommandExt, Utf8Output};
 use crate::execution_context::ExecutionContext;
--- a/src/steps/generic.rs
+++ b/src/steps/generic.rs
@@ -146,7 +146,6 @@ fn manual_cargo_updates(crates_json: PathBuf) -> Result<()> {
             return Err(SkipStep("Improper .crates2.json".to_string()).into());
         }
     }
-    println!("{feature_packages:?}");
 
     for (package, all_features, no_default, feature) in feature_packages {
         let feature: Vec<_> = feature.iter().map(|feat| feat.as_str().unwrap()).collect();
--- a/src/steps/generic.rs
+++ b/src/steps/generic.rs
@@ -1,5 +1,6 @@
 #![allow(unused_imports)]
 
+use std::fs::File;
 use std::path::PathBuf;
 use std::process::Command;
 use std::{env, path::Path};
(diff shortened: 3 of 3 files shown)


InnocentZero:
@SteveLauC can you tell me what the failing test means? Because I cannot even understand what it is doing

SteveLauC:
> @SteveLauC can you tell me what the failing test means? Because I cannot even understand what it is doing

No worries about it, the test coverage CI is simply broken

SteveLauC (commented):

> Fixes the dependency of cargo update on `cargo-install-update` binary 

Well, I do think it is good to get rid of an external dependency, but `cargo-install-update` has been pretty much the defacto option to update binaries installed through `cargo`, for Topgrade, using `cargo-install-update` makes us:

1. Easy to maintain
2. No need to reinvent the wheel

so I am not quite sure we should remove it.

For the impl of this PR, it checks the cargo version and only does a manual update if the version is greater than or equal to 1.77, so I am guessing the cargo folks have done something great to make cargo more like a package manager in that release?

And it seems that the current impl does not handle those binaries installed from git.

> fixes the issue where packages are installed without the relevant features.

So IIUC, this is a bug of `cargo-install-update`, I think we should file an issue in the upstream repo.

InnocentZero:
> For the impl of this PR, it checks the cargo version and only does a manual update if the version is greater than or equal to 1.77, so I am guessing the cargo folks have done something great to make cargo more like a package manager in that release?

Prior to that version, cargo would error out on reinstalling a package. Now it will cleanly exit and upgrade the package if needed. In that regard this is a massive improvement. See [this commit](https://github.com/rust-lang/cargo/commit/fb4415090f600bae51b0747bef2e7049070cd6ee) and this addition in cargo-update's [README](https://github.com/rust-lang/cargo/commit/fb4415090f600bae51b0747bef2e7049070cd6ee).
 
> And it seems that the current impl does not handle those binaries installed from git.

This is a WIP PR, sorry I should have made it clear.

> > fixes the issue where packages are installed without the relevant features.
> 
> So IIUC, this is a bug of `cargo-install-update`, I think we should file an issue in the upstream repo.

Actually not! (kinda yes but also no). So `cargo install-update -a` command does not check the features of the packages it is updating, but it can do so individually if supplied with the appropriate flags (similar to `cargo install xyz --features A,B`). However to check those flags you have to read `.crates2.json` and NOT `.crates.toml` (which I suppose is being read by both this repo and upstream rn). But if you go all the way to read `.crate2.json`, updating it yourself using `cargo install XYZ` isn't really a challenge after that (just one extra command). 

Apart from that, the imported dependencies are very standard and relatively small, so it did not bother me much to personally import them. `serde_json` will also be really helpful in the future considering a lot of packages store their package configs as json.

InnocentZero:
One argument made against this could be that if you have multiple packages with feature flags installed in your system, you'd spawn multiple cargo commands from within topgrade to install them but the bottleneck is the compilation anyways if only cargo step is run and topgrade in general prompts users for manual intervention if multiple steps are being run.

SteveLauC:
Sorry for the late response, quite busy with my work on workdays.

> > @SteveLauC can you tell me what the failing test means? Because I cannot even understand what it is doing
> 
> No worries about it, the test coverage CI is simply broken

The CI has been fixed, you can rebase your branch.

-----

> Prior to that version, cargo would error out on reinstalling a package. Now it will cleanly exit and upgrade the package if needed. In that regard this is a massive improvement. See [this commit](https://github.com/rust-lang/cargo/commit/fb4415090f600bae51b0747bef2e7049070cd6ee) and this addition in cargo-update's [README](https://github.com/rust-lang/cargo/commit/fb4415090f600bae51b0747bef2e7049070cd6ee).

Thanks for showing me this! I tested this locally and it was indeed changed. Though the linked commit is relatively old, would you like to show me the exact commit that made this change to 1.77? I try to find it in the [CHANGELOG](https://github.com/rust-lang/cargo/blob/master/CHANGELOG.md#cargo-177-2024-03-21), but no luck.


-----

> Actually not! (kinda yes but also no). So cargo install-update -a command does not check the features of the packages it is updating, but it can do so individually if supplied with the appropriate flags (similar to cargo install xyz --features A,B). However to check those flags you have to read .crates2.json and NOT .crates.toml (which I suppose is being read by both this repo and upstream rn). But if you go all the way to read .crate2.json, updating it yourself using cargo install XYZ isn't really a challenge after that (just one extra command).

Well, I see this as a bug and it is a fair reason to me to implement this in Topgrade. So I am ok with the PR.

> There was already issues reporting this to `cargo-update`: https://github.com/nabijaczleweli/cargo-update/issues/251

-----

> Apart from that, the imported dependencies are very standard and relatively small, so it did not bother me much to personally import them. serde_json will also be really helpful in the future considering a lot of package managers store their package configs as json.

`serde_json` is light, it is ok to import it.

-----

> One argument made against this could be that if you have multiple packages with feature flags installed in your system, you'd spawn multiple cargo commands from within topgrade to install them but the bottleneck is the compilation anyways if only cargo step is run and topgrade in general prompts users for manual intervention if multiple steps are being run.

I think we should not spawn multiple child processes unless we find a good way to handle errors that happens in child processes. And, IMHO, Topgrade is not something that needs to be performant, so I think we can do the update sequentially.

SteveLauC:
One question I have in my mind, is the JSON defined in this `.crates2.json` file structured? If so, we can define some concrete helper types in Topgrade, and deserialize the file to the types rather than `serde_json::Value`.

SteveLauC:
BTW, looks like [cargo-update #251](https://github.com/nabijaczleweli/cargo-update/issues/251) has just been fixed, see comments of that issue for more information.

InnocentZero:
> BTW, looks like [cargo-update #251](https://github.com/nabijaczleweli/cargo-update/issues/251) has just been fixed, see comments of that issue for more information.

Oh, if that is the case, then idt we need to have this PR anymore since just using that would be better. I'd have tested the latest PR but I don't have any outdated binaries that need feature flags either.

SteveLauC:
> Oh, if that is the case, then idt we need to have this PR anymore since just using that would be better. 

So closing:)

glima:
Weird, because I'm pretty sure one of my packages got updated with less features (via topgrade) and I have:

❯ cargo install-update -V
cargo-install-update 14.0.0

SteveLauC:
> Weird, because I'm pretty sure one of my packages got updated with less features (via topgrade) and I have:

Hmm, that's weird, per the upstream issue, this should be fixed in cargo-install-update 14.0.0, if you can reproduce it, please report it to the upstream repo.

PR #797, opened 2024-05-12
feat: allow self renamer to work if tempdir is on a different device

davidkna:
## Standards checklist:

## For new steps
  the underlying command

Currently, the self renamer fails with the following error if the tempdir is on a different device from the system temporary directory:
```
Error:
   0: The system cannot move the file to a different disk drive. (os error 17)
```
This PR works around this by using a temporary file in the binary directory instead in that case.

### DIFF
--- a/src/self_renamer.rs
+++ b/src/self_renamer.rs
@@ -2,7 +2,7 @@
 
 use color_eyre::eyre::Result;
 use std::{env::current_exe, fs, path::PathBuf};
-use tracing::{debug, error};
+use tracing::{debug, error, warn};
 
 pub struct SelfRenamer {
     exe_path: PathBuf,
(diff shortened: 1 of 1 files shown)


SteveLauC (commented):
Hi, thanks for your contributing to Topgrade!

I wonder can we do `std::fs::copy(); std::fs::remove_file()` here?

davidkna:
> I wonder can we do `std::fs::copy(); std::fs::remove_file()` here?

It would be possible to use this, but `std::fs::rename()` is something of a atomic operation which helps avoid issues around concurrent access and possibly mangling files.

SteveLauC:
Yeah, that is one benefit of `rename()`.


May I step back a little bit, and ask you what is the functionality of this `SelfRenamer`, I have never used the self-update feature and I am not the author of that code.

From my understanding, the renamer itself is simply doing a round-trip, I do not quite get why it is necessary on Windows.

davidkna:
Executable cleanup is now implemented via the `self-replace` crate that was already a transitive dependency via `self_update`.

PR #798, opened 2024-05-13
Support rustup self update

alok:
## Standards checklist:

## For new steps
  the underlying command

### DIFF
--- a/src/steps/generic.rs
+++ b/src/steps/generic.rs
@@ -225,10 +225,10 @@ pub fn run_rustup(ctx: &ExecutionContext) -> Result<()> {
 
     print_separator("rustup");
     ctx.run_type()
-        .execute(rustup)
+        .execute(&rustup)
         .args(["self", "update"])
         .status_checked()?;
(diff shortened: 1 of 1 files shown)


src/steps/generic.rs
```diff
@@ -224,7 +224,11 @@ pub fn run_rustup(ctx: &ExecutionContext) -> Result<()> {
     let rustup = require("rustup")?;
 
     print_separator("rustup");
-    ctx.run_type().execute(rustup).arg("update").status_checked()
+    ctx.run_type()
```
SteveLauC, line 227: On my mac, `rustup` is installed through Homebew, executing this command results in an error:

```sh
$ rustup self update
error: self-update is disabled for this build of rustup
error: you should probably use your system package manager to update rustup
```


-----


I kinda think this feature is not necessary? if one installed rustup through the official script, then `rustup update` will update itself if the configuration is set to `enable`. And for folks that are using Nix or Homebrew(like me), they can do the update through the package manager.

Ref: https://rust-lang.github.io/rustup/basics.html#keeping-rustup-up-to-date

PR #813, opened 2024-05-23
add upgrade stuff for bedrock linuxmint strata

lachsdachs:
run the debian upgrade steps if there is a linuxmint bedrock stratum.


SteveLauC (approved):
Thanks!

PR #817, opened 2024-05-26
Add linux mint support

lachsdachs:
Detect linuxmint distros and treat them as debian. 

Question: how does ubuntu work here? 
Mint is a Ubuntu fork and i have seen linuxmint strata getting the Ubuntu treatment in the upgrade_bedrock section, but i have not found any code that handles Ubuntu distros. But topgrade surely has support for the most popular distro of all, doesn't it? is it okay to treat ubuntu distros as debian?

### DIFF
--- a/src/steps/os/linux.rs
+++ b/src/steps/os/linux.rs
@@ -73,7 +73,7 @@ impl Distribution {
 
             Some("nobara") => Distribution::Nobara,
             Some("void") => Distribution::Void,
-            Some("debian") | Some("pureos") | Some("Deepin") => Distribution::Debian,
+            Some("debian") | Some("pureos") | Some("Deepin") | Some("linuxmint") => Distribution::Debian,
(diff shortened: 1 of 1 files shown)


SteveLauC:
> Question: how does ubuntu work here?

It will be treated as Debian given it ID_LIKE field is `debian`: https://github.com/which-distro/os-release/blob/main/ubuntu/22.04

SteveLauC (approved):
Thanks! I am actually surprised that this wasn't already supported by Topgrade:<

lachsdachs:
> It will be treated as Debian given it ID_LIKE field is `debian`: https://github.com/which-distro/os-release/blob/main/ubuntu/22.04
> 

Does that mean that the distro detection logic is currently implemented in different places? Because that seems like a bad idea.

SteveLauC:
Even though Linux mint has its ID_LIKE set to ubuntu, it actually doesn't matter since it's unique ID is sufficient for detection.

PR #820, opened 2024-06-06
fix(RubyGems): support no-sudo updating for rbenv and rvm

yaroslav:
## What does this PR do

Currently, when updating RubyGems, the code defaults to sudo-ing it, unless there is `asdf` in the path. I've added checks for `rbenv` and `rvm` Ruby version managers to avoid using sudo.

The idea behind that piece of code may be faulty in the first place, and we should probably check the location or the file permissions of installed RubyGems instead of the pathname, but for a quick fix, should do.

## Standards checklist

## For new steps

  the underlying command


SteveLauC (approved):
Thanks!

Also thanks for your explanation of the implementation!

PR #822, opened 2024-06-12
Optimize GitHub Actions Workflow for Code Quality and Security

niStee:
Optimize GitHub Actions workflow for code quality and security

— Reuse checkout step from shared-setup job to reduce duplication
— Add caching for Rust toolchain to improve build times
— These changes should improve the overall efficiency and performance of the workflow without compromising the code quality and security checks

## Standards checklist

## For new steps

- [ ] *Optional:* Top grade skips this step where needed
  the underlying command

If you developed a feature or a bug fix for someone else, and you do not have the


.github/workflows/check_security_vulnerability.yml
```diff
@@ -30,3 +40,49 @@ jobs:
...
+          ref: ${{ needs.shared-setup.outputs.checkout_ref }}
+
+      - name: Cache Rust toolchain
+        uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
```
SteveLauC, line 63: `~/.rustup/` should be cached if we want to cache toolchain.

.github/workflows/check_security_vulnerability.yml
```diff
@@ -30,3 +40,49 @@ jobs:
...
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
```
SteveLauC, line 64: Emm, why are we using `hash(Cargo.lock)` as the cache key, this is generally for caching dependencies.

.github/workflows/check_security_vulnerability.yml
```diff
@@ -30,3 +40,49 @@ jobs:
...
+
+      - name: Install Rust toolchain
+        uses: actions-rs/toolchain@16499b5e05bf2e26879000db0c1d13f7e13fa3af #@v1
```
SteveLauC, line 67: This action has been archived, please use [`dtolnay/rust-toolchain`](https://github.com/dtolnay/rust-toolchain).

.github/workflows/check_security_vulnerability.yml
```diff
@@ -1,27 +1,37 @@
...
-name: Check Security Vulnerability
+name: Code Quality and Security
 
 on:
   pull_request:
   push:
     branches:
       - main
+  schedule:
+    - cron: '0 0 * * 0' # Run every Sunday at 00:00 (midnight)
```
SteveLauC, line 9: If we run this on every PR and push to main, do we still need to run it regularly?

.github/workflows/check_security_vulnerability.yml
```diff
@@ -1,27 +1,37 @@
-# This workflow uses actions that are not certified by GitHub.
-# They are provided by a third-party and are governed by
-# separate terms of service, privacy policy, and support
-# documentation.
-
-name: Check Security Vulnerability
```
SteveLauC, line 6: Nice cleanup!

.github/workflows/check_security_vulnerability.yml
```diff
@@ -1,27 +1,37 @@
...
+    name: Shared Setup
     runs-on: ubuntu-latest
+    outputs:
+      checkout_ref: ${{ steps.checkout.outputs.ref }}
+    steps:
+      - name: Checkout code
+        id: checkout
+        uses: actions/checkout@v4
+
+  devskim:
```
SteveLauC, line 23: This job needs a `runs-on` key.

.github/workflows/check_security_vulnerability.yml
```diff
@@ -30,3 +40,49 @@ jobs:
         uses: github/codeql-action/upload-sarif@v3
         with:
           sarif_file: devskim-results.sarif
+
+  rust-clippy:
```
SteveLauC, line 44: Same here, this job needs a `runs-on` key.

SteveLauC (changes requested):
Hi, sorry for the late reply! Hope it is not too late.

.github/workflows/check_security_vulnerability.yml
niStee: Using hashFiles('**/Cargo.lock'), the workflow generates a unique hash based on the contents of the Cargo.lock file. This hash is then used as the cache key. If the Cargo.lock file remains unchanged between workflow runs, the cache key will remain the same, indicating that the dependencies have not changed. This allows the workflow to skip the dependency installation step and use the cached dependencies instead, saving time and resources.
To handle platform-specific dependencies, we include the operating system in the cache key. This ensures that the cache is specific to the platform being used. This change ensures that the cache key is unique for each operating system, preventing problems that can arise from platform-specific dependencies.

.github/workflows/check_security_vulnerability.yml
```diff
@@ -1,27 +1,37 @@
...
-# documentation.
-
-name: Check Security Vulnerability
+name: Code Quality and Security
 
 on:
   pull_request:
   push:
     branches:
       - main
```
niStee, line 7: Running the job regularly, say once a week, can still be beneficial, even if it is already running on pull_request and pushing to the main branch. Here are some reasons why:

Dependency updates: Regular runs can catch vulnerabilities introduced by dependency updates, which may not be covered by code changes in pull requests or direct pushes.

Environment changes: Changes in the environment or external services may introduce new vulnerabilities or problems that regular runs can help to identify.

Missed changes: If there are changes made directly to the repository that bypass the usual pull request process, a scheduled run can ensure that these changes are also checked.

niStee:
> Hi, sorry for the late reply! Hope it is not too late.

Hi, no need to apologize.

.github/workflows/check_security_vulnerability.yml
SteveLauC: > Dependency updates: Regular runs can catch vulnerabilities introduced by dependency updates, which may not be covered by code changes in pull requests or direct pushes.

Emm, we record the used dependencies in `Cargo.lock` using full version number, they won't change without PR/direct push.

> Missed changes: If there are changes made directly to the repository that bypass the usual pull request process, a scheduled run can ensure that these changes are also checked.


Directly pushing to the main branch is barely used, so every change would be made through a PR.

--------

> Environment changes: Changes in the environment or external services may introduce new vulnerabilities or problems that regular runs can help to identify.

This makes sense to me.

.github/workflows/check_security_vulnerability.yml
```diff
@@ -1,32 +1,82 @@
...
+          rustup toolchain install stable
+          cargo install clippy-sarif sarif-fmt --force
+      - name: Run rust-clippy
+        run: |
+          cargo clippy --all-features --message-format=json | clippy-sarif | tee rust-clippy-results.sarif | sarif-fmt
+        continue-on-error: true
+      - name: Upload Clippy analysis results to GitHub
+        uses: github/codeql-action/upload-sarif@v3
+        with:
+          sarif_file: rust-clippy-results.sarif
```
SteveLauC, line 76: We are using the same file name for the sarif file on all the OSes, will they overwrite each other after uploading?

.github/workflows/check_security_vulnerability.yml
```diff
@@ -1,32 +1,82 @@
...
+      os: ${{ matrix.os }}
     steps:
       - name: Checkout code
+        id: checkout
         uses: actions/checkout@v4
 
+  devskim:
+    name: DevSkim Security Scan
+    needs: shared-setup
+    runs-on: ubuntu-latest
```
SteveLauC, line 33: I saw this was changed in the last commit, why this change?

.github/workflows/check_security_vulnerability.yml
```diff
@@ -1,32 +1,82 @@
...
-
       - name: Upload DevSkim scan results to GitHub Security tab
         uses: github/codeql-action/upload-sarif@v3
         with:
           sarif_file: devskim-results.sarif
+
+  rust-clippy:
+    name: Rust Clippy Analysis
+    needs: shared-setup
+    runs-on: ${{ needs.shared-setup.outputs.os }}
```
SteveLauC, line 49: From the CI output, this job was only executed on [Windows](https://github.com/topgrade-rs/topgrade/actions/runs/10127215929/job/28004716053), any idea?

> info: syncing channel updates for 'stable-x86_64-pc-windows-msvc'

.github/workflows/check_security_vulnerability.yml
SteveLauC: Right, this is eactly right.

.github/workflows/check_security_vulnerability.yml
niStee: Are you referring to https://github.com/topgrade-rs/topgrade/actions/runs/10402662730 ?

niStee: Or https://github.com/topgrade-rs/topgrade/actions/runs/10402662748 ?

.github/workflows/check_security_vulnerability.yml
niStee: I noticed devSkim scanner failed on macos and windows

.github/workflows/check_security_vulnerability.yml
SteveLauC: > Or https://github.com/topgrade-rs/topgrade/actions/runs/10402662748 ?

This and the CI referred in my comment both run the clippy only on Windows

SteveLauC: I guess

> ${{ needs.shared-setup.outputs.os }}

this is not working?

niStee: > > Or https://github.com/topgrade-rs/topgrade/actions/runs/10402662748 ?
> 
> This and the CI referred in my comment both run the clippy only on Windows

I do not see it. Do you have the direct link to the output or provide a screenshot?

SteveLauC: image

See the "install toolchain" step, it is running on Windows

niStee:
@SteveLauC  I am not sure if we want the DevSkim check to be --fail-on-vuln=true" or false? What do you suggest?

SteveLauC:
> @SteveLauC I am not sure if we want the OSV Scanner check to be --fail-on-vuln=true" or false? What do you suggest?

It is good to have vulnerability checked, but the reported vulnerabilities are all about Topgrade's indirect dependencies, so they are pretty hard to fix, we have to disable this option if we cannot find other workarounds.

niStee:
@SteveLauC, please review this pull request once more and tell me what needs to be changed before merging.

SteveLauC:
Ahh, sorry, I thought this was still WIP given the CI error, I will try to take a look tomorrow, I am too drowsy now and can barely think about things.

niStee:
> > @SteveLauC I am not sure if we want the OSV Scanner check to be --fail-on-vuln=true" or false? What do you suggest?
> 
> It is good to have vulnerability checked, but the reported vulnerabilities are all about Topgrade's indirect dependencies, so they are pretty hard to fix, we have to disable this option if we cannot find other workarounds.

Maybe we can make this an optional check for a pull request, so a maintainer can still push it to the main branch?

PR #839, opened 2024-06-30
feat: new step Lensfun's database update

SteveLauC:
## What does this PR do

This PR adds support for updating [Lensfun](https://github.com/lensfun/lensfun)'s lenses database via its cli tool `lensfun-update-data`. 

Per the [documentation](https://lensfun.github.io/manual/v0.3.2/lensfun-update-data.html), this update command can be invoked with or without `sudo` so that one can decide where to install the update. If `sudo` is used, then the update will be installed system-wide, available to all users. Otherwise, update will be install under user's home directory and thus exclusive to the user.

This option is controlled through the following configuration entry in Topgrade:

```toml
[lensfun]
# If disabled, Topgrade invokes `lensfun‑update‑data` without root priviledge, 
# then the update will be only available to you. Otherwise, `sudo` is required,
# and the update will be installed system-wide, i.e., available to all users.
# (default: false)
# use_sudo = false
```
 

## Standards checklist

## For new steps

It should work as this is a one command step.

  the underlying command

The `lensfun-update-data` command does not accept options.

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -244,3 +244,10 @@
 [containers]
 # Specify the containers to ignore while updating (Wildcard supported)
 # ignored_containers = ["ghcr.io/rancher-sandbox/rancher-desktop/rdx-proxy:latest", "docker.io*"]
+
+[lensfun]
+# If disabled, Topgrade invokes `lensfun‑update‑data` without root priviledge, 
+# then the update will be only available to you. Otherwise, `sudo` is required,
+# and the update will be installed system-wide, i.e., available to all users.
+# (default: false)
+# use_sudo = false
--- a/src/config.rs
+++ b/src/config.rs
@@ -101,6 +101,7 @@ pub enum Step {
     Helix,
     Krew,
     Lure,
+    Lensfun,
     Macports,
     Mamba,
     Miktex,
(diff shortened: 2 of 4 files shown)


SteveLauC:
cc @Tokariew, would you like to give this a test?

Tokariew:
It look good, for root/user when there is update, in case there isn't one the `lensfun-update-data` return `1` as exit status and topgrade return that lensfun failed
```
✗  ./topgrade --only lensfun

── 12:39:31 - Lensfun's database update ────────────────────────────────────────
Info: root privileges needed for updating the system database.
Info: updating user DB in '/home/tokariew/.local/share/lensfun/updates'
Reading http://lensfun.sourceforge.net/db/versions.json …
Reading http://wilson.bronger.org/lensfun-db/versions.json …
Info: No newer database was found for last installed Lensfun.
Lensfun's database update failed: 
   0: Command failed: `/usr/bin/lensfun-update-data`
   1: `/usr/bin/lensfun-update-data` failed: exit status: 1

Location:
   src/steps/generic.rs:994

── 12:39:33 - Summary ──────────────────────────────────────────────────────────
Lensfun's database update: FAILED
```

So not sure, how topgrade should behave in case there is no update.

SteveLauC:
Thanks for testing it!

Try the latest commit, it now takes 1 as a normal exit code:)

Tokariew:
seems good for me, in all 4 basic cases root/normal user update/no update

tested with no internet and when added to pihole blacklist, and it fail in this case, so it look fine
without lensfun installed it's not called.
Not sure what else i can test

SteveLauC:
Thanks for the test, let's merge this pr.

PR #851, opened 2024-07-07
improve fedora immutable detection

Book-reader:
## What does this PR do
Improves the detection for rpm-ostree based fedora versions by checking for the OSTREE_VERSION section in os-release
This allows detection of systems built by [blue-build](https://blue-build.org) that do not contain a VARIANT section
e.g: my [os-release](https://github.com/user-attachments/files/16117705/os-release.txt)

I have kept the old detection method as some older rpm-ostree based distributions do not include the OSTREE_VERSION
section (see [fedoraonyx](https://github.com/user-attachments/files/16117714/fedoraonyx.txt) in src/steps/os/os_release)

## Standards checklist

## For new steps

  the underlying command

### DIFF
--- a/src/steps/os/linux.rs
+++ b/src/steps/os/linux.rs
@@ -59,7 +59,9 @@ impl Distribution {
             Some("centos") | Some("rhel") | Some("ol") => Distribution::CentOS,
             Some("clear-linux-os") => Distribution::ClearLinux,
             Some("fedora") => {
-                return if let Some(variant) = variant {
+                return if section.get("OSTREE_VERSION").is_some() {
+                    Ok(Distribution::FedoraImmutable)
+                } else if let Some(variant) = variant {
                     match variant {
                         "Silverblue" | "Kinoite" | "Sericea" | "Onyx" | "IoT Edition" | "Sway Atomic" => {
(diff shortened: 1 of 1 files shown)


stevelauc-ghost (commented):
Hi, so IIUC, the distro built through blue-bird does not have an actual distro name, it is simply Fedora immutable?

src/steps/os/linux.rs
```diff
@@ -59,7 +59,9 @@ impl Distribution {
             Some("centos") | Some("rhel") | Some("ol") => Distribution::CentOS,
             Some("clear-linux-os") => Distribution::ClearLinux,
             Some("fedora") => {
-                return if let Some(variant) = variant {
+                return if section.get("OSTREE_VERSION").is_some() {
+                    Ok(Distribution::FedoraImmutable)
```
stevelauc-ghost, line 63: Would you mind add a comment here to state the reason for this `OSTREE_VERSION` check?

Book-reader:
> Hi, so IIUC, the distro built through blue-bird does not have an actual distro name, it is simply Fedora immutable?

On further inspection I missed a part of the installation and it is usually there, so this pr is not needed.

Issue #709, opened 2024-02-22
Topgrade cannot upgrade zshell framework (https://wiki.zshell.dev)


scrambler-crypto:
## Erroneous Behavior
Topgrade cannot upgrade zshell framework (https://wiki.zshell.dev)
the upgrade process starts correctly but after few seconds it hangs up.
The error message is the following:
```
The update took 8.44 seconds
zsh:1: warning: 10 jobs SIGHUPed
```

## Expected Behavior
Topgrade upgrade correctly the framework

## Steps to reproduce
install the zshell framework
launch `topgrade`

## Problem persists without calling from topgrade

- [ ] Yes
- [x] No

## Did you run topgrade through `Remote Execution`

- [ ] Yes
- [x] No

If yes, does the issue still occur when you run topgrade directly in your
remote host

- [ ] Yes
- [ ] No

## Configuration file (Optional)
(default file)

## Additional Details
- Operation System/Version
Ubuntu 23.10

- Installation
topgrade installed with `cargo install`

- Topgrade version (`topgrade -V`)
❯ topgrade -V
Topgrade 14.0.1

## Verbose Output (`topgrade -v`)

<details>
<pre>
DEBUG Executing command `/usr/bin/zsh -i -c 'source /home/davide/.zshrc && zi self-update && zi update --all -p'`
Updating »»»» ❮ Zi ❯ …
From https://github.com/z-shell/zi
 * branch            main       -> FETCH_HEAD
Already up to date.
Compiling »»» ❮ Zi ❯ …
Reloading »»» ❮ Zi ❯ …
Initiating parallel update …
Spawning the next 15 concurrent update jobs: snippets …
Spawning the next 15 concurrent update jobs: plugins …
Spawning the next 15 concurrent update jobs: plugins …
The update took 6.13 seconds
zsh:1: warning: 10 jobs SIGHUPed
DEBUG Command failed: Err(
   0: Command failed: `/usr/bin/zsh -i -c 'source /home/davide/.zshrc && zi self-update && zi update --all -p'`
   1: `/usr/bin/zsh` failed: exit status: 1

Location:
   /home/davide/.cargo/registry/src/index.crates.io-6f17d22bba15001f/topgrade-14.0.1/src/steps/zsh.rs:144)
DEBUG Step "zi" failed:
   0: Command failed: `/usr/bin/zsh -i -c 'source /home/davide/.zshrc && zi self-update && zi update --all -p'`
[...]

PR #857, opened 2024-07-08
update zi and zinit update to use parallel flag

taymoork2:
## What does this PR do

Speeds up zi and zinit update commands by adding parallel flag

## Standards checklist

- [ ] The PR title is descriptive.
 
## For new steps

  the underlying command

### DIFF
--- a/src/steps/zsh.rs
+++ b/src/steps/zsh.rs
@@ -124,7 +124,7 @@ pub fn run_zinit(ctx: &ExecutionContext) -> Result<()> {
 
     print_separator("zinit");
 
-    let cmd = format!("source {} && zinit self-update && zinit update --all", zshrc.display());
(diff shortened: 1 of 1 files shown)


SteveLauC (commented):
Hi, we already tried this, and it doe not seem to work, see: https://github.com/topgrade-rs/topgrade/issues/709

taymoork2:
Is there a way to test this change with the config file before I install rust and debug that way
Dont think theres a way to patch the currently binary installed w/ brew?

SteveLauC:
> Is there a way to test this change with the config file before I install rust and debug that way Dont think theres a way to patch the currently binary installed w/ brew?

I guess no, you need to build it from src.

SteveLauC:
BTW, if you want to test this parallel option, try Topgrade v14.0.0 or v14.0.1, it was added in v14.0.0 and removed in v15.0.0.

PR #878, opened 2024-07-28
Update SECURITY.md

niStee:
## Standards checklist

## For new steps

  the underlying command

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -6,6 +6,6 @@ We only support the latest major version and each subversion.
 
 | Version  | Supported          |
 | -------- | ------------------ |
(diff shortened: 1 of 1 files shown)


SteveLauC (approved):
Have never noticed this file before, thanks for updating it!

niStee:
> Have never noticed this file before, thanks for updating it!

Maybe there is a way to automate this change according to the release?

SteveLauC:
> Maybe there is a way to automate this change according to the release?

I haven't tried to modify the repo file in the CI, but I think we can add this to the [release procedure](https://github.com/topgrade-rs/topgrade/blob/main/RELEASE_PROCEDURE.md)

PR #881, opened 2024-07-30
Allow Nix unfree packages to be upgraded

### LINKED ISSUE #611: Non-free nix packages with Flakes are failing to upgrade
This is related to #389, with a twist.

## Steps to reproduce
1. Install Nix on macOS, with the [Determinate Systems](https://zero-to-nix.com/start/install) installer. Using the "official" installer would work, but with more steps (enable Flakes, disable nixpkgs) and I'm not an expert on that 😬 
2. Install an unfree flake: `NIXPKGS_ALLOW_UNFREE=1 nix profile install --impure 'nixpkgs#_1password'`
3. Run `topgrade`

## Erroneous Behavior
topgrade fails with the following error:

```console
MacPro% env|grep ALLOW
NIXPKGS_ALLOW_UNFREE=1

MacPro% nix --extra-experimental-features nix-command profile upgrade '.*' --verbose

       error: Package ‘1password-cli-2.22.0’ in /nix/store/mrckm9b3v6wg2fljdp3z4pja1sz07g3z-source/pkgs/applications/misc/1password/default.nix:71 has an unfree license (‘unfree’), refusing to evaluate.

       a) To temporarily allow unfree packages, you can use an environment variable
          for a single invocation of the nix tools.

            $ export NIXPKGS_ALLOW_UNFREE=1

          Note: When using `nix shell`, `nix build`, `nix develop`, etc with a flake,
                then pass `--impure` in order to allow use of environment variables.

       b) For `nixos-rebuild` you can set
         { nixpkgs.config.allowUnfree = true; }
       in configuration.nix to override this.

       Alternatively you can configure a predicate to allow specific packages:
         { nixpkgs.config.allowUnfreePredicate = pkg: builtins.elem (lib.getName pkg) [
             "1password-cli"
           ];
         }

       c) For `nix-env`, `nix-build`, `nix-shell` or any other Nix command you can add
         { allowUnfree = true; }
       to ~/.config/nixpkgs/config.nix.
```

Options `b`/`c` do not work, because of Flakes (from what I understand) - and the env.var alone isn't enough, `--impure` must be passed.

## Expected Behavior
One of:
* Documentation to explain why this doesn't work - which I'm trying to achieve with this ticket.
* The command used by topgrade should work out of the box.
* Topgrade should allow setting `--impure`.
* Topgrade should automagically add `--impure` when `NIX_*` environment variables are present?

- [x] Yes, but with a workaround
- [ ] No

## Additional Details
- Operation System/Version: MacOS Sonoma
- Topgrade version (`topgrade -V`): Topgrade 13.0.0, installed with Nix

## Relevant code

https://github.com/topgrade-rs/topgrade/blob/78dec892cf97013a1123900dcb23cfc2365961e7/src/steps/os/unix.rs#L397

dashmoho:
## What does this PR do
Allow unfree packages to be upgraded

Fixes #611.

## Standards checklist

## For new steps

  the underlying command

### DIFF
--- a/src/steps/os/unix.rs
+++ b/src/steps/os/unix.rs
@@ -413,10 +413,10 @@ pub fn run_nix(ctx: &ExecutionContext) -> Result<()> {
 
     debug!("Nix version: {:?}", version);
 
-    let mut packages = "--all";
+    let mut packages: Vec<&str> = vec!["--all", "--impure"];
 
     if !matches!(version, Ok(version) if version >= Version::new(2, 21, 0)) {
-        packages = ".*";
+        packages = vec![".*"];
     }
 
     if Path::new(&manifest_json_path).exists() {
@@ -425,7 +425,7 @@ pub fn run_nix(ctx: &ExecutionContext) -> Result<()> {
             .args(nix_args())
             .arg("profile")
             .arg("upgrade")
-            .arg(packages)
+            .args(&packages)
             .arg("--verbose")
             .status_checked()
     } else {


SteveLauC (approved):
Thanks for the fix!

PR #883, opened 2024-08-01
feat: add volta packages

diogonr:
## What does this PR do

Currently there's no volta upgrade command, so this implementation has to list all installed packages first, and update them separately

Example output: 
CleanShot 2024-08-01 at 09 51 27@2x

## Standards checklist

## For new steps

  the underlying command

### DIFF
--- a/src/steps/node.rs
+++ b/src/steps/node.rs
@@ -13,7 +13,7 @@ use semver::Version;
 use tracing::debug;
 
 use crate::command::CommandExt;
-use crate::terminal::print_separator;
+use crate::terminal::{print_info, print_separator};
 use crate::utils::{require, PathExt};
 use crate::{error::SkipStep, execution_context::ExecutionContext};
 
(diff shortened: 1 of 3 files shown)


src/steps/node.rs
```diff
@@ -276,3 +276,43 @@ pub fn deno_upgrade(ctx: &ExecutionContext) -> Result<()> {
...
+
+/// There is no `volta upgrade` command, so we need to upgrade each package
+pub fn run_volta_packages_upgrade(ctx: &ExecutionContext) -> Result<()> {
+    let volta = require("volta")?;
+
+    print_separator("Volta");
+
+    let list_output = ctx
+        .run_type()
+        .execute(volta.clone())
```
SteveLauC, line 288: You can pass it by reference to avoid the clone:

```suggestion
        .execute(&volta)
```

src/steps/node.rs
```diff
@@ -276,3 +276,43 @@ pub fn deno_upgrade(ctx: &ExecutionContext) -> Result<()> {
...
+
+    let installed_packages: Vec<String> = list_output
+        .lines()
+        .filter_map(|line| {
+            // format is 'kind package@version ...'
+            let mut parts = line.split_whitespace();
+            parts.next();
+            let package_part = parts.next()?;
+            let version_index = package_part.rfind('@').unwrap_or(package_part.len());
+            Some(package_part[..version_index].trim().to_string())
```
SteveLauC, line 301: `.trim()` still returns a `&str`, so no need to allocate:

```suggestion
    let installed_packages: Vec<&str> = list_output
        .lines()
        .filter_map(|line| {
            // format is 'kind package@version ...'
            let mut parts = line.split_whitespace();
            parts.next();
            let package_part = parts.next()?;
            let version_index = package_part.rfind('@').unwrap_or(package_part.len());
            Some(package_part[..version_index].trim())
```

src/steps/node.rs
```diff
@@ -276,3 +276,43 @@ pub fn deno_upgrade(ctx: &ExecutionContext) -> Result<()> {
...
+        .collect();
+
+    if installed_packages.is_empty() {
+        print_info("No packages installed with Volta");
+        return Ok(());
+    }
+
+    for package in installed_packages.iter() {
+        ctx.run_type()
+            .execute(volta.clone())
```
SteveLauC, line 312:
```suggestion
            .execute(&volta)
```

src/steps/node.rs
```diff
@@ -276,3 +276,43 @@ pub fn deno_upgrade(ctx: &ExecutionContext) -> Result<()> {
     print_separator("Deno");
     ctx.run_type().execute(&deno).arg("upgrade").status_checked()
 }
+
+/// There is no `volta upgrade` command, so we need to upgrade each package
+pub fn run_volta_packages_upgrade(ctx: &ExecutionContext) -> Result<()> {
+    let volta = require("volta")?;
+
+    print_separator("Volta");
+
```
SteveLauC, line 285: We need to support the `--dry-run` option, so should check it here:

```suggestion

    if ctx.run_type().dry() {
        print_info("Updating Volta packages...");
        return Ok(());
    }
```

SteveLauC (changes requested):
Thanks for contributing to Topgrade, I have left some comments:)

diogonr:
Thank for the feedback, i'm clueless in rust... Suggestions applied

SteveLauC (approved):
Thanks!

PR #889, opened 2024-08-19
Add aqua tool installer cli

boris-smidt-klarrio:
## What does this PR do
It adds the Aqua cli installer. 
https://aquaproj.github.io/

## Standards checklist

## For new steps

  the underlying command

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -300,7 +300,6 @@ fn run() -> Result<()> {
         runner.execute(Step::HomeManager, "home-manager", || unix::run_home_manager(&ctx))?;
         runner.execute(Step::Asdf, "asdf", || unix::run_asdf(&ctx))?;
         runner.execute(Step::Mise, "mise", || unix::run_mise(&ctx))?;
-        runner.execute(Step::Aqua, "aqua", || unix::run_aqua(&ctx))?;
         runner.execute(Step::Pkgin, "pkgin", || unix::run_pkgin(&ctx))?;
         runner.execute(Step::Bun, "bun", || unix::run_bun(&ctx))?;
         runner.execute(Step::BunPackages, "bun-packages", || unix::run_bun_packages(&ctx))?;
@@ -421,6 +420,7 @@ fn run() -> Result<()> {
     })?;
     runner.execute(Step::Poetry, "Poetry", || generic::run_poetry(&ctx))?;
     runner.execute(Step::Zvm, "ZVM", || generic::run_zvm(&ctx))?;
+    runner.execute(Step::Aqua, "aqua", || generic::run_aqua(&ctx))?;
 
     if should_run_powershell {
         runner.execute(Step::Powershell, "Powershell Modules Update", || {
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -300,6 +300,7 @@ fn run() -> Result<()> {
         runner.execute(Step::HomeManager, "home-manager", || unix::run_home_manager(&ctx))?;
         runner.execute(Step::Asdf, "asdf", || unix::run_asdf(&ctx))?;
         runner.execute(Step::Mise, "mise", || unix::run_mise(&ctx))?;
+        runner.execute(Step::Aqua, "aqua", || unix::run_aqua(&ctx))?;
```
SteveLauC, line 303: Why limit this to UNIX, looks like Windows is also supported: https://aquaproj.github.io/docs/install

src/steps/os/unix.rs
```diff
@@ -749,6 +749,16 @@ pub fn run_maza(ctx: &ExecutionContext) -> Result<()> {
     ctx.run_type().execute(maza).arg("update").status_checked()
 }
 
+pub fn run_aqua(ctx: &ExecutionContext) -> Result<()> {
```
SteveLauC, line 752: Have you tried the dry run option, does it work with this new step?

```sh
$ topgrade --only aqua --dry-run
```

Also the `--yes` option, does Aqua have this option or something similar?

>  Optional: The --dry-run option works with this step
 Optional: The --yes option works with this step if it is supported by
the underlying command

SteveLauC (changes requested):

src/steps/os/unix.rs
boris-smidt-klarrio: There isn't any -y or --yes on the update commands, So i don't think there is a dry run available.

src/main.rs
boris-smidt-klarrio: I've moved it to generic.rs

src/steps/os/unix.rs
SteveLauC: > There isn't any -y or --yes on the update commands

Get it, then we don't need to worry about this.

Dry run is another option, which is not related to --yes, when it is passed to Topgrade, Topgrade should simply echo what will be done and exit, does this PR implement it?

boris-smidt-klarrio: There isn't a dry-run option either.

SteveLauC: So 

```sh
$ topgrade --only aqua --dry-run
```

would actually do the update, right? We need to emulate this `--dry-run` option on the Topgrade's side, see:

https://github.com/topgrade-rs/topgrade/blob/a2375b48203322e39d36a6235db6aabdc4d3e5e0/src/self_update.rs#L20-L23

for an example.

SteveLauC (approved):
Thanks!

boris-smidt-klarrio:
Thank you for the fast response it was a pleasure to contribute to topgrade!

PR #890, opened 2024-08-23
Adds `uv` step

lucaspar:
This `uv` step runs global update tasks for the relatively new [`uv`](https://github.com/astral-sh/uv) python package and project manager.

On update, two actions are executed:

1. `uv self update` [docs](https://docs.astral.sh/uv/getting-started/installation/#standalone-installer)
2. `uv tool upgrade --all` [docs](https://docs.astral.sh/uv/guides/tools/#installing-tools)

## Standards checklist:

    - `cargo run -- --verbose --only uv` works.
 
## For new steps

  the underlying command
    - underlying command does not support `--yes`.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -327,6 +327,7 @@ fn run() -> Result<()> {
             unix::upgrade_gnome_extensions(&ctx)
         })?;
         runner.execute(Step::Pyenv, "pyenv", || unix::run_pyenv(&ctx))?;
+        runner.execute(Step::Uv, "uv", || unix::run_uv(&ctx))?;
         runner.execute(Step::Sdkman, "SDKMAN!", || unix::run_sdkman(&ctx))?;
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -327,6 +327,7 @@ fn run() -> Result<()> {
             unix::upgrade_gnome_extensions(&ctx)
         })?;
         runner.execute(Step::Pyenv, "pyenv", || unix::run_pyenv(&ctx))?;
+        runner.execute(Step::Uv, "uv", || unix::run_uv(&ctx))?;
```
SteveLauC, line 330: Looks like this tool is also available on Windows, we should probably move the impl to `general.rs`👀

SteveLauC (approved):
LGTM, just one tiny issue. 😁

lucaspar:
@SteveLauC I think that should do it.

SteveLauC (approved):
Thanks!

PR #896, opened 2024-08-31
Containers step: new runtime option to configuration

lucaspar:
## What does this PR do

+ Adds a `containers.runtime` option to the TOML configuration, letting the user select what used to be an implicit choice.
+ Defaults to `docker`, as it's the most commonly used runtime.

Open to suggestions.

## Standards checklist

### DIFF
--- a/.github/workflows/release_to_aur.yml
+++ b/.github/workflows/release_to_aur.yml
@@ -14,8 +14,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Publish AUR package
-        uses: ATiltedTree/create-aur-release@v1
+        uses: aksh1618/update-aur-package@v1.0.5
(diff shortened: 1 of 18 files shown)


SteveLauC (approved):
LGTM, but this is a breaking change when a user has both Podman and Docker installed, before this PR, Podman will be used, now the default option becomes Docker. We need to document this in [BREAKINGCHANGES_dev.md](https://github.com/topgrade-rs/topgrade/blob/main/BREAKINGCHANGES_dev.md).

lucaspar:
Right - this is actually my situation and what motivated the change. Most of my containers are in fact under docker and they couldn't be updated with topgrade before. I'll document it soon.

SteveLauC (approved):
Thanks!

PR #901, opened 2024-09-05
feat: tmux session attach mode

wetfloo:
## What does this PR do
This will allow the user to choose between just creating a session when in tmux, or switching the client to the newly created session. It's useful for my workflow revolving around tmux, where having forced a separate session with no quick way to attach to it created friction

## Standards checklist

## For new steps

  the underlying command

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -47,6 +47,12 @@
 # Run inside tmux (default: false)
 # run_in_tmux = true
 
+# Changes the way topgrade interacts with
+# the newly crated tmux session, creating the session
+# and only attaching to it if not inside tmux
+# (default: create, allowed values: "create", "create_and_switch_client")
+# tmux_session_attach_mode = "create"
+
 # Cleanup temporary or old files (default: false)
 # cleanup = true
 
--- a/src/config.rs
+++ b/src/config.rs
@@ -403,6 +403,8 @@ pub struct Misc {
 
     run_in_tmux: Option<bool>,
 
+    tmux_session_attach_mode: Option<TmuxSessionAttachMode>,
+
     cleanup: Option<bool>,
 
     notify_each_step: Option<bool>,
@@ -417,6 +419,19 @@ pub struct Misc {
     no_self_update: Option<bool>,
 
     log_filters: Option<Vec<String>>,
+}
+
+#[derive(Clone, Copy, Debug, Deserialize, ValueEnum)]
+#[clap(rename_all = "snake_case")]
+#[serde(rename_all = "snake_case")]
+pub enum TmuxSessionAttachMode {
+    Create,
+    CreateAndSwitchClient,
+}
+
+pub struct TmuxConfig {
+    pub args: Vec<String>,
+    pub session_attach_mode: TmuxSessionAttachMode,
 }
 
 #[derive(Deserialize, Default, Debug, Merge)]
@@ -710,6 +725,10 @@ pub struct CommandLineArgs {
     #[arg(short = 't', long = "tmux")]
     run_in_tmux: bool,
 
+    /// The preferred way to run the new tmux session
+    #[arg(long = "tmux-session-attach-mode")]
+    tmux_session_attach_mode: Option<TmuxSessionAttachMode>,
+
     /// Cleanup temporary or old files
     #[arg(short = 'c', long = "cleanup")]
     cleanup: bool,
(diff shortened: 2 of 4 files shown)


SteveLauC:
Hi, sorry for the late reply!


Would you mind telling me the difference between `tmux attach-session` and `tmux switch-session`, I tried it locally, seems like after running the command, I just "attached" to the newly-created `topgrade` session🤔

wetfloo:
Thanks for the reply! Basically, here's how it works
if the mode is `create_and_switch_client`, no matter if we're in- or outside of tmux, it will always attach to the new session
if the mode is `create`, it will only attach to the new session when outside of tmux, just like before (that's why it's picked as a default)

Here's a video that, hopefully, demonstrates that.

https://github.com/user-attachments/assets/0b6cee62-aef2-4aea-a90c-aa8e694e9245

SteveLauC:
> If the mode is create_and_switch_client, no matter if we're in- or outside of tmux, it will always attach to the new session

I get it, thanks for the explanation and the demo video:)

config.example.toml
```diff
@@ -47,6 +47,12 @@
 # Run inside tmux (default: false)
 # run_in_tmux = true
 
+# Changes the way topgrade interacts with
+# the newly crated tmux session, creating the session
+# and only attaching to it if not inside tmux
+# (default: create, allowed values: "create", "create_and_switch_client")
+# tmux_session_attach_mode = "create"
```
SteveLauC, line 54: I am thinking about the naming, what about we call them:

1. AttachIfNotInASession
2. AlwasyAttach

I think `switch_client` is too technical, thoughts on this?

src/config.rs
```diff
@@ -710,6 +725,10 @@ pub struct CommandLineArgs {
     #[arg(short = 't', long = "tmux")]
     run_in_tmux: bool,
 
+    /// The preferred way to run the new tmux session
+    #[arg(long = "tmux-session-attach-mode")]
+    tmux_session_attach_mode: Option<TmuxSessionAttachMode>,
```
SteveLauC, line 730: Is it necessary to add this CLI option, I think this feature is more like a static feature, one won't need to frequently change it 🤔

src/config.rs
```diff
@@ -967,6 +986,19 @@ impl Config {
                 .unwrap_or(false)
     }
 
+    /// The preferred way to run the new tmux session.
+    pub fn tmux_session_attach_mode(&self) -> TmuxSessionAttachMode {
```
SteveLauC, line 990: We should make this function private as well if we want developers to use `.tmux_config()` function

src/steps/tmux.rs
```diff
@@ -144,25 +146,32 @@ pub fn run_in_tmux(args: Vec<String>) -> Result<()> {
...
+            println!("Topgrade launched in a new tmux session");
+            return Ok(());
+        }
+        TmuxSessionAttachMode::CreateAndSwitchClient if is_inside_tmux => {
+            tmux.build().args(["switch-client", "-t", &session]).exec()
+        }
+        TmuxSessionAttachMode::Create | TmuxSessionAttachMode::CreateAndSwitchClient => {
+            tmux.build().args(["attach-session", "-t", &session]).exec()
+        }
+    };
```
SteveLauC, line 169: What about writing it in this way, would it be a little bit clearer?

```rs
TmuxSessionAttachModule::Create => {
    if is_inside_tmux {
        println!("Topgrade launched in a new tmux session");
        return Ok(());
    } else {
        tmux.build().args(["attach-session", "-t", &session]).exec()
    }
}
TmuxSessionAttachModule::CreateAndSwitchClient => {
     if is_inside_tmux {
        tmux.build().args(["switch-client", "-t", &session]).exec()
     } else {
        tmux.build().args(["attach-session", "-t", &session]).exec()
     }
}
```

SteveLauC (commented):
General impl looks good, left some comments:)

src/steps/tmux.rs
wetfloo: I can do it if it's easier, just felt bad to repeat `attach-session` call twice

SteveLauC: > just felt bad to repeat attach-session call twice

Yeah, that's indeed one drawback. Though I prefer readability, which will make maintenance easier:)

src/config.rs
wetfloo: done

src/config.rs
wetfloo: removed

config.example.toml
wetfloo: I changed some names, also dropping `attach` infix from tmux params, to not repeat it

SteveLauC (approved):
LGTM, thanks for implementing the feature!

PR #904, opened 2024-09-11
feat(brew): use sudo if Homebrew owned by another user on Linux

tomaszn:
NOTE: This is my first contact with Rust, and the implementation is surely suboptimal.

## What does this PR do

Currently, on Linux, if the Homebrew installation directory is owned by a different user, the step fails.

Homebrew directory, which on Linux is `/home/linuxbrew/.linuxbrew/`, is typically owned by the user who installed Homebrew, or a dedicated user account.

This change ensures that Homebrew updates can proceed smoothly even when its directory ownership does not match the current user's UID. It accomplishes this by running the update command with `sudo`, changing the UID to the proper owner account. Appropriate sudo configuration is assumed for this to work properly, for example:
```
%wheel  ALL=(linuxbrew) CWD=/tmp NOPASSWD: /home/linuxbrew/.linuxbrew/Homebrew/bin/brew
```

If this feature is acceptable, I am happy to work on the implementation further. Probably, the existing Sudo class would need extending to support providing the UID, to be reused when constructing the command. Any constructive criticism is highly welcome.

## Standards checklist

### DIFF
--- a/src/steps/os/unix.rs
+++ b/src/steps/os/unix.rs
@@ -13,6 +13,8 @@ use color_eyre::eyre::Context;
 use color_eyre::eyre::Result;
 use home;
 use ini::Ini;
+#[cfg(target_os = "linux")]
+use nix::unistd::Uid;
 use semver::Version;
 use tracing::debug;
 
@@ -270,6 +272,23 @@ pub fn upgrade_gnome_extensions(ctx: &ExecutionContext) -> Result<()> {
         .status_checked()
 }
 
+#[cfg(target_os = "linux")]
+pub fn brew_linux_sudo_uid() -> Option<u32> {
+    let linuxbrew_directory = "/home/linuxbrew/.linuxbrew";
+    if let Ok(metadata) = std::fs::metadata(linuxbrew_directory) {
+        let owner_id = metadata.uid();
+        let current_id = Uid::effective();
+        // print debug these two values
+        debug!("linuxbrew_directory owner_id: {}, current_id: {}", owner_id, current_id);
+        return if owner_id == current_id.as_raw() {
+            None // no need for sudo if linuxbrew is owned by the current user
+        } else {
+            Some(owner_id) // otherwise use sudo to run brew as the owner
+        };
+    }
+    None
+}
+
 #[cfg(any(target_os = "linux", target_os = "macos"))]
 pub fn run_brew_formula(ctx: &ExecutionContext, variant: BrewVariant) -> Result<()> {
     #[allow(unused_variables)]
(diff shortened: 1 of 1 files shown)


SteveLauC:
Hi, have you checked this [answer](https://stackoverflow.com/a/55021458/14092446) of question [How to use Homebrew on a Multi-user MacOS Sierra Setup](https://stackoverflow.com/questions/41840479/how-to-use-homebrew-on-a-multi-user-macos-sierra-setup), looks like the officially suggested way to share Homebrew with multiple users is to install the core globally and then install a clone locally, so that `sudo` can be avoided🤔

tomaszn:
Hi, thanks for joining me here, and pointing out an alternative approach. 

Unfortunately, it uses nonstandard "prefix" directories, which causes recompilation of the packages in the best case, and failure to install in the worst case. There is a comment about such issues: https://stackoverflow.com/questions/41840479/how-to-use-homebrew-on-a-multi-user-macos-sierra-setup/55021458#comment133690270_55021458

This is also mentioned in the FAQ: https://docs.brew.sh/FAQ#why-should-i-install-homebrew-in-the-default-location

When it comes to using sudo, there is an FAQ entry: https://docs.brew.sh/FAQ#why-does-homebrew-say-sudo-is-bad

It implicitly refers to running Homebrew as root, not to any sudo usage. The last sentence suggests creating a dedicated role account and using sudo to switch. Or maybe my interpretation is wrong...

SteveLauC:
Hi, thanks for the detailed explanation! So IIUC, sharing the same insatllaction of Homebrew with multiple users is basically unsupported and not suggested?

tomaszn:
Yes, that's what I understand.
I set up Homebrew on Linux with a separate "linuxbrew" account, and with this sudo configuration snippet, brew seems to work properly for several accounts.

SteveLauC:
Get it, the idea makes sense to me.

src/steps/os/unix.rs
```diff
@@ -282,6 +301,25 @@ pub fn run_brew_formula(ctx: &ExecutionContext, variant: BrewVariant) -> Result<
         }
     }
 
+    #[cfg(target_os = "linux")]
+    {
+        let sudo_uid = brew_linux_sudo_uid();
+        // if brew is owned by another user, execute "sudo -Hu <uid> brew update"
+        if let Some(uid) = sudo_uid {
+            let sudo_uid_str = sudo_uid.map_or_else(|| "".to_string(), |uid| uid.to_string());
+            print_separator(format!("{} (sudo as {})", variant.step_title(), sudo_uid_str));
```
SteveLauC, line 310: Let's use the user name rather than ID here:

```suggestion
            let user_name = nix::unistd::User::from_uid(uid)
                .expect("failed to call getpwuid()")
                .expect("this user should exist");
            print_separator(format!("{} (sudo as {})", variant.step_title(), user_name));
```

src/steps/os/unix.rs
```diff
@@ -282,6 +301,25 @@ pub fn run_brew_formula(ctx: &ExecutionContext, variant: BrewVariant) -> Result<
...
+            Command::new("sudo")
```
SteveLauC, line 311: We can use the `sudo` from `ctx`:

```suggestion
             let sudo = crate::util::require_option(ctx.sudo().as_ref(), "sudo is needed to run the update".into())?;
            Command::new(sudo)
```

src/steps/os/unix.rs
```diff
@@ -282,6 +301,25 @@ pub fn run_brew_formula(ctx: &ExecutionContext, variant: BrewVariant) -> Result<
...
+                .current_dir("/tmp") // brew needs a writable current directory
+                .args([
+                    "--set-home",
+                    &format!("--user=#{}", &uid.to_string()),
+                    "/home/linuxbrew/.linuxbrew/bin/brew",
```
SteveLauC, line 316: We can use that `binary_name` variable here, though we need to convert it to a `&str` first as `.args()` takes an array, all the elements within the same array should have the same type

src/steps/os/unix.rs
SteveLauC: BTW, can we use the `ctx.run_type().execute(sudo)` instead of `Command::new()` here?

SteveLauC:
> If this feature is acceptable, I am happy to work on the implementation further. Probably, the existing Sudo class would need extending to support providing the UID, to be reused when constructing the command.

The current impl looks generally good, so I think there is currently no need to do this. But we can do this when we have more steps that need this feature.


------

Have you tried if Topgrade's `--dry-run` option works with this new impl?

tomaszn:
Thanks for reviewing this.
I applied all the changes which you recommended, please take a look at your convenience.

Dry run seems to be working:
```
── 13:53:19 - Brew (sudo as user 'linuxbrew') ──────────────────────────────────────
Dry running: /usr/bin/sudo --set-home '--user=linuxbrew' /home/linuxbrew/.linuxbrew/Homebrew/bin/brew update in /tmp
```

SteveLauC (approved):
Looks good, thanks for the patch!

PR #923, opened 2024-10-04
 [FIXES #922] properly check for -CURRENT in OpenBSD steps and pass the correct flags to the respective commands

### LINKED ISSUE #922: Update OpenBSD step to incorporate checking if on -CURRENT so we can pass the correct flags to sysupgrade(8) and pkg_add(1)
# Update OpenBSD step to incorporate checking if on `-CURRENT` so we can pass the correct flags to `sysupgrade(8)` and `pkg_add(1)`

## Which tool is this about? 

Topgrade OpenBSD step

## Which operating systems are supported by this tool?

OpenBSD-STABLE (& OpenBSD-CURRENT)

## What should Topgrade do to figure out if the tool needs to be invoked?

Pull from `uname -r` to check if suffixed by `-current`

## Which exact commands should Topgrade run?

If `uname -r` has `-current` suffix:
- `pkg_add -Dsnap -u` (for pkg)
- `sysupgrade -sn` (for system)

## Does it have a `--dry-run` option? i.e., print what should be done and exit.

No.

## Does it need the user to confirm the execution? And does it provide a `--yes` option to skip this step?

No.

## More information

I'm self-assigning this. I'll send a PR once finished testing.

Izder456:
## What does this PR do

See issue #922

## Standards checklist

- [*] The PR title is descriptive.
- [*] I have read `CONTRIBUTING.md`
- [*] *Optional:* I have tested the code myself


Izder456:
Sorry, rustfmt is borked on my OpenBSD install, had to force `--amend` to fix that.

src/steps/os/openbsd.rs
```diff
@@ -1,16 +1,31 @@
...
 pub fn upgrade_openbsd(ctx: &ExecutionContext) -> Result<()> {
     let sudo = require_option(ctx.sudo().as_ref(), get_require_sudo_string())?;
     print_separator(t!("OpenBSD Update"));
-    ctx.run_type()
-        .execute(sudo)
-        .args(["/usr/sbin/sysupgrade", "-n"])
-        .status_checked()
+
+    let mut args = vec!["/usr/sbin/sysupgrade", "-n"];
+    if is_openbsd_current() {
```
SteveLauC, line 24: This function uses std::process::Command and Topgrade's dry-run option does not work with it, and thus this PR breaks the dry-run option.

We should use ctx.run_type().execute() instead, though I kinda think it won't work either. 😮‍💨

I guess we should skip all the update procedures and print something like "will update the system package" if this is a dry run(ctx has a dry_run() function to check if the current run is a dry-run or not)

src/steps/os/openbsd.rs
```diff
@@ -24,10 +39,12 @@ pub fn upgrade_packages(ctx: &ExecutionContext) -> Result<()> {
             .status_checked()?;
     }
 
-    ctx.run_type()
-        .execute(sudo)
-        .args(["/usr/sbin/pkg_add", "-u"])
-        .status_checked()?;
+    let mut args = vec!["/usr/sbin/pkg_add", "-u"];
+    if is_openbsd_current() {
+        args.push("-Dsnap");
```
SteveLauC, line 44: Same here

SteveLauC (commented):
Hi, thanks for sending the patch, left some comments

SteveLauC:
BTW, CI fixed in the main branch, you can rebase your branch:)

Izder456:
@SteveLauC This good?

Just found out `ctx.run_type()` does *not* export the `dry_run()` function. Its in `ctx.config()` actually.

SteveLauC (approved):
Looks good, thanks for the patch!

PR #953, opened 2024-10-15
ci: add OpenBSD vm-action for clippy

Izder456:
probably not an "ideal" way to do this, but as cross-rs doesn't support OpenBSD targets, this will have to do. These vm-action tools could likely be extended to produce release binaries and whatever else too.

## What does this PR do

Adds OpenBSD CI by way of OpenBSD VM-Actions https://github.com/vmactions/openbsd-vm

## Standards checklist

- [*] The PR title is descriptive.
- [*] I have read `CONTRIBUTING.md`
- [*] *Optional:* I have tested the code myself

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,3 +85,21 @@ jobs:
         # ONLY run test with cargo
         if: matrix.use_cross == false
         run: cargo test --locked --target ${{ matrix.target }} 
+
+  openbsd:
+    needs: fmt
+    name: OpenBSD (check, clippy)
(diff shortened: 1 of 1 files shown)


Izder456:
Closing as this is not something I want to continue supporting. It is a poor solution. We should do things via cross compilation instead when capable.

SteveLauC:
> Closing as this is not something I want to continue supporting. It is a poor solution. We should do things via cross compilation instead when capable.

Makes sense, and TIL, there are VM actions available for those *BSDs. 😄

PR #954, opened 2024-10-16
Improve OpenBSD -CURRENT detection and Dry-run feedback

Izder456:
## What does this PR do

This commit improves the -CURRENT detection by way of parsing `/etc/motd`. This change is more future-proof as when OpenBSD nears a stable release, `uname` will temporarily report like -STABLE.

This commit *also* adds feedback if -CURRENT is found to make debugging this feature easier with `--dry-run`, or, just a regular run as well.

## Standards checklist

### DIFF
--- a/src/steps/os/openbsd.rs
+++ b/src/steps/os/openbsd.rs
@@ -4,16 +4,16 @@ use crate::terminal::print_separator;
 use crate::utils::{get_require_sudo_string, require_option};
 use color_eyre::eyre::Result;
 use rust_i18n::t;
+use std::fs;
 
 fn is_openbsd_current(ctx: &ExecutionContext) -> Result<bool> {
+    let motd_content = fs::read_to_string("/etc/motd")?;
+    let is_current = motd_content.contains("-current");
     if ctx.config().dry_run() {
         println!("Would check if OpenBSD is -current");
-        Ok(false) // Default to false for dry-run
+        Ok(is_current)
     } else {
-        let output = ctx.run_type().execute("uname").arg("-r").output_checked()?;
-
-        let version = String::from_utf8_lossy(&output.stdout);
-        Ok(version.trim().ends_with("-current"))
+        Ok(is_current)
     }
 }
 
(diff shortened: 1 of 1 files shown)


Izder456:
@SteveLauC

Force pushed a fix.

This good?

Izder456:
@SteveLauC 

Force pushed a fix to my fork. Thanks for the review.

This good?

src/steps/os/openbsd.rs
```diff
@@ -4,30 +4,36 @@ use crate::terminal::print_separator;
...
+        if is_current {
+            println!("Would use -s flag when upgrading system");
+        }
         return Ok(());
     }
 
     let mut args = vec!["/usr/sbin/sysupgrade", "-n"];
-    if is_openbsd_current(ctx)? {
+    if is_current {
+        println!("OpenBSD is running -current, passing -s flag");
```
SteveLauC, line 36: Is this necessary, we are evening telling users that we are running `sysupgrade -n` (though they should know this)

src/steps/os/openbsd.rs
```diff
@@ -4,30 +4,36 @@ use crate::terminal::print_separator;
 use crate::utils::{get_require_sudo_string, require_option};
 use color_eyre::eyre::Result;
 use rust_i18n::t;
+use std::fs;
 
 fn is_openbsd_current(ctx: &ExecutionContext) -> Result<bool> {
+    let motd_content = fs::read_to_string("/etc/motd")?;
```
SteveLauC, line 10: Reading a file in a dry run harms nothing! Nice!

src/steps/os/openbsd.rs
```diff
@@ -51,7 +62,8 @@ pub fn upgrade_packages(ctx: &ExecutionContext) -> Result<()> {
     }
 
     let mut args = vec!["/usr/sbin/pkg_add", "-u"];
-    if is_openbsd_current(ctx)? {
+    if is_current {
+        println!("OpenBSD is running -current, passing -Dsnap flag");
```
SteveLauC, line 66: Same here

src/steps/os/openbsd.rs
Izder456: this is a valid point. I can improve the output to display the full command, but you're right, we can just not take that output to mirror other os steps.

src/steps/os/openbsd.rs
Izder456: I cant tell if this is genuine or sarcastic.

SteveLauC: Emm, I seriously mean this is good, I think we can do anything in a dry run as long as we don't do the update. 😁

src/steps/os/openbsd.rs
Izder456: alright, Ill fix up the commit to be less talky then.

src/steps/os/openbsd.rs
Izder456: gotcha, as you may have noticed, I ain't a great rust dev, so I couldn't tell if you were playfully mocking me.

thanks for clarifying.

Izder456:
@SteveLauC 

Pushed a commit to make it less talky. I also swapped "update" for "upgrade" to mirror other step's verbiage.

Ok?

SteveLauC (approved):
LGTM, thanks!

PR #970, opened 2024-10-27
feat: allow version specification for deno

sehnryr:
## What does this PR do
Add config field for specifying deno version.

## Standards checklist

## For new steps

  the underlying command

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -223,6 +223,11 @@
 # use_sudo = true
 
 
+[deno]
+# Upgrade deno executable to the given version.
+# version = stable
+
+
 [vim]
 # For `vim-plug`, execute `PlugUpdate!` instead of `PlugUpdate`
 # force_plug_update = true
--- a/config.example.toml
+++ b/config.example.toml
@@ -225,7 +225,7 @@
 
 [deno]
 # Upgrade deno executable to the given version.
-# version = stable
+# version = "stable"
 
 
 [vim]
--- a/src/steps/node.rs
+++ b/src/steps/node.rs
@@ -206,7 +206,10 @@ impl Deno {
                 match version {
                     "stable" => {}
                     "rc" => {
-                        return Err(SkipStep("Deno v1.x cannot be upgraded to a release candidate".to_string()).into());
+                        return Err(SkipStep(
+                            "Deno (1.6.0-2.0.0) cannot be upgraded to a release candidate".to_string(),
+                        )
+                        .into());
                     }
                     "canary" => args.push("--canary"),
                     _ => {
@@ -222,7 +225,9 @@ impl Deno {
                 match version {
                     "stable" | "rc" | "canary" => {
                         // Prior to v1.6.0, `deno upgrade` is not able fetch the latest tag version.
-                        return Err(SkipStep("Deno v1.x cannot be upgraded to a named channel".to_string()).into());
+                        return Err(
+                            SkipStep("Deno (1.0.0-1.6.0) cannot be upgraded to a named channel".to_string()).into(),
+                        );
                     }
                     _ => {
                         if Version::parse(version).is_err() {
--- a/src/steps/node.rs
+++ b/src/steps/node.rs
@@ -184,6 +184,32 @@ impl Yarn {
     }
 }
 
+struct Deno {
+    command: PathBuf,
+}
+
+impl Deno {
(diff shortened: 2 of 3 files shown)


SteveLauC:
> Add config field for specifying deno version.

What was the previous behavior?

sehnryr:
> > Add config field for specifying deno version.
> 
> What was the previous behavior?

`deno upgrade`s to "stable" channel by default and overwrites any other version.

SteveLauC:
> and overwrites any other version.

Say I have all the 3 channels installed, `deno upgrade` will upgrade the "stable" channel to remove all the other installations? Or can I have 3 channels installed? (Rust allows you to do so, I am not sure about the case of Deno)

Stupid question, if I already have "1.45.0" installed, what will `deno upgrade 1.45` do?

config.example.toml
```diff
@@ -223,6 +223,11 @@
 # use_sudo = true
 
 
+[deno]
+# Upgrade deno executable to the given version.
+# version = stable
```
SteveLauC, line 228:
```suggestion
# version = "stable"
```

If it should be a string

SteveLauC (changes requested):
Just one tiny issue

sehnryr:
> Say I have all the 3 channels installed, `deno upgrade` will upgrade the "stable" channel to remove all the other installations? Or can I have 3 channels installed? (Rust allows you to do so, I am not sure about the case of Deno)
> 
> Stupid question, if I already have "1.45.0" installed, what will `deno upgrade 1.45` do?

I don't think you can install multiple versions of Deno :thinking:.

`deno upgrade 1.45` won't work as it only recognize semver:
```
error: Invalid version passed (1.45)

Pass a semver, or a full 40 character git commit hash, or a release channel name.

Usage:
Latest
  deno upgrade

Specific version
  deno upgrade 1.45.0
  deno upgrade 1.46.0-rc.1
  deno upgrade 9bc2dd29ad6ba334fd57a20114e367d3c04763d4

Channel
  deno upgrade stable
  deno upgrade rc
  deno upgrade canary
```

And if we try upgrading to the same version, deno will recognize it and not replace it.

sehnryr:
I just found out there's an issue with deno's v1 `upgrade` arguments as they don't match with v2:
- v1: `deno upgrade [--version <version>]/[--canary]`
- v2: `deno upgrade [<version>]/[stable/rc/canary]`

src/steps/node.rs
```diff
@@ -184,6 +184,79 @@ impl Yarn {
...
+        if let Some(version) = version {
+            let bin_version = self.version()?;
+
+            if bin_version >= Version::new(2, 0, 0) {
+                args.push(version);
+            } else if bin_version >= Version::new(1, 6, 0) {
+                match version {
+                    "stable" => {}
+                    "rc" => {
+                        return Err(SkipStep("Deno v1.x cannot be upgraded to a release candidate".to_string()).into());
```
SteveLauC, line 209: Nit: this and line 225 are both called Deno v1.x, it might be better to make them different, e.g., `Deno (1.6.0-2.0.0)` and `Deno (1.0.0-1.6.0)`

src/steps/node.rs
```diff
@@ -184,6 +184,79 @@ impl Yarn {
...
+
+        let version = ctx.config().deno_version();
+        if let Some(version) = version {
+            let bin_version = self.version()?;
+
+            if bin_version >= Version::new(2, 0, 0) {
+                args.push(version);
+            } else if bin_version >= Version::new(1, 6, 0) {
+                match version {
+                    "stable" => {}
```
SteveLauC, line 207: Nit: To make it clear:

```suggestion
                    "stable" => { /* do nothing, as stable is the default channel to upgrade */ }
```

src/steps/node.rs
```diff
@@ -184,6 +184,79 @@ impl Yarn {
...
+        ctx.run_type()
+            .execute(&self.command)
+            .arg("upgrade")
+            .args(args)
+            .status_checked()?;
+        Ok(())
+    }
+
+    fn version(&self) -> Result<Version> {
+        let version_str = Command::new(&self.command)
```
SteveLauC, line 252: Could you please add one example output of this command here, which would make the code easier to understand

SteveLauC (commented):
Just some trivial comments, the impl looks generally good

PR #983, opened 2024-11-15
refactor: disable julia startup file for julia package update

Laura7089:
## What does this PR do

This PR adds the `--startup-file=no` argument to the `julia` invocation for updating Julia packages. See [the upstream doc](https://docs.julialang.org/en/v1/manual/command-line-interface/#command-line-interface).

This is useful because if Julia has been recently updated (eg. with the `juliaup` step), then old versions of packages configured with `using` directives in `~/.julia/config/startup.jl` may fail to precompile or load with the new version, in turn causing the package update step to fail unnecessarily. By skipping the startup file load, broken or outdated packages will not attempt to be loaded.

## Standards checklist

- ~~[ ] If this PR introduces new user-facing messages they are translated~~

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -453,6 +453,7 @@ pub struct Lensfun {
 }
 
 #[derive(Deserialize, Default, Debug, Merge)]
+#[serde(deny_unknown_fields)]
 pub struct JuliaConfig {
     startup_file: Option<bool>,
 }
--- a/src/config.rs
+++ b/src/config.rs
@@ -453,6 +453,11 @@ pub struct Lensfun {
 }
 
 #[derive(Deserialize, Default, Debug, Merge)]
+pub struct JuliaConfig {
+    startup_file: Option<bool>,
+}
+
+#[derive(Deserialize, Default, Debug, Merge)]
 #[serde(deny_unknown_fields)]
 /// Configuration file
 pub struct ConfigFile {
@@ -518,6 +523,9 @@ pub struct ConfigFile {
 
     #[merge(strategy = crate::utils::merge_strategies::inner_merge_opt)]
     lensfun: Option<Lensfun>,
+
+    #[merge(strategy = crate::utils::merge_strategies::inner_merge_opt)]
+    julia: Option<JuliaConfig>,
 }
 
 fn config_directory() -> PathBuf {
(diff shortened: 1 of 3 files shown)


SteveLauC:
Hi, thanks for the PR!

I would like to know if it would affect the update in other cases by not loading this `startup.jl` file?

SteveLauC:
Sorry for accidentally closing the PR. 😵‍💫

SteveLauC:
Thanks for the explanation!

>  If any breakage is a concern, perhaps it could be gated behind a configuration flag?

Yeah, adding a configuration entry for this makes sense:

```toml
[julia]
# If true, Topgrade will pass the `--startup-file=yes` option to Julia when updating it.
#
# <!!!Here, please add the use case where this option would be helpful!!!>
# startup_file = false
```

We need a new `[julia]` section, and a `startup_file` (you can use other names if you prefer) config entry in the example config file. This new section [will be added to this `struct ConfigFile` type](https://github.com/topgrade-rs/topgrade/blob/444689c899bc6a26c696fcfbccd0ef574a8f7870/src/config.rs#L520), you can add [such a helper function](https://github.com/topgrade-rs/topgrade/blob/444689c899bc6a26c696fcfbccd0ef574a8f7870/src/config.rs#L1166-L1172) to `struct Config` to make accessing this config easier.

Laura7089:
I've set the flag to default to `true` to mimic the (current at time of writing) default behaviour of julia. Is this appropriate or should we prefer the "fixed" case?

SteveLauC:
>  or should we prefer the "fixed" case?

Emm, what does the fixed case look like? You mean the change presented in [this commit](https://github.com/topgrade-rs/topgrade/pull/983/commits/08e9c002a6b80354025a80dca3bf42d49e01a82b)? I am not a user of Julia, but based on [our discussion](https://github.com/topgrade-rs/topgrade/pull/983#issuecomment-2480480513), making it configurable and don't change the current behavior is generally preferred, just in  case.

src/config.rs
```diff
@@ -452,6 +452,11 @@ pub struct Lensfun {
     use_sudo: Option<bool>,
 }
 
+#[derive(Deserialize, Default, Debug, Merge)]
+pub struct JuliaConfig {
```
SteveLauC, line 456:
```suggestion
#[derive(Deserialize, Default, Debug, Merge)]
#[serde(deny_unknown_fields)]
pub struct JuliaConfig {
```

SteveLauC (changes requested):
Just one nit:)

SteveLauC (approved):
Thanks!

Issue #997, opened 2024-12-08
bootc exists on system but disabled via config will ignore rpm-ostree


befanyt:
## Erroneous Behavior

Tried to update my system. With `rpm-ostree=true` and `bootc=false` (default). My system has bootc on it, so it found it but it is disabled via config and tries to update via `dnf` now.

## Expected Behavior

topgrade should have used `rpm-ostree` to update the system

## Steps to reproduce

Update the system with topgrade, where bootc is available but disabled on the system.

## Possible Cause (Optional)

https://github.com/topgrade-rs/topgrade/blob/4df30c25878ee3713152ea6cf3f4ce4225336fd6/src/steps/os/linux.rs#L233
I think that needs to be just an if statement, and not if else. Because bootc is found, then checks config which is disabled. Because bootc exists it will never go into the rpm-ostree flow when it finds out that it is disabled via config.

## Problem persists without calling from topgrade

- [ ] Yes
- [x] No

## Did you run topgrade through `Remote Execution`

- [ ] Yes
- [x] No

If yes, does the issue still occur when you run topgrade directlly in your
remote host

- [ ] Yes
- [ ] No

## Configuration file (Optional)

```toml
[misc]
no_self_update = true
only = ["system", "flatpak", "distrobox", "firmware"]
disable = ["self_update", "toolbx", "containers", "helm"]
ignore_failures = ["distrobox", "flatpak", "brew_cask", "brew_formula", "nix", "node", "pip3", "home_manager", "firmware"]
assume_yes = true
no_retry = false

[linux]
rpm_ostree = true
```

## Additional Details
- Operation System/Version
  
Universal Blue Bluefin 
- Installation
  
Via pip I think, its included in bluefin
- Topgrade version (`topgrade -V`)
`topgrade 16.0.2`
## Verbose Output (`topgrade -v`)

<details>
<pre>
── 14:44:42 - System update ────────────────────────────────────────────────────
DEBUG Detected "/usr/bin/bootc" as "bootc"
DEBUG Detected "/usr/bin/dnf" as "dnf"
DEBUG Executing command `/usr/bin/sudo /usr/bin/dnf upgrade`
</pre>
</details>

PR #986, opened 2024-11-17
feat: add bootc support

tulilirockz:
## What does this PR do

This PR adds [Bootc](https://github.com/containers/bootc) support for redhat_upgrade as they are going to ship this as the default starting with F42 Atomic AFAIK. It takes precedence over `rpm-ostree` if it is enabled.

## Standards checklist

- [O] The PR title is descriptive
- [O] I have read `CONTRIBUTING.md`
- [O] *Optional:* I have tested the code myself
- [O] If this PR introduces new user-facing messages they are translated

If you developed a feature or a bug fix for someone else and you do not have the
means to test it, please tag this person here.

PR #999, opened 2024-12-09
fix: dont ignore rpm-ostree when bootc is found

befanyt:
## What does this PR do
This PR should fix https://github.com/topgrade-rs/topgrade/issues/997 where it wouldn't take the rpm-ostree flow when bootc was found on the system, even if it was disabled via the config.

Tested with `rpm-ostree=false` & `bootc=false` it used `dnf`
Tested with `rpm-ostree=true` & `bootc=false` it used `rpm-ostree`
Tested with `rpm-ostree=false` & `bootc=true` it used `bootc`

## Standards checklist

## For new steps

  the underlying command

This was introducued in PR https://github.com/topgrade-rs/topgrade/pull/986
@tulilirockz maybe you can verify this also works for you.


tulilirockz (approved):

SteveLauC (approved):
Thanks for catching and fixing it! Also for the thorough tests!

PR #1000, opened 2024-12-10
refactor(uv step): check self update result if self-update feat is available

SteveLauC:
## What does this PR do

For the `uv` step, if the self-update feature is available, we should do the update and **check** its result.

cc @lucaspar 

## Standards checklist

## For new steps

  the underlying command

### DIFF
--- a/src/steps/generic.rs
+++ b/src/steps/generic.rs
@@ -1129,8 +1129,7 @@ pub fn run_uv(ctx: &ExecutionContext) -> Result<()> {
         ctx.run_type()
             .execute(&uv_exec)
             .args(["self", "update"])
-            .status_checked()
-            .ok();
+            .status_checked()?;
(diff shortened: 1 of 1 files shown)


lucaspar (approved):

SteveLauC:
Thanks for reviewing!

Jisu-Woniu:
Some uv distributions like Arch Linux allow `uv self --help`, but `uv self update` will produce an error:

screenshot

Maybe we should revert this.

GideonBear:
Hi @Jisu-Woniu, could you open a separate issue for this? This seems to imply that `uv` can be in three possible states, of which two are handled correctly. Can you share (in the issue) a way to reproduce the state it's in (`uv self --help` succeeeds, but `uv self update` fails)?

PR #1002, opened 2024-12-10
fix: Executor::spawn()/output() should not use their _checked() variants

SteveLauC:
## What does this PR do

`CommandExt::output_checked()` returns `Err()` if the command fails, using it in `Executor::output()` will make its behavior different from `std::process::Command::output()`.

The same applies to `spawn()` as well. `Executor` does not have a `status()` command.

These changes were made 2 years ago by @9999years, it would be great if we can have you here and take a look at this PR:)

## Standards checklist

## For new steps

  the underlying command

### DIFF
--- a/src/executor.rs
+++ b/src/executor.rs
@@ -152,7 +152,10 @@ impl Executor {
         let result = match self {
             Executor::Wet(c) => {
                 debug!("Running {:?}", c);
-                c.spawn_checked().map(ExecutorChild::Wet)?
+                // We should use `spawn()` here rather than `spawn_checked()` since
+                // their semantics and behaviors are different.
+                #[allow(clippy::disallowed_methods)]
+                c.spawn().map(ExecutorChild::Wet)?
             }
             Executor::Dry(c) => {
                 c.dry_run();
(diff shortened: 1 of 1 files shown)


clippy.toml
```diff
@@ -1,5 +0,0 @@
-disallowed-methods = [
```
9999years, line 1: We _might_ want to keep these for the rest of the codebase and do `#[allow(clippy::disallowed_methods)]` on the calls in `src/executor.rs` (where we have a reason to use the raw `spawn` method).

9999years (approved):
This looks good! Thanks for the ping —I really love `topgrade` so I'm happy to review PRs like this. Thanks for all you do to maintain it!

9999years:
BTW, since I wrote the `CommandExt` trait used in `topgrade`, I've expanded and polished it into [the `command_error` crate](https://docs.rs/command-error/latest/command_error/) (which builds on [the `utf_command`](https://docs.rs/utf8-command/latest/utf8_command/) crate); we might consider switching to that at some point.

clippy.toml
SteveLauC: Makes sense:)

SteveLauC:
> BTW, since I wrote the `CommandExt` trait used in `topgrade`, I've expanded and polished it into [the `command_error` crate](https://docs.rs/command-error/latest/command_error/) (which builds on [the `utf_command`](https://docs.rs/utf8-command/latest/utf8_command/) crate); we might consider switching to that at some point.

Neat! Topgrade should switch to it!

PR #1030, opened 2025-02-07
Add zigup step

Laura7089:
## What does this PR do

This PR adds a step to run [`zigup`](https://github.com/marler8997/zigup), a zig version manager comparable to the already-supported ZVM. The step can be configured to update multiple version tracking strings, mark them for keeping/clean up old versions, and change the directory they're installed into as well as the symlink path that they're "installed with".

By default, the step updates the `master` branch, which is currently considered the recommended version of zig (see "Tagged release or nightly build?" [here](https://ziglang.org/learn/getting-started/)), falling back to the default behaviour of `zigup` with paths, and not deleting any files.

## Standards checklist

## For new steps

- [x] *Optional:* The `--dry-run` option works with this step - **I believe this should work as the step doesn't perform anything outside of the `command` API**
- [ ] ~~*Optional:* The `--yes` option works with this step if it is supported by 
  the underlying command~~ **N/A**

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -297,10 +297,11 @@
 # (default: false)
 # set_default = false
 
-# Specifies the version string passed to zigup.
-# This may be a pinned version such as "0.13.0" or a branch such as "master".
-# (default: "master")
-# target_version = "master"
+# Version strings passed to zigup.
+# These may be pinned versions such as "0.13.0" or branches such as "master".
+# Each one will be updated in its own zigup invocation.
+# (default: ["master"])
+# target_versions = ["master", "0.13.0"]
 
 # Specifies the directory that the zig files will be installed to.
 # If defined, passed with the --install-dir command line flag.
@@ -315,6 +316,7 @@
 # (default: not defined)
 # path_link = "/home/user/.bin/zig"
 
-# If enabled, run `zigup clean` after updating the version.
+# If enabled, run `zigup clean` after updating all versions.
+# If enabled, each updated version above will be marked with `zigup keep`.
 # (default: false)
 # cleanup = false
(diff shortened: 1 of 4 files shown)


Laura7089:
Apologies for that force-push, wanted to reword some commits.

Laura7089:
Frustratingly, I couldn't get `~` to expand correctly in the path options.

SteveLauC:
> Frustratingly, I couldn't get ~ to expand correctly in the path options.

You can use `shellexpand:: tilde()`, see an example here:

https://github.com/topgrade-rs/topgrade/blob/224bb96a98b06f1000106f511012c12963f2e115/src/config.rs#L701

config.example.toml
```diff
@@ -290,3 +290,28 @@
...
+# If defined, passed with the --path-link command line flag.
+# If not defined, zigup will use its default behaviour.
+# This is not meaningful if set_default is not enabled.
+# (default: not defined)
+# path_link = "/home/user/.bin/zig"
+
+# If enabled, run `zigup clean` after updating all versions.
+# If enabled, each updated version above will be marked with `zigup keep`.
+# (default: false)
+# cleanup = false
```
SteveLauC, line 317: Topgrade already has a `clean` option, can we use it?

https://github.com/topgrade-rs/topgrade/blob/224bb96a98b06f1000106f511012c12963f2e115/config.example.toml#L56-L57

Laura7089: I think that this is closely related, but unfortunately the `zigup clean` command will clean up any toolchains which the user might have manually requested unless they've marked them with `zigup keep` (as the step does). Thus, it could result in unexpected behaviour as the default.

SteveLauC: Get it

SteveLauC (approved):
Thanks!

PR #1042, opened 2025-02-16
Fix winget release handling

niStee:
… package to Windows Package Manager

## Standards checklist

## For new steps

  the underlying command

@SteveLauC

### DIFF
--- a/.github/workflows/release_to_winget.yml
+++ b/.github/workflows/release_to_winget.yml
@@ -1,13 +1,61 @@
-name: Publish to WinGet
+name: Submit Topgrade-rs.Topgrade package to Windows Package Manager Community Repository
+
 on:
+  workflow_dispatch:
   release:
-    types: [released]
+    types: [published]
+
 jobs:
-  publish:
+  sync-repo:
+    name: Sync winget-pkgs repo
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout winget-pkgs repository
+        uses: actions/checkout@v2
+        with:
+          repository: cinnamon-msft/winget-pkgs
+          token: ${{ secrets.WINGET_PAT }}
+
+  winget:
+    name: Publish winget package
+    needs: sync-repo
     runs-on: windows-latest
     steps:
-      - uses: vedantmgoyal2009/winget-releaser@main
-        with:
-          identifier: topgrade-rs.topgrade
-          max-versions-to-keep: 5 # keep only latest 5 versions
-          token: ${{ secrets.WINGET_TOKEN }}
+      - name: Checkout repository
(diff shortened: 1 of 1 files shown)


niStee:
@SteveLauC This PR seems not to be necessary anymore since there was a successful release of Topgrade for Winget, am I right?

niStee:
> @SteveLauC This PR seems not to be necessary anymore since there was a successful release of Topgrade for Winget, am I right?

Our _Publish to WinGet_ action still seems to have a problem. Looks like someone else published it there.

.github/workflows/release_to_winget.yml
```diff
@@ -1,13 +1,61 @@
...
-      - uses: vedantmgoyal2009/winget-releaser@main
+      - name: Checkout winget-pkgs repository
+        uses: actions/checkout@v2
         with:
-          identifier: topgrade-rs.topgrade
-          max-versions-to-keep: 5 # keep only latest 5 versions
-          token: ${{ secrets.WINGET_TOKEN }}
\ No newline at end of file
+          repository: cinnamon-msft/winget-pkgs
+          token: ${{ secrets.WINGET_PAT }}
```
SteveLauC, line 17: What is this token? Where can we generate one?

SteveLauC: We indeed have a `WINGET_TOKEN` set in the CI

SteveLauC (approved):
Thanks!

.github/workflows/release_to_winget.yml
```diff
@@ -1,13 +1,58 @@
...
-      - uses: vedantmgoyal2009/winget-releaser@main
+      - name: Checkout winget-pkgs repository
+        uses: actions/checkout@v4
         with:
-          identifier: topgrade-rs.topgrade
-          max-versions-to-keep: 5 # keep only latest 5 versions
-          token: ${{ secrets.WINGET_TOKEN }}
\ No newline at end of file
+          repository: microsoft/winget-pkgs
+          token: ${{ secrets.WINGET_TOKEN }}
```
SteveLauC, line 17: This job is needed by the next job,  but they will be run in 2 separate runners, i.e., the result of this job won't be available to the ohter job, is this something correct?

niStee: Yes, your observation is correct. These jobs run on separate runners, so the output from the sync-repo job is not automatically available to the winget job. After considering this, and noticing that the template I was using relied on a fork of the winget community repository, I realized that they were synchronizing their fork first to ensure up-to-date data and avoid conflicts. However, I want to try pushing directly to the community repository instead, so I removed the dependency. I think we need to test if this works now.

PR #1045, opened 2025-02-17
feat(step): nix-helper

### LINKED ISSUE #1044: [feature request] Add support for viperML/nh
`nh`, the nix helper, used for updating nixos and home-manager through flakes
GitHub: https://github.com/viperML/nh
Found in nixpkgs at: https://github.com/NixOS/nixpkgs/blob/nixos-24.11/pkgs/by-name/nh/nh/package.nix

Related to #719 because `nh` relies on `nvd` and `nom` to improve UX. Read `nh`'s README for a better elevator pitch...

A `FLAKE=/home/user/nix-configuration` is set to specify where the NixOS and home-manager configuration lives. This is useful for people who keep their nix dotfiles version-controlled in a flake.

They only target NixOS-supported systems.

In their `flake.nix` they specify:
```nix
[
    "x86_64-linux"
    "aarch64-linux"
    # experimental
    "x86_64-darwin"
    "aarch64-darwin"
]
```

Check wether `$FLAKE` exists, is a directory and contains a `flake.nix`. Then we know that a NixOS and/or home-manager configuration should live in that directory.

To check wether a NixOS configuration is present. You could parse the output of `nix flake show $FLAKE --json` and look for `nixosConfigurations.$(hostname)` and check wether its type is `nixos-configuration`...

Example output of `nix flake show $FLAKE --json`:
```json
{
  "homeConfigurations": {
    "type": "unknown"
  },
  "nixosConfigurations": {
    "iso": {
      "type": "nixos-configuration"
    },
    "tom-pc": {
      "type": "nixos-configuration"
    },
    "tom-pc-vm": {
      "type": "nixos-configuration"
    }
  }
}
```

To check wether a home-manager configuration is present, you could parse the output of `nix eval --json $FLAKE#homeConfigurations --apply builtins.attrNames` and look for an entry in the form of `$(whoami)@$(hostname)` or just `$(whoami)`.

Example output of `nix eval --json $FLAKE#homeConfigurations --apply builtins.attrNames`:
```json
["tomvd@aurora", "tomvd@tom-pc", "alice"]
```

- When "$FLAKE" is a valid flake: `nix --experimental-features "nix-command flakes" flake update $FLAKE`;
- When we've found a NixOS configuration: `nh os switch`;
- When we've found a home-manager configuration: `nh home switch`.

`nix flake update` does NOT have `--dry-run`, but `nh` has `--dry`.

- `nix flake update` never asks for user input;
- `nh os switch` never asks by itself unless you specify `--ask`, but requires a sudo password to actually finish the update;
- `nh home switch` never asks for confirmation unless `--ask` is specified.

## More information

I know how to test this feature if someone were to implement it. I am willing to do it myself though.

dtomvan:
## What does this PR do
Add support for viperML/nh. Closes #1044.

## Standards checklist

## For new steps

  the underlying command

### DIFF
--- a/src/steps/os/unix.rs
+++ b/src/steps/os/unix.rs
@@ -679,7 +679,14 @@ pub fn run_nix_helper(ctx: &ExecutionContext) -> Result<()> {
     if nixos_cfg.is_ok_and(|x| x.contains(&format!(r#""{hostname}""#))) {
         print_separator("nh os");
 
-        run_type.execute(&nix_helper).arg("os").arg("switch").status_checked()?;
+        let mut cmd = run_type.execute(&nix_helper);
+        cmd.arg("os");
+        cmd.arg("switch");
+
+        if !ctx.config().yes(Step::NixHelper) {
+            cmd.arg("--ask");
+        }
+        cmd.status_checked()?;
     }
 
     if let Ok(home_cfg) = home_cfg {
@@ -687,11 +694,14 @@ pub fn run_nix_helper(ctx: &ExecutionContext) -> Result<()> {
         if home_cfg.contains(&format!(r#""{username}""#)) || home_cfg.contains(&username_at_hostname) {
             print_separator("nh home");
 
-            run_type
-                .execute(&nix_helper)
-                .arg("home")
-                .arg("switch")
-                .status_checked()?;
+            let mut cmd = run_type.execute(&nix_helper);
+            cmd.arg("home");
+            cmd.arg("switch");
+
+            if !ctx.config().yes(Step::NixHelper) {
+                cmd.arg("--ask");
+            }
+            cmd.status_checked()?;
         }
     }
 
--- a/src/steps/os/unix.rs
+++ b/src/steps/os/unix.rs
@@ -18,9 +18,7 @@ use std::path::PathBuf;
 use std::process::Command;
 use std::sync::LazyLock;
 use std::{env::var, path::Path};
-use tracing::debug;
-use whoami::fallible::hostname;
-use whoami::fallible::username;
+use tracing::{debug, warn};
 
 #[cfg(target_os = "linux")]
 use super::linux::Distribution;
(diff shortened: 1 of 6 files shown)


src/steps/os/unix.rs
```diff
@@ -622,8 +624,92 @@ fn nix_profile_dir(nix: &Path) -> Result<Option<PathBuf>> {
...
+            let mut cmd = run_type.execute(&nix_helper);
+            cmd.arg("home");
+            cmd.arg("switch");
+
+            if !ctx.config().yes(Step::NixHelper) {
+                cmd.arg("--ask");
+            }
+            cmd.status_checked()?;
+        }
+    }
```
GideonBear: This is duplicated, and in slightly different format; I would prefer deduplicating this and merging the `if let Ok(<...>) = <...>` and `if <...>.is_ok_and(|x| <...>)`

src/steps/os/unix.rs
```diff
@@ -622,8 +624,92 @@ fn nix_profile_dir(nix: &Path) -> Result<Option<PathBuf>> {
...
+
+    Ok(())
+}
+
 fn nix_args() -> [&'static str; 2] {
-    ["--extra-experimental-features", "nix-command"]
+    ["--extra-experimental-features", "nix-command flakes"]
```
GideonBear: What is this for?
Is `nix_args` used anywhere else, and is this change desired there?

src/steps/os/unix.rs
```diff
@@ -622,8 +624,92 @@ fn nix_profile_dir(nix: &Path) -> Result<Option<PathBuf>> {
...
+    if nixos_cfg.is_ok_and(|x| x.contains(&format!(r#""{hostname}""#))) {
+        print_separator("nh os");
+
+        let mut cmd = run_type.execute(&nix_helper);
+        cmd.arg("os");
+        cmd.arg("switch");
+
+        if !ctx.config().yes(Step::NixHelper) {
+            cmd.arg("--ask");
+        }
```
GideonBear, line 678: If `--ask` is not the default, should it be the default in Topgrade?
This is more about the semantics of `assume_yes`, @SteveLauC what are your thoughts?

GideonBear (commented):
I don't know nix, so I can't say much about the functionality itself. Looks fine though!

src/steps/os/unix.rs
dtomvan: `nix` will refuse to work with flakes at all if that feature flag isn't enabled, hence updating the system with `nh` will most likely not work. For example, without this, the `nix env` call at line 645 won't work at all. https://nix.dev/manual/nix/2.29/development/experimental-features.html?highlight=features#xp-feature-flakes

src/steps/os/unix.rs
dtomvan: > deduplicating this

I agree now that I look at it 4 months later that the format discrepancy isn't preferable, but how would this be deduplicated? 

1. They are looking for slightly different strings inside of the `nix eval` output
2. They are looking inside two distinct `nix eval` invocations (one in `nixosConfigurations` and one in `homeConfigurations`)
3. They run different commands
4. They have a different separator

I think they look different enough. I could factor it out into some `run_nh_update(attr: &'static str, formats: &[&'static str], nh_subcommand: &'static str)` function which does all of that in a more generic way, but I don't think that makes it any more readable/usable.

dtomvan:
Also, due to [a breaking change in nh](https://github.com/nix-community/nh/pull/151) ever since, the `FLAKE` variable got renamed to `NH_FLAKE`. Not only that, but it also got split into `NH_OS_FLAKE`, `NH_DARWIN_FLAKE` an `NH_HOME_FLAKE`. Allow me to refactor a bit.

Edit: I'll just remove the `nix eval` command (which previously would check whether or not the quite generic `FLAKE` variable would contain any desired configuration), since after the breaking change the environment variables are a lot more specific and `nh` will do an eval check beforehand anyways.

src/steps/os/unix.rs
dtomvan: I assumed that if `--yes` _isn't_ passed, you'd expect `--ask` to be passed to `nh` here, since they are functionally inverse of each other

src/steps/os/unix.rs
GideonBear: Oh I see, I did not realize it was a list of arguments for `--extra-experimental-features`. Thanks for clearing that up!

dtomvan: So I kinda changed my mind with the new way `nh` works (with 4 different environment variables, with much more explicit names). I decided for simplicity that it would be less spaghetti if I just let topgrade throw an error if any of the environment variables set are wrong or don't contain a valid configuration. I removed the evaluation check and passed `-u` to `nh` (updates the flake inputs), which replaces both reasons to set the `flakes` feature flag at all....

src/steps/os/unix.rs
```diff
@@ -622,6 +622,83 @@ fn nix_profile_dir(nix: &Path) -> Result<Option<PathBuf>> {
...
+    if home_flake_path.is_some() || fallback_flake_path.is_some() {
+        nh_switch("home")?;
+    }
+
+    #[cfg(target_os = "linux")]
+    if matches!(Distribution::detect(), Ok(Distribution::NixOS))
+        && (nixos_flake_path.is_some() || fallback_flake_path.is_some())
+    {
+        nh_switch("os")?;
+    }
```
GideonBear, line 700: Is there a reason fallback is missing here?

src/steps/os/unix.rs
GideonBear: Looks like this is resolved after your refactor

src/steps/os/unix.rs
dtomvan: What do you mean by that? `nh fallback switch` isn't a command?

dtomvan:
Disclaimer: I've just added the `darwin` functionality, which I haven't tested as I'm only on Linux.

src/steps/os/unix.rs
GideonBear: Sorry I didn't pay attention to the code

GideonBear (approved):

src/steps/os/unix.rs
GideonBear: `assume_yes` is explicitly stated to have "no default", seeming to indicate that `Some(false)` and `None` can have a different meaning, although from what I can find, none of the steps currently use that.
Anyway since `Config::yes` doesn't really support that idea right now, and there's no prior art, I'm fine with this.

GideonBear (approved):

GideonBear:
I've edited your original message to contain `Closes #1044` instead of `See #1044` so that GitHub automatically closes the issue.

GideonBear:
Thanks! (ik moest lachen om "puntbestanden" :joy: Doet me denken aan [opnieuw](https://pypi.org/project/opnieuw/))

PR #1055, opened 2025-03-03
Add "Cinnamon spices" step

### LINKED ISSUE #1054: New step: `cinnamon-spice-updater`
`cinnamon-spice-updater`, a CLI tool built into [Linux Mint 20.2 and later](https://www.linuxmint.com/rel_uma_cinnamon_whatsnew.php)
Linux Mint 20.2 and later
Running `cinnamon-spice-updater --update-all` will output either `There are no Spice updates`, or `Updating all spices.`, followed by a list of spices that it then updates. You can also run `cinnamon-spice-updater --list-simple <spice-type>` or `cinnamon-spice-updater --list-json <spice-type>` to list updates for each of the spice types, in human/machine readable format respectively.
`cinnamon-spice-updater --update-all`
Yes, by running `cinnamon-spice-updater --list-simple <spice-type>` for each of the spice types ('applet', 'desklet', 'extension', 'theme')
No

GideonBear:
Closes #1054

## What does this PR do
Adds a "Cinnamon spices" step that uses "cinnamon-spice-updater"

## Standards checklist

## For new steps

  the underlying command

The step is currently guarded only for Linux and not for Linux Mint specifically. But I don't see why it would matter, since the command is only available on Linux Mint anyway. What do you think?

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -69,6 +69,7 @@ pub enum Step {
     Chezmoi,
     Chocolatey,
     Choosenim,
+    CinnamonSpices,
     ClamAvDb,
     Composer,
     Conda,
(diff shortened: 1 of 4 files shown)


SteveLauC (approved):
Thanks!

Issue #1057, opened 2025-03-04
pre_sudo doesn't cache credentials with GSudo


Andreas02-dev:
## Erroneous Behavior

When GSudo is used with [PathPrecedence](https://github.com/gerardog/gsudo/blob/master/docs/docs/gsudo-vs-sudo.md#user-experience) set to true, `pre-sudo` doesn't cache the credentials.

## Expected Behavior

`pre-sudo` caches credentials correctly with GSudo when PathPrecedence is enabled.

## Steps to reproduce

1. On a recent Windows 11 version, install GSudo and set PathPrecedence to true as mentioned in [PathPrecedence](https://github.com/gerardog/gsudo/blob/master/docs/docs/gsudo-vs-sudo.md#user-experience).
2. Enable `pre-sudo` in `topgrade.toml`
3. Run topgrade without admin, and verify that `sudo -v` is executed instead of `sudo echo`.

topgrade.toml
```toml
[misc]
# Run `sudo -v` to cache credentials at the start of the run
# This avoids a blocking password prompt in the middle of an unattended run
# (default: false)
pre_sudo = true
```

## Possible Cause (Optional)

Since the `sudo` binary exists by default in Windows 11 now and MS `sudo` is evaluated before `GSudo` in [Sudo::detect](https://github.com/topgrade-rs/topgrade/blob/b1ffe7d5537dad5250181868f22edce0d1b78d6f/src/sudo.rs#L26), even with GSudo PathPrecedence set to true it will issue `SudoKind::Sudo` instead of `SudoKind::GSudo`.

Additionally, `gsudo status` only shows the current credential cache status and doesn't elevate & cache credentials. This should probably be changed to something that *does* require elevation, like a dummy `echo` command that other elevation tools use in `sudo.rs`.

## Problem persists without calling from topgrade

- [ ] Yes
- [x] No

## Did you run topgrade through `Remote Execution`

- [ ] Yes
- [x] No

## Configuration file (Optional)

Relevant part:
```toml
[misc]
# Run `sudo -v` to cache credentials at the start of the run
# This avoids a blocking password prompt in the middle of an unattended run
# (default: false)
pre_sudo = true
```

## Additional Details
- Operation System/Version
  

Windows 11, version 24H2 build 26100.3194

- Installation
  

Winget

- Topgrade version (`topgrade -V`)

topgrade 16.0.2

## Verbose Output (`topgrade -v`)

<details>
<pre>
[...]

PR #1058, opened 2025-03-04
fix(sudo): fix sudo detection & pre-sudo for GSudo

Andreas02-dev:
## What does this PR do
When the `sudo` binary is detected, an additional check is added to see whether it's actually `GSudo` masquerading as the `sudo` binary and the right `SudoKind` will be applied.
GSudo does this by making a `sudo.exe` that's symlinked to the `gsudo.exe` binary. We check if the canonical path of `sudo.exe` is equal to the canonical path of `gsudo.exe`, and apply `SudoKind::Gsudo` when true.

Additionally, during testing I've found that `gsudo status` doesn't cache credentials (it only shows whether credentials are currently cached and doesn't elevate), so `pre_sudo` likely never worked with `gsudo` in the first place. I've changed this to run `gsudo echo` instead, which does elevate and caches the credentials (if the credential cache has been enabled beforehand, of course).

Fixes issue #1057.

## Standards checklist

### DIFF
--- a/src/sudo.rs
+++ b/src/sudo.rs
@@ -22,12 +22,26 @@ pub struct Sudo {
 }
 
 impl Sudo {
+    /// Get the `sudo` binary or the `gsudo` binary in the case of `gsudo`
+    /// masquerading as the `sudo` binary.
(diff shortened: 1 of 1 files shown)


SteveLauC (approved):
Thanks!

PR #1061, opened 2025-03-06
feat(winget): winget uses sudo when `[windows] winget_use_sudo = true`

### LINKED ISSUE #1060: `Feature` | Use `sudo` for `winget`
Winget now supports elevation using sudo. This can get rid of a lot of pesky UAC-prompts during the `winget` upgrade step.

I'll see if I can work this out tonight after I get off work & submit a PR to improve the `winget` step to use `sudo`.

Andreas02-dev:
## What does this PR do

Allow winget to elevate using sudo when `winget_use_sudo` is set to true in the `windows` section of the configuration file. When utilizing sudo credential-caching, this allows for just 1 elevation prompt instead of multiple elevation prompts (for every update that needs elevation).

Since some winget installers don't support upgrades/installs in root-mode and will cause the winget step to fail if run in root-mode, this option will be turned off by default and can optionally be toggled on by advanced users. By utilizing sudo for winget, we don't need to run every topgrade step in root-mode which could otherwise cause similar issues in other steps.

Closes #1060

## Standards checklist


config.example.toml
```diff
@@ -228,6 +228,15 @@
...
 # self_rename = true
 
+# Use sudo to elevate privileges for the Windows Package Manager (winget)
+# Only use this option if you want to run the Winget step in sudo-mode.
+# Running winget in sudo-mode is generally not recommended, as not every
+# package supports installing / upgrading in sudo-mode and it may cause issues
+# with some packages or may even cause the Winget-step to fail.
+# If any problems occur, please try running Topgrade without this option first
+# before reporting an issue.
+# winget_use_sudo = true
```
GideonBear:
```suggestion
# before reporting an issue.
# (default: false)
# winget_use_sudo = true
```

src/config.rs
```diff
@@ -1236,6 +1237,15 @@ impl Config {
             .unwrap_or(false)
     }
 
+    // Should use sudo for Winget
```
```suggestion
    /// Should use sudo for Winget
```

GideonBear:
Nits, looks good!

Andreas02-dev:
> Nits, looks good!

Thank you for the review and the suggestions!

GideonBear (approved):

PR #1063, opened 2025-03-08
Enhancement/refactor windows.rs

niStee:
## Standards checklist

- [ ] The PR title is descriptive
 
## For new steps

  the underlying command


locales/app.yml
```diff
@@ -1,5 +1,10 @@
 _version: 2
 
+"Could not find Topgrade in any WSL distribution":
```
SteveLauC, line 3: Thanks for fixing the typo, you should fix the existing localization entries rather than adding new ones:

https://github.com/topgrade-rs/topgrade/blob/ed37763d301f005ae8a108e31402d855386b0db4/locales/app.yml#L751-L755

src/steps/os/windows.rs
```diff
@@ -77,103 +131,19 @@ pub fn update_wsl(ctx: &ExecutionContext) -> Result<()> {
...
 
-/// Detect if WSL is installed or not.
-///
-/// For WSL, we cannot simply check if command `wsl` is installed as on newer
-/// versions of Windows (since windows 10 version 2004), this commmand is
-/// installed by default.
-///
-/// If the command is installed and the user hasn't installed any Linux distros
-/// on it, command `wsl -l` would print a help message and exit with failure, we
-/// use this to check whether WSL is install or not.
```
SteveLauC, line 101: As a non-Windows user, reviewing Windows change is pretty hard for me, please don't remove these explanation comments, without them, maintenance would be seriously harder. And, if possible, please add more such comments to make Topgrade maintainable:)

src/steps/os/windows.rs
```diff
@@ -230,8 +198,6 @@ pub fn microsoft_store(ctx: &ExecutionContext) -> Result<()> {
 }
 
 pub fn reboot() -> Result<()> {
-    // If this works, it won't return, but if it doesn't work, it may return a useful error
-    // message.
```
SteveLauC, line 234: Same here, please don't remove them. If they are wrong, please give them an update:)

src/steps/os/windows.rs
```diff
@@ -37,36 +44,83 @@ pub fn run_chocolatey(ctx: &ExecutionContext) -> Result<()> {
...
-        .args(["upgrade", "--all"])
-        .status_checked()
+    command.status_checked()
+}
+
+fn find_topgrade_in_wsl(wsl: &Path, dist: &str) -> Result<String> {
+    Ok(Command::new(wsl)
+        .args(["-d", dist, "bash", "-lc", "which topgrade"])
+        .output_checked_utf8()
+        .map_err(|_| SkipStep(t!("Could not find Topgrade installed in WSL").to_string()))?
```
niStee, line 100: > Thanks for fixing the typo, you should fix the existing localization entries rather than adding new ones:
> 
> https://github.com/topgrade-rs/topgrade/blob/ed37763d301f005ae8a108e31402d855386b0db4/locales/app.yml#L751-L755

@SteveLauC

src/steps/os/windows.rs
```diff
@@ -199,7 +169,7 @@ pub fn run_wsl_topgrade(ctx: &ExecutionContext) -> Result<()> {
     if ran {
         Ok(())
     } else {
-        Err(SkipStep(t!("Could not find Topgrade in any WSL disribution").to_string()).into())
+        Err(SkipStep(t!("Could not find Topgrade in any WSL distribution").to_string()).into())
```
niStee, line 200: > Thanks for fixing the typo, you should fix the existing localization entries rather than adding new ones:
> 
> https://github.com/topgrade-rs/topgrade/blob/ed37763d301f005ae8a108e31402d855386b0db4/locales/app.yml#L751-L755

@SteveLauC

locales/app.yml
SteveLauC: Emm, this is still a new entry, you should update the existing entry:

https://github.com/topgrade-rs/topgrade/blob/a601d8429dd351f47f128685b6a5eb8472459fa7/locales/app.yml#L1187-L1194

src/steps/os/windows.rs
niStee: I think I finally fixed it.

src/steps/os/windows.rs
niStee: Now it will return "ok" or a useful error. Do you want a explanation comment for this?

locales/app.yml
niStee: See https://github.com/topgrade-rs/topgrade/pull/1063#discussion_r2036036087

src/steps/os/windows.rs
SteveLauC: Yeah, it is fixed now

src/steps/os/windows.rs
```diff
@@ -13,62 +13,180 @@ use crate::{error::SkipStep, steps::git::RepoStep};
...
-        ctx.run_type().execute(&scoop).args(["cleanup", "*"]).status_checked()?;
-        ctx.run_type()
-            .execute(&scoop)
-            .args(["cache", "rm", "-a"])
-            .status_checked()?
+        cleanup_scoop(ctx, &scoop)?;
     }
 
     Ok(())
 }
```
SteveLauC, line 154: This function is less than 20 lines, quite clean and readable. Do we really need to split into 3 functions?

src/steps/os/windows.rs
```diff
@@ -13,62 +13,180 @@ use crate::{error::SkipStep, steps::git::RepoStep};
...
-    let yes = ctx.config().yes(Step::Chocolatey);
-
-    print_separator("Chocolatey");
+// Common command arguments for package managers
+const UPGRADE_ALL: &[&str] = &["upgrade", "all"];
+const UPDATE: &[&str] = &["update"];
+const CLEANUP: &[&str] = &["cleanup", "*"];
+const CACHE_RM: &[&str] = &["cache", "rm", "-a"];
+const WSL_UPDATE: &[&str] = &["--update"];
+const WSL_LIST: &[&str] = &["--list", "-q"];
```
SteveLauC, line 22: I kinda think we are over-engineering, and it can bite us in the future.

src/steps/os/windows.rs
niStee: I argue that this separation follows the pattern of breaking operations into small, focused helpers. While combining them would be concise, the separation provides better documentation and maintainability, and follows the principle of separation of concerns established in the rest of the code. It makes future changes to the update or cleanup logic more isolated.

src/steps/os/windows.rs
niStee: From my perspective, the constants seem to provide clear documentation of command structures and centralize command definitions, which has the potential to streamline future changes as package manager syntax evolves. However, you make a valid point that if some constants are used only once, they add an unnecessary layer of abstraction that complicates rather than simplifies the code. A balanced approach might be to keep constants for truly reused command patterns, while inlining single-use commands, or possibly grouping them by package manager into dedicated modules as the application continues to grow. I would very much appreciate your thoughts on the best way to proceed.

PR #1064, opened 2025-03-08
Add JetBrains Toolbox via jetbrains-toolbox-updater

### LINKED ISSUE #1056: Add support for JetBrains Toolbox
Hello! I figured out how to update JetBrains IDE's via JetBrains Toolbox via the CLI, even though it's a GUI app. Here's how:
1. Close the app if open
2. Set auto-updating to true in the configuration
3. Start the app in the background
4. Monitor the logs for possible updates
5. Wait for updates to complete, or until 10 seconds pass without the log mentioning updates.
6. Quit
7. Reset the configuration
8. Restart the app if it was open

Here's my POC python script:
```py
import json
import time
from pathlib import Path
from subprocess import Popen
from time import sleep

import psutil


# TODO: make Windows-friendly
dir = Path("~/.local/share/JetBrains/Toolbox/").expanduser()
bin = dir / "bin/jetbrains-toolbox"
# conf = dir / ".settings.json"
confs = dir / "channels"
log = dir / "logs/toolbox.log"


def add_update_param(d):
    if "autoUpdate" in d["channel"]:
        if d["channel"]["autoUpdate"] == True:
            # TODO: skip this one then?
            #  Make sure to not reset it with remove_update_param then
            raise Exception("This tool is already updated automatically!")
        else:
            raise Exception("'channel.autoUpdate' is not true! This should not be possible")

    d["channel"]["autoUpdate"] = True


def remove_update_param(d):
    d["channel"].pop("autoUpdate")


def update_conf(file: Path, operation):
    with file.open("r+") as file:
        d = json.load(file)
        operation(d)
        file.seek(0)
        json.dump(d, file)
        file.truncate()


def update_all_confs(operation):
    for file in confs.iterdir():
        update_conf(file, operation)


def main():
    update_all_confs(add_update_param)
    toolbox_was_open = None

    try:
        # Kill toolbox if open
        processes = list(filter(lambda p: p.name() == "jetbrains-toolbox", psutil.process_iter()))
        if len(processes) == 0:
            toolbox_was_open = False
        elif len(processes) == 1:
            process, = processes
            process.terminate()
            process.wait()
            toolbox_was_open = True
        else:
            raise Exception("Multiple toolboxes open? This should not be possible")

        p = Popen([bin, "--minimize"])

        updates = 0
        correct_checksums_expected = 0
        start_time = time.time()
        with log.open("r") as file:
            file.seek(0,2)  # Go to the end of file
            while True:
                # TODO: Unfortunately there is no log message indicating there are no updates; so waiting is necessary it looks like.
                #  Unless we can do something with "Downloaded fus-assistant.xml"? Maybe shorten the time to 1/2 seconds after that message, seems to be fine.
                if updates == 0 and start_time + 10 < time.time():
                    # 10 seconds without recognizing updating, so presuming no update.
                    print("No updates found.")
                    break
                curr_position = file.tell()
                line = file.readline()
                if not line:
                    file.seek(curr_position)
                    sleep(0.1)
                else:
                    # TODO: downloading two updates (one large download, one no download) is currently broken; but that never happens in practice anyway.
                    # If the download is already there, it won't say "Downloading from", but immediately "Correct checksum for".
                    if "Correct checksum for" in line or "Downloading from" in line:
                        if "Correct checksum for" in line and correct_checksums_expected > 0:
                            correct_checksums_expected -= 1
                            continue
                        # Update started
                        print("Found an update, waiting until it finishes...")
                        updates += 1
                        if "Downloading from" in line:
                            # We expect "Correct checksum for" to be broadcast exactly once after the downloading from.
                            correct_checksums_expected += 1
                    elif "Show notification" in line:
                        # Update finished
                        updates -= 1
                        if updates == 0:
                            print("Update finished, exiting...")
                            sleep(2)  # Letting it finish up
                            break
                        else:
                            print("Update finished, waiting for other update(s) to finish")

        p.terminate()
        p.wait()

    finally:  # Still restore original state after crash
        update_all_confs(remove_update_param)
        if toolbox_was_open:
            print("Re-opening...")
            Popen([bin, "--minimize"])
            # Simply exiting now will not quit the toolbox process


if __name__ == "__main__":
    main()
```

My question: is it okay to integrate this into Topgrade, and have it enabled by default? I personally think it's fine, and the convenience factor for people using Toolbox outweighs the possible issues, but let me know what you think.

[JetBrains Toolbox](https://www.jetbrains.com/toolbox-app/), properietary and not Open-Source.
All? My script supports Linux only currently, but this can easily be expanded.
Only run when the binary exists
See the script
No
No

ref #1014; my need for that is much less now that this is possible.

GideonBear:
Closes #1056

## What does this PR do
Adds a new step: JetBrains Toolbox

## Standards checklist

## For new steps

  the underlying command

Hello! I ported the code to Rust and everything's working. I've got a few questions about how best to integrate it into Topgrade though:
### What should be localized? (the `t!` strings don't actually have localization right now)
Is it okay if I leave `jetbrains-toolbox-updater` itself unlocalized? Anything else apart from what I've already wrapped in `t!` that should be localized?
### How do I handle the errors?
Currently, I'm `println!`ing the error and the debug representation, then returning `StepFailed`. Is this acceptable?

Own todo:
- [x] Remove/add `println!`s from `jetbrains-toolbox-updater`: `Found 3 processes`, possible good places for additions?
- [ ] Add translations
- [ ] Add Windows support


src/steps/generic.rs
```diff
@@ -1344,3 +1345,34 @@ pub fn run_zigup(ctx: &ExecutionContext) -> Result<()> {
 
     Ok(())
 }
+
+pub fn run_jetbrains_toolbox(_ctx: &ExecutionContext) -> Result<()> {
+    let installation = find_jetbrains_toolbox();
+    match installation {
+        Err(FindError::NotFound) => {
+            // Skip
+            Err(SkipStep(format!("{}", t!("No JetBrains Toolbox installation found"))).into())
```
SteveLauC, line 1354: > the t! strings don't actually have localization right now

That's because the message wrapped by `t!` has not been added to https://github.com/topgrade-rs/topgrade/blob/main/locales/app.yml:

```yml
"No JetBrains Toolbox installation found":
  en: "No JetBrains Toolbox installation found"
  es: "No se encontró ninguna instalación de JetBrains Toolbox"
  zh_TW: "找不到 JetBrains Toolbox 安裝"
```

src/steps/generic.rs
```diff
@@ -1344,3 +1345,34 @@ pub fn run_zigup(ctx: &ExecutionContext) -> Result<()> {
...
+            println!("jetbrains-toolbox-updater had an unexpected error during finding: {e:?}");
+            Err(StepFailed.into())
+        }
+        Ok(installation) => {
+            print_separator("JetBrains Toolbox");
+
+            match update_jetbrains_toolbox(installation) {
+                Err(e) => {
+                    // Unexpected error
+                    println!("jetbrains-toolbox-updater had an unexpected error during updating: {e:?}");
```
SteveLauC, line 1371: > Anything else apart from what I've already wrapped in t! that should be localized?

These 2 `println!()` need to be localized:)

SteveLauC (commented):
> Is it okay if I leave jetbrains-toolbox-updater itself unlocalized?

Yeah

> How do I handle the errors?
>
> Currently, I'm println!ing the error and the debug representation, then returning StepFailed. Is this acceptable?

The current approach looks good to me, the PR is ready for merge once the i18n issue is resolved:)

GideonBear:
@SteveLauC I resolved the issues. Do the translations look okay? They were done with Google Translate.
Please don't hesitate to ping me if there are ever issues reported with this step, and I will try to fix them as soon as possible. Thanks for letting me include this! I hope that it will benefit a lot more people than just me.

SteveLauC (approved):
Thanks!

Issue #1075, opened 2025-03-19
Emacs step uses unmaintained Paradox package manager


aveyrenc:
## Erroneous Behavior

Commit 2585156 removes embedded `package.el` support and only leaves [Paradox](https://github.com/Malabarba/paradox), which is now unmaintained.

The commit message points to #942 which is about uv, so I don't figure why `package.el` support was removed.

## Expected Behavior

At least support Emacs' embedded package manager. I'm not familiar with Straight / Elpaca so I don't know if we should support them also.

I can send a pull request to reverse current behavior, so remove Paradox support and re-enable `package.el`.

## Steps to reproduce

```
topgrade --only emacs
```

## Did you run topgrade through `Remote Execution`

- [ ] Yes
- [X] No

## Additional Details
- Operation System/Version
  

Tested on macOS and Ubuntu.

- Installation
  

Homebrew

- Topgrade version (`topgrade -V`)

16.0.2

SteveLauC:
> The commit message points to https://github.com/topgrade-rs/topgrade/issues/942 which is about uv, so I don't figure why package.el support was removed.

This Topgrade repo is a fork of https://github.com/r-darwish/topgrade, that commit was done there


That commit was for this issue, could you please take a look at it https://github.com/r-darwish/topgrade/issues/757? I know nothing about Emacs so I cannot help:<

aveyrenc:
Thank you for the pointer, now that makes sense.

They were waiting for the availability of the `package-upgrade-all` function which is now part of Emacs since 29.1, out in July 2023.

I will send a PR which will change the behavior this way: check if the upgrade function is available and run it. If not, show a message to the user.

SteveLauC:
Close as this is fixed by #1076

PR #1076, opened 2025-03-20
fix:(emacs): fix issue #1075

aveyrenc:
## What does this PR do

Fix issue #1075. We'll use Emacs' embedded package upgrade functionality if available.

I don't know if we can use the i18n system to translate the string inside `emacs.el`.

Tested with Emacs 29.3 on Linux and Emacs 30.1 on macOS.

### DIFF
--- a/src/steps/emacs.el
+++ b/src/steps/emacs.el
@@ -1,9 +1,4 @@
-(when (fboundp 'paradox-upgrade-packages)
-    (progn
-      (unless (boundp 'paradox-github-token)
-        (setq paradox-github-token t))
-      (paradox-upgrade-packages)
-      (princ
-       (if (get-buffer "*Paradox Report*")
-           (with-current-buffer "*Paradox Report*" (buffer-string))
-         "\nNothing to upgrade\n"))))
+(when (featurep 'package)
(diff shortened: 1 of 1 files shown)


SteveLauC:
> I don't know if we can use the i18n system to translate the string inside emacs.el.

No need to do i18n outside of Topgrade

SteveLauC (commented):
> They were waiting for the availability of the package-upgrade-all function which is now part of Emacs since 29.1, out in July 2023.

For users who are still running an older version of Emacs, there is currently no appropriate way to do the update for them, right?

aveyrenc:
> > They were waiting for the availability of the package-upgrade-all function which is now part of Emacs since 29.1, out in July 2023.
> 
> For users who are still running an older version of Emacs, there is currently no appropriate way to do the update for them, right?

Indeed, they were automating actions you usually do interactively, like open the packages menu, select all the upgradable packages, validate, etc. and it wasn't reliable.

Maybe we can assume that most of Topgrade users will use a relatively recent Emacs version? For older Emacs versions, the code in this PR detects that the upgrade functionality is not available and just exists after displaying a message.

SteveLauC (approved):
Thanks!

PR #1077, opened 2025-03-21
Update jetbrains-toolbox-updater

GideonBear:
## What does this PR do
I've bumped `jetbrains-toolbox-updater`, because I added a feature that should be included in the next release. It avoids accidentally permanently modifying Toolbox configuration when an error occurs.


SteveLauC (approved):
Thanks!

Issue #1088, opened 2025-03-27
Winget Silent Install


The-RedWizard:
Winget has the option to perform it's upgrade silently, by using the `-h` or `--silent` flag as part of the upgrade command `winget upgrade --silent`. Under the Windows section in the config file, we should add a new config property that allows us to specify `silent = true`, and its default value should be set to `true`. 

Without the Silent option specified, the installer applications are launched in an unattended state (typically non-interactive, but visible to the end user), stealing focus from whatever active window you might be using at the time it was launched.

PR #1089, opened 2025-03-27
added silent install option for winget

The-RedWizard:
This PR implements issue #1088 

## What does this PR do

Before I describe what this PR does, let me define "silent" in the Windows context. On Windows, historically, silent or quite when discussing installations means the installer does not present a GUI to the user and installs the software without interaction. **It does not mean suppressing command line output**. 

This PR makes Winget perform its upgrade silently, unless configured to not, by using the new `winget_install_silent` configuration option and setting it to `false` in the config file. Without this option set to `false`, Windows installers will launch an unattended installer window for the user. With this option set to `true` or left default, it will use the `--silent` command line argument tell Winget to attempt to run all application installers silently. 

**It should be noted:** Support for silent installations is highly dependent on the maintainer of the Windows Application. Software using the `.msi` installer standard from Microsoft all support silent installations. Packages using InstallShield, Inno Setup, or Nullsoft installers also support silent installations. Vendors packaging their applications with an alternative installer solution will have mixed support. So, with all that said, **this feature does not guarantee all packages will install without presenting a GUI**, only that **all packages that support silent GUI-less installs will do so by default.**

## Standards checklist

### DIFF
--- a/src/steps/os/windows.rs
+++ b/src/steps/os/windows.rs
@@ -42,20 +42,12 @@ pub fn run_winget(ctx: &ExecutionContext) -> Result<()> {
 
     print_separator("winget");
 
-    // check if silent install is enabled
+    let mut args = vec!["upgrade", "--all"];
     if ctx.config().winget_silent_install() {
-        // execute with silent command
-        ctx.run_type()
-            .execute(winget)
-            .args(["upgrade", "--all", "--silent"])
-            .status_checked()
-    } else {
-        // execute without silent command
-        ctx.run_type()
-            .execute(winget)
-            .args(["upgrade", "--all"])
-            .status_checked()
+        args.push("--silent");
     }
+
+    ctx.run_type().execute(winget).args(args).status_checked()
 }
 
 pub fn run_scoop(ctx: &ExecutionContext) -> Result<()> {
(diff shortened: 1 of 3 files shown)


The-RedWizard:
Noticed that the formatting check failed. I have corrected the formatting issue.

src/steps/os/windows.rs
```diff
@@ -42,10 +42,20 @@ pub fn run_winget(ctx: &ExecutionContext) -> Result<()> {
...
+            .execute(winget)
+            .args(["upgrade", "--all", "--silent"])
+            .status_checked()
+    } else {
+        // execute without silent command
+        ctx.run_type()
+            .execute(winget)
+            .args(["upgrade", "--all"])
+            .status_checked()
+    }
```
SteveLauC, line 58: Nit: Let's do this to avoid a small amount of duplication:)

``` suggestion
    let mut args = vec!["upgrade", "--all"];
    if ctx.config().winget_silent_install() {
        args.push("--silent");
    }

    ctx.run_type().execute(winget).args(args).status_checked()
```

SteveLauC: The suggested code is not formatted, you may need to run `cargo fmt` yourself:D

The-RedWizard:
Those changes make sense, not sure why I didn't think to do it that way. It looks like the formatting is good, I pulled the changes and ran `cargo fmt` and it produced no changes. Thanks!

SteveLauC (approved):
Thanks!

PR #1090, opened 2025-03-28
added a check to disable dnf upgrading when

Pyro57000:
## What does this PR do
Adds a check that skips dnf upgrades if rpm-ostree is enabled.
The DNF upgrade command will always fail on universal blue based distros.
This check makes sure that it just doesn't run when rpm-ostree is the enabled upgrade mechanism.

## Standards checklist

- [x] If this PR introduces new user-facing messages they are translated (N/A)
 
## For new steps

- [  ] *Optional:* Topgrade skips this step where needed
    - This hasn't been tested since I don't have a platform that would skip this, but I think the logic stands, its a pretty minimal change.
  the underlying command

Did this for my AuroraOS system since its based on universal blue, it should fix an issue where the dnf upgrade step will be called even if rpm-ostree is enabled.

### DIFF
--- a/src/steps/os/linux.rs
+++ b/src/steps/os/linux.rs
@@ -237,25 +237,26 @@ fn upgrade_redhat(ctx: &ExecutionContext) -> Result<()> {
         }
     };
 
-    let sudo = require_option(ctx.sudo().as_ref(), get_require_sudo_string())?;
-    let mut command = ctx.run_type().execute(sudo);
-    command
-        .arg(which("dnf").unwrap_or_else(|| Path::new("yum").to_path_buf()))
-        .arg(if ctx.config().redhat_distro_sync() {
-            "distro-sync"
-        } else {
-            "upgrade"
-        });
+    if !ctx.config().rpm_ostree(){
+        let sudo = require_option(ctx.sudo().as_ref(), get_require_sudo_string())?;
+        let mut command = ctx.run_type().execute(sudo);
+        command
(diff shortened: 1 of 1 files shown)


SteveLauC:
Is this an immutable system? If so, `dnf` should not be invoked at all, it should run instead:

https://github.com/topgrade-rs/topgrade/blob/a601d8429dd351f47f128685b6a5eb8472459fa7/src/steps/os/linux.rs#L304-L317

SteveLauC:
You should modify this function to fix the Linux distro detection mechanism:)

https://github.com/topgrade-rs/topgrade/blob/a601d8429dd351f47f128685b6a5eb8472459fa7/src/steps/os/linux.rs#L50-L106

Pyro57000:
> You should modify this function to fix the Linux distro detection mechanism:)
> 
> https://github.com/topgrade-rs/topgrade/blob/a601d8429dd351f47f128685b6a5eb8472459fa7/src/steps/os/linux.rs#L50-L106

I can take a look at that, it is indeed an immutable system, but dnf is still definitely trying to be called.

GideonBear:
> I can take a look at that, it is indeed an immutable system, but dnf is still definitely trying to be called.

Like @SteveLauC said:
This is a bug (or rather, an OS missing support), but the way you're trying to solve it is incorrect, we should fix the OS detection on your system, as the rpm-ostree functionality is already there. I'm going to close this PR as the PR is not useful in this state, but please open an issue!

PR #1104, opened 2025-04-08
Sort `Step` enum and keep it sorted in the `ci.yml` workflow

GideonBear:
## What does this PR do
Currently it's not correctly sorted, and I can't stand it being *almost* sorted. Feel free to close this if you disagree that this should be ran in CI, but IMO if it is enforced, it should be checked :)
The workflow was written using ChatGPT (my bash/awk/grep/sed skills are not good enough), but it works and it looks fine.

## Standards checklist


SteveLauC (approved):
Looks good, thanks!

PR #1110, opened 2025-04-08
Add `output_changed_message!`, replace some `.expect`s

GideonBear:
## What does this PR do
* Adds `output_changed_message!` to replace duplicated "please file an issue to Topgrade" string
* Replaces some `.expect`s and other error raisers with `.wrap_err`
  * I haven't replaced all of them, but these are some obvious ones

## Standards checklist

## Questions
Some questions:
1. Is the text of `output_changed_message!` okay like this?
2. Apparently I'm supposed to use `.context` instead of `.wrap_err`? At least in the `Option` case at `NIX_VERSION_REGEX.captures`?

### DIFF
--- a/src/steps/generic.rs
+++ b/src/steps/generic.rs
@@ -22,12 +22,12 @@ use crate::execution_context::ExecutionContext;
 use crate::executor::ExecutorOutput;
 use crate::terminal::{print_separator, shell};
 use crate::utils::{self, check_is_python_2_or_shim, get_require_sudo_string, require, require_option, which, PathExt};
-use crate::Step;
 use crate::HOME_DIR;
 use crate::{
     error::{SkipStep, StepFailed, TopgradeError},
     terminal::print_warning,
 };
+use crate::{output_changed_message, Step};
 
 #[cfg(target_os = "linux")]
 pub fn is_wsl() -> Result<bool> {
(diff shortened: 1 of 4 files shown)


SteveLauC:
Thanks for doing this!

> Is the text of output_changed_message! okay like this?

Yeah, I think it is good.

> Apparently I'm supposed to use .context instead of .wrap_err? At least in the Option case at NIX_VERSION_REGEX.captures?

What's the difference between these 2? From the doc, [`wrap_err`](https://docs.rs/eyre/latest/eyre/trait.WrapErr.html#tymethod.wrap_err) is a method that is exclusive to `eyre` (the crate that Topgrade uses for error handling), and `with_context()` is same as `wrap_err()`, it is created for simply keeping compatibility with `anyhow`

GideonBear:
Heres the answer to both of our questions:
https://docs.rs/eyre/latest/eyre/trait.ContextCompat.html

GideonBear:
Specifically on `Options` youre supposed to use something else

SteveLauC:
Yeah, then using `with_context` looks better

SteveLauC (approved):
Thanks!

GideonBear:
Added a runtime counterpart, which was necessary because `run_vscode_compatible` is generic, and using constant generics in constant expressions is unfortunately a nightly feature.

SteveLauC:
> Added a runtime counterpart, which was necessary because run_vscode_compatible is generic, and using constant generics in constant expressions is unfortunately a nightly feature.

I think you can use `format!()` rather than `concat!()` directly in `macro_rules! output_changed_message`, generally, Topgrade does not care about performance

SteveLauC:
> It is especially bad on performance because it is used in wrap_err, not in wrap_err_with, so it would evaluate the format! every time, regardless of if it's needed.

Right, can we use `wrap_err_with()`  to do lazy evaluation then?

> Mostly I'm also just doing this to learn Rust better, so I will follow best practices even when it is not strictly necessary within the context :)

Understand, I just want to make Topgrade easier to maintain

> Do you want me to make them both execute at runtime?

If feasible, I think we should:

1. Replace `concat!()` with `format!()` within the impl of the macro so that we have only 1 utility macro/function
2. Use `wrap_err_with()` to do lazy evaluation

GideonBear:
I understand, done

SteveLauC (approved):
Thanks!

PR #1113, opened 2025-04-09
🌐 i18n(app): add new translation strings for windows and .net features

tdslot:
I tested the new release version 16.0.3 and found that some messages on screen were not translated.
So I went through the code and found that some messages were missing in the translation file `app.yaml` so I added. Maybe it is some automated way but I did it manually. 

- add translations for windows update admin prompt message
- add translations for .NET package update failure message
- add translations for tmux arguments parsing error

🐛 fix(config): use translated strings in error messages

- replace hardcoded error messages with translated versions in config parsing
- update .NET package update error message to use translations
- update windows update admin prompt to use translations

### DIFF
--- a/src/steps/os/windows.rs
+++ b/src/steps/os/windows.rs
@@ -211,7 +211,10 @@ pub fn windows_update(ctx: &ExecutionContext) -> Result<()> {
     print_separator(t!("Windows Update"));
 
     if powershell.supports_windows_update() {
-        println!("{}", t!("The installer will request to run as administrator, expect a prompt."));
+        println!(
+            "{}",
+            t!("The installer will request to run as administrator, expect a prompt.")
+        );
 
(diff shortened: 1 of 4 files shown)


src/steps/os/windows.rs
```diff
@@ -211,7 +211,10 @@ pub fn windows_update(ctx: &ExecutionContext) -> Result<()> {
     print_separator(t!("Windows Update"));
 
     if powershell.supports_windows_update() {
-        println!("The installer will request to run as administrator, expect a prompt.");
```
SteveLauC, line 214: This should be translated as it is a user-facing message. Other changes are not needed because they are for developers

tdslot: Understood so probably this can be closed.

PR #1122, opened 2025-04-12
Add "uv python" step

GideonBear:
## What does this PR do
Adds a disabled-by-default "uv python" step, that checks for new available versions and installs them automatically.
Is this acceptable?

## Standards checklist

## For new steps

  the underlying command

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -349,9 +349,9 @@ checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
 
 [[package]]
 name = "bytes"
-version = "1.6.0"
+version = "1.10.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


GideonBear:
There are plans to add something similar within `uv` (https://github.com/astral-sh/uv/issues/7325, https://github.com/astral-sh/uv/issues/7892, https://github.com/astral-sh/uv/issues/7287, https://github.com/astral-sh/uv/issues/8028, https://github.com/astral-sh/uv/issues/4718), but that's probably still a long time away.

GideonBear:
I realize that `uv tool install --reinstall` does not work with `uv tool install --with`. Potentially this is something to be added in uv itself?

GideonBear:
> but that's probably still a long time away.

That aged badly... Astral is way quicker with this stuff than I anticipated :joy:

I'll revisit this soon, to add `uv python --upgrade`, and maybe still a step for upgrading the major version?

PR #1123, opened 2025-04-13
Fix "Only one instance of <IDE> can be run at a a time." error

GideonBear:
## What does this PR do
* Deduplicates `run_android_studio`
* Records output of `<ide> update` command to check for "Only one instance of <IDE> can be run at a a time.", and skip it in that case

## Standards checklist

### DIFF
--- a/src/steps/generic.rs
+++ b/src/steps/generic.rs
@@ -1499,21 +1499,22 @@ pub fn run_jetbrains_toolbox(_ctx: &ExecutionContext) -> Result<()> {
     }
 }
 
-fn run_jetbrains_ide(ctx: &ExecutionContext, bin: PathBuf, name: &str) -> Result<()> {
-    print_separator(format!("JetBrains {name} plugins"));
+fn run_jetbrains_ide_generic<const IS_JETBRAINS: bool>(ctx: &ExecutionContext, bin: PathBuf, name: &str) -> Result<()> {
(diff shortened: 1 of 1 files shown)


src/steps/generic.rs
```diff
@@ -1543,21 +1543,53 @@ pub fn run_jetbrains_toolbox(_ctx: &ExecutionContext) -> Result<()> {
...
+    // "Only one instance of RustRover can be run at a time."
+    if stdout.contains("Only one instance of ") && stdout.contains(" can be run at a time.") {
+        // It's always paired with status code 1
+        if output
+            .status
+            .code()
+            .ok_or_eyre("Failed to get status code; was killed with signal")?
+            != 1
+        {
+            return Err(eyre!("Expected status code 1 ('Only one instance of <IDE> can be run at a time.'), but received successful exit code. Output: {output:?}"));
```
SteveLauC, line 1571: Nit: feels like the exit code is not guaranteed to be 0, i.e., the successful exit code 🤔

SteveLauC (commented):
Looks good to me

SteveLauC (approved):

src/steps/generic.rs
GideonBear: Good catch

SteveLauC (approved):

Issue #1141, opened 2025-04-28
Quicker release schedule


GideonBear:
Hi! Due to the nature of Topgrade there are often issues and crashes that are resolved with a change in Topgrade's code. This means that the end user only gets these changes at the next release, or if they install it from the `main` branch (`cargo install --git https://github.com/topgrade-rs/topgrade.git`). This can result in duplicate issues (e.g. #809, #939, https://github.com/GideonBear/jetbrains-toolbox-updater/issues/4) after the issue is already fixed on the `main` branch.
Is there a reason we don't do quicker releases? It would reduce the need for people to compile Topgrade from source, and increase the value for people that don't.
If we automate the release process as much as possible, we could do a release after every PR that fixes a major issue in a step. IMO that's just necessary for a tool like Topgrade.

SteveLauC:
Currently, Topgrade is released upon request. I believe a regular release cycle is better, and we should do that. cc @lucaspar (related discussion: https://github.com/topgrade-rs/topgrade/discussions/917)

> Is there a reason we don't do quicker releases?

I would say no? But there are cases where I do not have time to take care of this project, and thus a release will get delayed repeatedly

> If we automate the release process as much as possible, we could do a release after every PR that fixes a major issue in a step.

Yeah, I agree

GideonBear:
Ref #844

aslowwriter:
I am fine with the current release schedule, but I also have quite some experience with release automation so I thought I'd mention that I'd be willing to help if necessary. 

I will also mention that https://github.com/release-plz/release-plz is becoming a very well respected ready made solution for this. Not saying you'd have to use it, but it might be worth looking into :)

SteveLauC:
Hi @savente93, thanks for your interest:)

> I will also mention that https://github.com/release-plz/release-plz is becoming a very well respected ready made solution for this. Not saying you'd have to use it, but it might be worth looking into :)

That looks cool, I will take a look at it:)

a5ehren:
To add on to this, the release process md says that uploading to crates.io is hard to do in GH Actions, but that has not been my experience - you just need to pop your token into the secrets vault and get it out in an environment with the toolchain. I have an automated crate upload [here, for reference](https://github.com/a5ehren/compiledb-rs/blob/main/.github/workflows/release.yml#L25). 

Checking the tag to compare to the Cargo.toml was something I added as a check on myself forgetting to do it, but in theory the bump could be automated.

GideonBear:
[...]

PR #1144, opened 2025-05-02
Streamline release workflows

GideonBear:
## What does this PR do
Ref #1141
* Reduces inconsistency in release workflow triggers. I cannot really test this, but I looked at the GitHub docs and I think this still does what we want. Can you check?

## Questions
@SteveLauC 
- [ ] `And the CI will publish the new binary to: 3. Homebrew (seems that this is not working correctly)`; is this still the case / why is this not working correctly? I cannot find an open issue for this.
- [ ] `Manually release it to Crates.io`; isn't this already automatically done? See https://github.com/topgrade-rs/topgrade/blob/main/.github/workflows/release_to_crates_io.yml. Although this comment was added after the `release_to_crates_io.yml` workflow was created.

## Standards checklist


GideonBear:
Closing in favor of #1333

PR #1146, opened 2025-05-03
fix(tmux): support all default `tpm` locations (xdg and both hardcoded locations)

### LINKED ISSUE #1142: Path to update_plugins is wrong when TMUX_PLUGIN_MANAGER_PATH is set
Thank you for your work, topgrade is really great!

There is a small issue with the tmux step:

## Erroneous Behavior

It seems to me that the path to the `update_plugins` executable from TPM is set wrongly when  TMUX_PLUGIN_MANAGER_PATH is set. Commit 3f2efdd78e981e42c493594b2b5c680fe54320df in pull #835 made a change (I believe based on the feedback in issue #832 setting appending to TMUX_PLUGIN_MANAGER_PATH the path:
`bin/install_plugins/tpm/bin/update_plugins`.

That seems like a custom path specific to that user in the aforementioned issue. In fact, `install_plugins` is normally an executable, not a directory.

Image

## Expected Behavior

The TMUX_PLUGIN_MANAGER_PATH is meant to point to the path that contains the `tpm` plugin. After that the path to the `update_plugins` executable should be:
`tpm/bin/update_plugins`.

These docs explain it:
[Updating plugins from command line](https://github.com/tmux-plugins/tpm/blob/master/docs/managing_plugins_via_cmd_line.md#updating-plugins)
[Changing plugins install dir](https://github.com/tmux-plugins/tpm/blob/master/docs/changing_plugins_install_dir.md)

Quick fix is to remove the `bin/install_plugins` part. :)

## Steps to reproduce

NA

## Additional Details
- Operation System/Version
Pop!_OS 22.04 LTS x86_64
 
topgrade 16.0.3

noirbizarre:
## What does this PR do

This PR fixes `TMUX_PLUGIN_MANAGER_PATH` handling (fix #1142) and add support for all default plugin path cases in the same order as tmux and ([tpm](https://github.com/tmux-plugins/tpm/blob/master/tpm#L24-L35)):
- `XDG_CONFIG` path
- hardcoded `~/.config/tmux` dir
- hardcoded `~/.tmux` dir

## Standards checklist

## For new steps

  the underlying command

### DIFF
--- a/src/steps/tmux.rs
+++ b/src/steps/tmux.rs
@@ -5,16 +5,17 @@ use std::process::Command;
 use color_eyre::eyre::eyre;
 use color_eyre::eyre::Context;
 use color_eyre::eyre::Result;
+use etcetera::base_strategy::BaseStrategy;
 
 use crate::command::CommandExt;
 use crate::config::TmuxConfig;
 use crate::config::TmuxSessionMode;
 use crate::terminal::print_separator;
-use crate::HOME_DIR;
 use crate::{
     execution_context::ExecutionContext,
     utils::{which, PathExt},
 };
+use crate::{HOME_DIR, XDG_DIRS};
 
 use rust_i18n::t;
 #[cfg(unix)]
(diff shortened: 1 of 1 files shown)


src/steps/tmux.rs
```diff
@@ -23,10 +24,21 @@ use std::os::unix::process::CommandExt as _;
...
+            // Otherwise, use the default XDG location `~/.config/tmux/plugins/tpm/bin/update_plugins`
+            // or fallback on the standard default location `~/.tmux/plugins/tpm/bin/update_plugins`
+            #[cfg(unix)]
+            let xdg_path = XDG_DIRS.config_dir().join("tmux/plugins/tpm/bin/update_plugins");
+            #[cfg(windows)]
+            let xdg_path = HOME_DIR.join(".config/tmux/plugins/tpm/bin/update_plugins");
+            if xdg_path.exists() {
+                xdg_path
+            } else {
+                HOME_DIR.join(".tmux/plugins/tpm/bin/update_plugins")
```
GideonBear: Can the string `plugins/tpm/bin/update_plugins` be put in a variable?

GideonBear (approved):

src/steps/tmux.rs
noirbizarre: Done
Actually, I introduced 2 constants for standard (sub)paths

GideonBear (approved):

src/steps/tmux.rs
GideonBear: Perfect, thank you!

PR #1147, opened 2025-05-05
Adds framework for listing updated components in the summary

### LINKED ISSUE #943: List updated programs / components post upgrade in a bullet point list
I'm finding that `topgrade` is extremely useful, but it's a bit easy to miss some things that are updated that are important. Having a final bullet point list of things that have been upgraded post upgrade would be extremely helpful to see so as to note if something critical has been updated that needs attention.

An example of this is when the awesome `eloston-chromium` is updated and your browser tabs will start dying inexplicably unless you close and reopen the browser. Other similar tools and browsers behave in this fashion where they need to be quit and restarted and without an easy to reference list it's very likely one will not look closely enough and get caught up for a brief period wondering why things are blowing up around them...

GideonBear:
## What does this PR do
Closes #943 (I am planning to open a tracking issue for implementing this for all steps where it is possible if this is merged.)
* Adds the framework for listing updated components in the summary
* Implements that for `uv` (self-updating only for now)
* Adds references to this to the "new step" issue and PR templates.

What do you think @SteveLauC? This is non-disruptive for steps that don't yet or can't support it, but opens the possibility to implement this for steps where it is very useful. Note that for some steps, machine-readable output is available (e.g. `apt-get`) and can be used instead of regexes. We can also PR to the tools themselves to support outputting in a machine-readable format. (maybe a good idea for `uv`, which already supports `--output-format=json` on some commands)

Example: (`cinnamon_spices` included to show what it looks like for a step that does not support it (yet))
```console
$ cargo run -- --only uv --only cinnamon_spices
── 11:18:45 - Cinnamon spices ──────────────────────────────────────────────────
There are no Spice updates

── 11:18:46 - uv ───────────────────────────────────────────────────────────────
info: Checking for updates...
success: Upgraded uv from v0.7.1 to v0.7.2! https://github.com/astral-sh/uv/releases/tag/0.7.2
Nothing to upgrade

── 11:18:50 - Summary ──────────────────────────────────────────────────────────
Cinnamon spices: OK
uv: OK: Updated:
- (self-update) uv from v0.7.1 to v0.7.2
$ cargo run -- --only uv --only cinnamon_spices
── 11:18:52 - Cinnamon spices ──────────────────────────────────────────────────
There are no Spice updates

── 11:18:53 - uv ───────────────────────────────────────────────────────────────
info: Checking for updates...
success: You're on the latest version of uv (v0.7.2)
Nothing to upgrade

── 11:18:53 - Summary ──────────────────────────────────────────────────────────
Cinnamon spices: OK
uv: OK: No updates found
```

## Standards checklist

### DIFF
--- a/src/steps/generic.rs
+++ b/src/steps/generic.rs
@@ -9,7 +9,7 @@ use color_eyre::eyre::Context;
 use color_eyre::eyre::Result;
 use jetbrains_toolbox_updater::{find_jetbrains_toolbox, update_jetbrains_toolbox, FindError};
 use lazy_static::lazy_static;
-use regex::bytes::Regex;
+use regex::Regex;
 use rust_i18n::t;
 use semver::Version;
 use tempfile::tempfile_in;
@@ -18,6 +18,7 @@ use tracing::{debug, error};
 use crate::command::{CommandExt, Utf8Output};
 use crate::execution_context::ExecutionContext;
 use crate::executor::ExecutorOutput;
+use crate::report::UpdatedComponent;
 use crate::terminal::{print_separator, shell};
 use crate::utils::{check_is_python_2_or_shim, get_require_sudo_string, require, require_option, which, PathExt};
 use crate::HOME_DIR;
@@ -1234,7 +1235,8 @@ pub fn run_poetry(ctx: &ExecutionContext) -> Result<()> {
         use std::os::unix::ffi::OsStrExt;
 
         lazy_static! {
-            static ref SHEBANG_REGEX: Regex = Regex::new(r"^#![ \t]*([^ \t\n]+)(?:[ \t]+([^\n]+)?)?").unwrap();
+            static ref SHEBANG_REGEX: regex::bytes::Regex =
+                regex::bytes::Regex::new(r"^#![ \t]*([^ \t\n]+)(?:[ \t]+([^\n]+)?)?").unwrap();
         }
 
         let script = fs::read(poetry)?;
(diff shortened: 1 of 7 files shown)


src/steps/generic.rs
```diff
@@ -2,7 +2,7 @@ use color_eyre::eyre::Context;
 use color_eyre::eyre::Result;
 use color_eyre::eyre::{eyre, OptionExt};
 use jetbrains_toolbox_updater::{find_jetbrains_toolbox, update_jetbrains_toolbox, FindError};
-use regex::bytes::Regex;
+use regex::Regex;
```
lnksz, line 3: Why?

src/steps/generic.rs
```diff
@@ -1329,10 +1331,12 @@ pub fn run_poetry(ctx: &ExecutionContext) -> Result<()> {
     ctx.execute(&poetry).args(["self", "update"]).status_checked()
 }
 
-pub fn run_uv(ctx: &ExecutionContext) -> Result<()> {
+pub fn run_uv(ctx: &ExecutionContext) -> Result<Vec<UpdatedComponent>> {
```
lnksz, line 1592: Why not `UpdatedComponents` ?

src/steps/generic.rs
GideonBear: I needed to use `Regex`, and `regex::Regex` feels much more logical to have `use`d than `regex::bytes::Regex`

src/steps/generic.rs
GideonBear: To move all boilerplate out of `run_*` functions, it avoids an unnecessary `UpdatedComponents::new` when returning.

GideonBear:
See discussion in #943

PR #1163, opened 2025-06-07
Update Yazi upgrade step to use ya pkg.

slowsage:
## What does this PR do
Currently we get this message:
WARNING: `ya pack` is deprecated, use the new `ya pkg` instead. See https://github.com/sxyazi/yazi/pull/2770 for more details.

This PR removes this extra line of output.
## Standards checklist

## For new steps

  the underlying command

### DIFF
--- a/src/steps/generic.rs
+++ b/src/steps/generic.rs
@@ -1658,5 +1658,5 @@ pub fn run_yazi(ctx: &ExecutionContext) -> Result<()> {
 
     print_separator("Yazi packages");
 
-    ctx.run_type().execute(ya).args(["pack", "-u"]).status_checked()
+    ctx.run_type().execute(ya).args(["pkg", "upgrade"]).status_checked()
 }


wiraki:
> I think we need to keep the previous code for people who are still using Yazi (<25.5.28) as this `pkg` sub-command was newly introduced in that version

And what about those that have updated Yazi? There is a high chance that people who want to use the `yazi` step in topgrade also already keep their Yazi version updated via topgrade.

I propose that if a user is using older versions and want to stick to that, they can always have a custom command for `ya pack -u`, but the "official" topgrade yazi step should be up-to-date.

GideonBear:
> > I think we need to keep the previous code for people who are still using Yazi (<25.5.28) as this `pkg` sub-command was newly introduced in that version
> 
> And what about those that have updated Yazi? There is a high chance that people who want to use the `yazi` step in topgrade also already keep their Yazi version updated via topgrade.
> 
> I propose that if a user is using older versions and want to stick to that, they can always have a custom command for `ya pack -u`, but the "official" topgrade yazi step should be up-to-date.

What I presume @SteveLauC means here, is to check the yazi version and choose the method based on that; like we do with uv.

slowsage:
> What I presume @SteveLauC means here, is to check the yazi version and choose the method based on that; like we do with uv.

How about merging this in say a month (or X weeks)?

The uv case is different since it provided by both distro (on arch anyway) and pypi - can lead to inconsistencies that makes it worth it to do the check to run self-update or not.

Yazi is fairly standalone and want to avoid adding soon to be dead code paths. Upgrading also less likely to break things.

If that works, lmk if I should close and reopen.

GideonBear:
> The uv case is different since it provided by both distro (on arch anyway) and pypi - can lead to inconsistencies that makes it worth it to do the check to run self-update or not.

That's a good point; it's a different situation. I think we should have some sort of policy for this.

(BTW pls rebase/merge main to fix the CI failure)

SteveLauC:
> Yazi is fairly standalone and want to avoid adding soon to be dead code paths.

Fair enough. Let's merge the PR and see if we receive complaints from Yazi users. If not, it should be good.

SteveLauC (approved):
Thanks!

PR #1167, opened 2025-06-11
Add support for multiple binary names and idea having multiple binaries

themadprofessor:
## What does this PR do
Add support for tools to have multiple possible binaries because IntellIj IDEA on the AUR can have multiple depending on the edition.

[AUR Ultimate Edition](https://aur.archlinux.org/packages/intellij-idea-ultimate-edition) and [AUR Community Edition](https://aur.archlinux.org/packages/intellij-idea-community-edition-bin) renames the binary to `intellij-idea-ultimate-edition` and `intellij-idea-community-edition` respectively.

## Standards checklist

_Google translated_
 
## For new steps

  the underlying command

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -112,6 +112,29 @@ pub fn require<T: AsRef<OsStr> + Debug>(binary_name: T) -> Result<PathBuf> {
     }
 }
 
+pub fn require_one<T: AsRef<OsStr> + Debug>(binary_names: impl IntoIterator<Item = T>) -> Result<PathBuf> {
+    let mut failed_bins = Vec::new();
+    for bin in binary_names {
+        match require(&bin) {
+            Ok(path) => return Ok(path),
+            Err(_) => failed_bins.push(bin),
+        }
+    }
+
+    Err(SkipStep(format!(
+        "{}",
+        t!(
+            "Cannot find any of {binary_names} in PATH",
+            binary_names = failed_bins
+                .iter()
+                .map(|bin| format!("{:?}", bin))
+                .collect::<Vec<_>>()
+                .join(", ")
+        )
+    ))
+    .into())
+}
+
 #[allow(dead_code)]
 pub fn require_option<T>(option: Option<T>, cause: String) -> Result<T> {
     if let Some(value) = option {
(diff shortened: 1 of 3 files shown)


GideonBear (changes requested):
(I'm the author of the JetBrains IDE plugin steps)
Looks good! Could you check which other IDEs this applies to? For example if there's also `intellij-pycharm-community-edition`, and/or `intellij-rustrover`?

themadprofessor:
I believe that should be all of them. A good chunk of the AUR packages follow the pattern of `ide` and `ide-eap`, but not all of them.

src/utils.rs
```diff
@@ -112,6 +112,29 @@ pub fn require<T: AsRef<OsStr> + Debug>(binary_name: T) -> Result<PathBuf> {
...
+            Ok(path) => return Ok(path),
+            Err(_) => failed_bins.push(bin),
+        }
+    }
+
+    Err(SkipStep(format!(
+        "{}",
+        t!(
+            "Cannot find any of {binary_names} in PATH",
+            binary_names = failed_bins
```
SteveLauC, line 128: When this function fails, `failed_bins` is guaranteed to contain all the `binary_names`, which makes me think the use of `failed_bins` is unnecessary 👀. We can drop `failed_bins` and directly use `binary_names` instead?

SteveLauC (commented):
Regarding the code, only 1 nit:

SteveLauC:
> Add support for tools to have multiple possible binaries because IntellIj IDEA on the AUR can have multiple depending on the edition.


Question: Is the behavior described above general, or is it specific to AUR packages? In other words, have you guys seen this somewhere else?

src/utils.rs
themadprofessor: The issue with that is the function takes `IntoIterator` which takes ownership when converting to an `Iterator` and `Iterator`'s can only be iterated over once. Since each callsite has different numbers of bins to check, the function would need to take a `Vec`, which would mean in the best case (the first bin is found) we allocate, which we don't now since they're arrays.

SteveLauC: That makes sense

themadprofessor:
> > Add support for tools to have multiple possible binaries because IntellIj IDEA on the AUR can have multiple depending on the edition.
> 
> Question: Is the behavior described above general, or is it specific to AUR packages? In other words, have you guys seen this somewhere else?

I've not seen it elsewhere, but there's now a way to handle this sort of thing if it comes up.

SteveLauC (approved):
Thanks!

Will merge it later until I fix the CI stuff:)

Issue #1170, opened 2025-06-13
Where are the rest of the assests for 16.04?


onefish2:
I only see the 2 source code assets. Where are the rest of the builds? This is also causing topgrade to be unable to self update to the new 16.04 version.

10:56:04 - Self update ──────────────────────────────────────────────────────
Checking target-arch... aarch64-unknown-linux-gnu
Checking current version... v16.0.3
Checking latest released version... v16.0.4
New release found! v16.0.3 --> v16.0.4
New release is compatible
Self Update failed:
   0: ReleaseError: No asset found for target: `aarch64-unknown-linux-gnu`

Location:
   src/self_update.rs:29

jwesemann:
Same situation here. Error message on a WSL-Kali distro is : 

> Checking target-arch... x86_64-unknown-linux-gnu
> Checking current version... v16.0.3
> Checking latest released version... v16.0.4
> New release found! v16.0.3 --> v16.0.4
> New release is compatible
> Self Update fehlgeschlagen:
>    0: ReleaseError: No asset found for target: `x86_64-unknown-linux-gnu`

SinTan1729:
AUR releases have failed as well.

GideonBear:
Release CD failed due to the same reason the PR CI failed, the rustfmt. I'll make a PR.
@SteveLauC

onefish2:
Out of curiosity I just ran topgrade on my Mac with brew and it upgraded itself to 16.04.

onefish2:
@GideonBear Any update on when this issue will be resolved?

GideonBear:
I'm reliant on @SteveLauC here unfortunately. In the meantime you can install it from git with cargo.

BriegerCA:
I compiled it from Source for UBUNTU

sudo apt update
sudo apt install git curl build-essential
curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh

source $HOME/.cargo/env

git clone https://github.com/topgrade-rs/topgrade.git
cd topgrade

cargo build --release
**********************************
topgrade will be created in "target/release/topgrade"
Use whereis topgrade to find your current version and make a copy
Example: sudo cp /usr/bin/topgrade /usr/bin/topgrade_save

Then copy the new compile: sudo target/release/topgrade /usr/bin/topgrade

Run topgrade

If all works, you can remove rust/cargo with 
rustup self uninstall

and rm -rf topgrade
[...]

PR #1177, opened 2025-06-16
ci: add event workflow_dispatch to release pipelines

SteveLauC:
## What does this PR do

1. Add the event `workflow_dispatch` event to the release pipelines so that we can run them manually, which would be quite helpful when we encounter issues like https://github.com/topgrade-rs/topgrade/issues/1170
2. Drop the "release to crates.io" pipeline as it never worked (since I maintained this project)
3. Document why we are using the Ubuntu 22.04 image (to have a lower version of glibc)

## Standards checklist

## For new steps

  the underlying command


SteveLauC:
related: https://github.com/topgrade-rs/topgrade/pull/1180

Issue #1049, opened 2025-02-26
Custom step order


TheCodeLamp:
## I want to suggest some general feature
Topgrade should be able to stagger or delay certain upgrade steps to allow for a smoother experience. For example having distrobox be upgraded after chezmoi.

## More information
I have a script in my chezmoi setup that replaces a distrobox container with a new one every month. When topgrade is run on this day it first upgrades the container using the distrobox step, then it replaces it when it gets to the chezmoi step. I know this isn't remotely an important issue, but it would be nice to be able to specify steps that should be run after another, or at least be able to specify "after" steps that are run after everyone else.

It is possible to do something similar with post_commands, however, then you must re-implement the entire step yourself in a string.

Just having a
```toml
# Steps that are run at the end of the upgrade proccess.
after = [ "distrobox", "system" ]
```
would be nice.

sybereal:
In my case, I have the issue that my home-manager configuration is managed via chezmoi, but home-manager runs much earlier than chezmoi. To get the latest config, I have to run topgrade twice or manually invoke `home-manager switch` afterward.

Kladki:
I would like to mention that these usecases mentioned and mine might be better served by a design which makes it explicit that one step depends on some others, rather than mirroring `post_commands`. So for example, if I have a nixos config which I want to be pulled from git before building, the user should set something like:
```toml
[linux]
# or it could be called "depends", or something similar.
after = ["git"]
```

uwuclxdy:
I'd suggest a configuration like this:

```config.toml
first = [ "six", "seven" ]
last = [ "two", "one" ]
```

which would make the order of updates: `six`, `seven`, `three`, `four`, `five`, `two`, `one` 

(the numbers represent update steps if they were executed without an explicit setting)

With this, a user can define the entire update order or just the packages that must run first / last. IMO, this allows for the greatest customizability while still keeping configuration simple.  Did I miss something or is this good enough?

GideonBear:
> I'd suggest a configuration like this:
> 
> first = [ "six", "seven" ]
> last = [ "two", "one" ]
> which would make the order of updates: `six`, `seven`, `three`, `four`, `five`, `two`, `one`
> 
> (the numbers represent update steps if they were executed without an explicit setting)
> 
> With this, a user can define the entire update order or just the packages that must run first / last. IMO, this allows for the greatest customizability while still keeping configuration simple. Did I miss something or is this good enough?

[...]

Issue #877, opened 2024-07-27
parallel update(everything)?


littleblack111:
## I want to suggest a new step

* Which tool is this about? Where is its repository?
* Which operating systems are supported by this tool?
* What should Topgrade do to figure out if the tool needs to be invoked?
* Which exact commands should Topgrade run?
* Does it have a `--dry-run` option? i.e., print what should be done and exit
* Does it need the user to confirm the execution? And does it provide a `--yes`
  option to skip this step?

## I want to suggest some general feature
Topgrade should make it so it can do ALL in parallel. it should spawn sub shells and do them in split screens/multiple screens

## More information

add a option in config or add argument for it to be enabled.

niStee:
Instead of using the GitHub issues page, you could consider posting your suggestion on the Topgrade discussions page at https://github.com/topgrade-rs/topgrade/discussions/categories/ideas. The discussions section is often a better place to share ideas and get feedback from the project maintainers and community.

SteveLauC:
I have thought about this, we can technically spawn all the sub-processes in parallel, but when it comes to user input, all these processes share the same terminal (stdin), it would be totally chaos.

And Topgrade by the original design, is for interactive usage, when a step (tool) fails, it gives you a prompt so that you can take further actions, so I would say this parallel model is not suitable for Topgrade, or it would require a huge refactor that is not worthwhile.

littleblack111:
> I have thought about this, we can technically spawn all the sub-processes in parallel, but when it comes to user input, all these processes share the same terminal (stdin), it would be totally chaos.
> 
> 
> 
> And Topgrade by the original design, is for interactive usage, when a step (tool) fails, it gives you a prompt so that you can take further actions, so I would say this parallel model is not suitable for Topgrade, or it would require a huge refactor that is not worthwhile.

Maybe focus on one window. And use mouse or sth to control it. Kinda like tmux

SteveLauC:
> Maybe focus on one window. And use mouse or sth to control it. Kinda like tmux

You mean scattering the subprocesses across multiple tmux windows, this is indeed something theorically Topgrade can do.

Though I guess the UX will be bad, personally I have more than 10 steps to run in Topgrade, which means I need to repeatedly check the status of those steps to ensure they won't go wrong🤔

hey2022:
[...]

PR #1188, opened 2025-06-17
Move step running into enum for dynamic ordering

themadprofessor:
## What does this PR do
Move the running of steps into the enum itself so the order isn't fixed. First steps towards #1049.

Currently, it follows the same order as main (except pihole) through `step::default_order()`, but its a `Vec` so we can adjust it based on config down the line. The only difference to the order on main is pihole's `System` step has been moved forward to run with the other `System` steps, but it's still after the distribution's `System` step. Not sure if that is considered a "breaking change".

## Standards checklist

## For new steps

  the underlying command


CONTRIBUTING.md
```diff
@@ -16,7 +16,7 @@ In `topgrade`'s term, package manager is called `step`.
 To add a new `step` to `topgrade`:
 
 1. Add a new variant to
-   [`enum Step`](https://github.com/topgrade-rs/topgrade/blob/cb7adc8ced8a77addf2cb051d18bba9f202ab866/src/config.rs#L100)
+   [`enum Step`](https://github.com/topgrade-rs/topgrade/blob/master/src/step.rs)
```
GideonBear:
```suggestion
   [`enum Step`](https://github.com/topgrade-rs/topgrade/blob/main/src/step.rs)
```

GideonBear (approved):
Nice! Also note that apart from #1049 this is also a step towards #877.

CONTRIBUTING.md
themadprofessor: Seems I'm still a bit "old school" 😆

themadprofessor:
Not quite sure why the sorting checks are failing. Pulling the scripts out and running them manually shows they are sorted.

GideonBear:
> Not quite sure why the sorting checks are failing. Pulling the scripts out and running them manually shows they are sorted.

The version of `sort` is different on your machine and the CI; I had this problem as well. It has to do with the capital letters.

GideonBear:
(`GithubCliExtensions` needs to be below `GitRepos`)

themadprofessor:
> (`GithubCliExtensions` needs to be below `GitRepos`)

That makes sense.

src/step.rs
```diff
@@ -8,6 +8,7 @@ use strum::{EnumCount, EnumIter, EnumString, VariantNames};
 
 #[cfg(feature = "self-update")]
 use crate::self_update;
+use crate::step::Step::GitRepos;
```
GideonBear: IDE shenanigans?
```suggestion
```

GideonBear (approved):

src/step.rs
themadprofessor: Rustrover strikes again

SteveLauC:
Looks great at a glance. But it is still a big PR that would take me some time to review:>

SteveLauC:
BTW, I did give #1049 a thought in the past, I was thinking about using trait objects so that custom steps (user-defined steps) will also work. Just my previous thought, not sure if it is feasible

themadprofessor:
> BTW, I did give #1049 a thought in the past, I was thinking about using trait objects so that custom steps (user-defined steps) will also work. Just my previous thought, not sure if it is feasible

Unless there is a plan to allow plugins, not sure there is much point in using trait objects. Since we know all the steps, the added cost of trait objects don't make much sense.

GideonBear:
Yeah I don't think plugins make sense at all. You can always use a custom command.

CONTRIBUTING.md
```diff
@@ -88,17 +88,23 @@ To add a new `step` to `topgrade`:
 
    Congrats, you just added a new `step` :)
```
GideonBear: I assume this was forgotten, let's put this below 4.

src/main.rs
```diff
@@ -6,6 +6,14 @@ use std::path::PathBuf;
...
 use std::time::Duration;
 
+use self::config::{CommandLineArgs, Config};
+use self::error::StepFailed;
+#[cfg(all(windows, feature = "self-update"))]
+use self::error::Upgraded;
+#[allow(clippy::wildcard_imports)]
+use self::steps::{remote::*, *};
+#[allow(clippy::wildcard_imports)]
+use self::terminal::*;
```
GideonBear: Please either move these back, or move `use self::utils...` here as well

GideonBear:
I'm ready to merge once these issues and the merge conflict are resolved.

src/main.rs
themadprofessor: Gonna assume RustRover moved them.

GideonBear:
Thank you, sorry for the spread out review!

GideonBear:
And... 90% of open PR's have conflicts now. Worth it :)

Issue #1190, opened 2025-06-17
PowerShell Module Update step updates user-local modules using sudo, on Unix systems


AThePeanut4:
## Erroneous Behavior
On Unix systems, PowerShell modules installed with `-Scope CurrentUser` (the default for an unprivileged shell) are installed in `~/.local/share/powershell/Modules/`, a user-owned folder. But since v16.0.4, topgrade runs powershell with sudo in the "PowerShell Module Update" step, causing any upgraded files to be owned by `root` rather than the current user. That then means they cannot be removed without also using sudo.

The AllUsers scope module folder (`/usr/local/share/powershell/Modules`) is root-owned, so sudo is then required to update modules there.

Luckily I don't think should actually have affected anyone because of #1174 - the "Module Update" step is currently just broken entirely. But after #1187 is merged, this will become an issue.

## Expected Behavior
The step should probably do separate `Update-Module` for both scopes, with and without sudo:
`sudo pwsh -NoProfile -Command Update-Module -Scope AllUsers`
`pwsh -NoProfile -Command Update-Module -Scope CurrentUser`

## Steps to reproduce
First, build topgrade with #1187 applied.
Then in powershell, install an old version of a module with the CurrentUser scope:
```pwsh
> Install-Module -Name powershell-yaml -Scope CurrentUser -RequiredVersion 0.4.11
```
Then run topgrade:
```sh
$ topgrade --only powershell
```
Then observe that the directory for the updated module now has a root-owned directory in it:
```sh
$ ls -l ~/.local/share/powershell/Modules/powershell-yaml
total 0
drwxr-xr-x  9 user staff 288 Jun 17 21:15 0.4.11
drwxr-xr-x 10 root staff 320 Jun 17 21:15 0.4.12
```

## Did you run topgrade through `Remote Execution`

- [ ] Yes
- [x] No

## Additional Details
- Operation System/Version
  macOS 15.5

## Installation
Build from source

## Topgrade version (`topgrade -V`)
topgrade 16.0.4

atoerien:
I would have just make a fix for this myself, but I'm not sure of the best way to structure it and there seems to be several other powershell-related issues in v16.0.4 as well - I believe all of them resulting from #1041. Ping @niStee - are you able to work out fixes for these? If not I'm happy to give it a go.

[...]

Issue #1174, opened 2025-06-16
PowerShell Module Update Broken on Latest PowerShell


shoddyguard:
## Erroneous Behavior

The PowerShell Module Update step fails.

## Expected Behavior

Command should run without error.

## Steps to reproduce

Simply run topgrade with latest PowerShell 7 installed on Linux.

## Possible Cause (Optional)

I believe the issue stems from PowerShell core (aka `pwsh`) now having `-File` as the default argument when none is provided whereas older version used `-Command`.

As topgrade calls the following: `/usr/bin/sudo /opt/microsoft/powershell/7/pwsh -NoProfile Update-Module` this results in `Update-Module` being ambiguous and therefore being assigned to the default `-File` parameter which causes PowerShell to look for a file named `Update-Module` rather than running the command.
Indeed if I run this command manually replicating what topgrade calls PowerShell errors in the same way.

However changing the command to `/usr/bin/sudo /opt/microsoft/powershell/7/pwsh -NoProfile -Command Update-Module` results in a success.

So I believe updating the call topgrade uses to include the `-Command` parameter ahead of `Update-Module` should be sufficient to fix this.

Happy to submit a PR if folks are open to this suggested fix.

## Problem persists without calling from topgrade

- [x] Yes
- [ ] No

## Did you run topgrade through `Remote Execution`

- [ ] Yes
- [x] No

## Configuration file (Optional)

N/A

## Additional Details

- Operation System/Version: Fedora Workstation 42
- Installation: Cargo
- Topgrade version (`topgrade -V`): 16.0.4
- PowerShell Version 7.6.0-preview4

## Verbose Output (`topgrade -v`)

<details>
<pre>
── 01:35:03 - Powershell Modules Update ────────────────────────────────────────
Updating modules...
The argument 'Update-Module' is not recognized as the name of a script file. Check the spelling of the name, or if a path was included, verify that the path is correct and try again.

Usage: pwsh[.exe] [-Login] [[-File] <filePath> [args]]
                  [-Command { - | <script-block> [-args <arg-array>]
                                | <string> [<CommandParameters>] } ]
                  [-CommandWithArgs <string> [<CommandParameters>]
                  [-ConfigurationName <string>] [-ConfigurationFile <filePath>]
                  [-CustomPipeName <string>] [-EncodedCommand <Base64EncodedCommand>]
                  [-ExecutionPolicy <ExecutionPolicy>] [-InputFormat {Text | XML}]
[...]

PR #1192, opened 2025-06-18
fix/pwsh

ao-xing:
## What does this PR do
fixes https://github.com/topgrade-rs/topgrade/issues/1174
fixes https://github.com/topgrade-rs/topgrade/issues/1190

## Standards checklist

## For new steps

  the underlying command

If you developed a feature or a bug fix for someone else and you do not have the means to test it, please tag this person here.

@GideonBear 
@AThePeanut4 
https://github.com/shoddyguard

### DIFF
--- a/src/steps/powershell.rs
+++ b/src/steps/powershell.rs
@@ -22,6 +22,10 @@ impl Powershell {
         Powershell { path, profile }
     }
 
+    pub fn is_available(&self) -> bool {
+        self.path.is_some()
+    }
+
     #[cfg(windows)]
     pub fn windows_powershell() -> Self {
         Powershell {
@@ -79,7 +83,7 @@ impl Powershell {
     }
 
     pub fn update_modules(&self, ctx: &ExecutionContext) -> Result<()> {
-        print_separator("Powershell Modules Update");
+        print_separator(t!("Powershell Modules Update"));
 
         let mut cmd_args = vec!["-Command", "Update-Module"];
 
@@ -95,24 +99,15 @@ impl Powershell {
 
         #[cfg(not(windows))]
         {
-            // Update user-scope modules WITHOUT sudo (to avoid root ownership issues)
-            println!("Updating user-scope modules...");
-            let mut user_cmd_args = cmd_args.clone();
-            user_cmd_args.extend_from_slice(&["-Scope", "CurrentUser"]);
-            self.build_command_internal(ctx, &user_cmd_args, false)?
-                .status_checked()?;
-
-            // Update system-scope modules WITH sudo (if available)
-            println!("Updating system-scope modules...");
-            let mut system_cmd_args = cmd_args.clone();
-            system_cmd_args.extend_from_slice(&["-Scope", "AllUsers"]);
-            self.build_command_internal(ctx, &system_cmd_args, true)?
-                .status_checked()?;
+            // On Unix, run Update-Module without sudo since PowerShell Core defaults to CurrentUser scope
+            // and Update-Module updates all modules regardless of their original installation scope
+            self.build_command_internal(ctx, &cmd_args, false)?.status_checked()?;
         }
 
         #[cfg(windows)]
         {
-            // On Windows, run Update-Module once with sudo if available
+            // On Windows, use sudo if available since Windows PowerShell 5.1 defaults to AllUsers scope
+            // and may need administrator privileges
             self.build_command_internal(ctx, &cmd_args, true)?.status_checked()?;
         }
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -136,7 +136,7 @@ fn run() -> Result<()> {
     }
 
     let powershell = powershell::Powershell::new();
-    let should_run_powershell = powershell.profile().is_some() && config.should_run(Step::Powershell);
+    let should_run_powershell = powershell.is_available() && config.should_run(Step::Powershell);
     let emacs = emacs::Emacs::new();
     #[cfg(target_os = "linux")]
     let distribution = linux::Distribution::detect();


GideonBear:
Small tip, if this closes (fixes) both #1174 and #1190, put the keyword (`fixes` or `closes`) in front of both of them separately. Right now this PR is only marked to fix #1174.
Example:
```
Fixes #1174
Fixes #1190
```

src/steps/powershell.rs
```diff
@@ -77,18 +77,51 @@ impl Powershell {
...
+    fn build_command_without_sudo<'a>(
+        &self,
+        ctx: &'a ExecutionContext,
+        additional_args: &[&str],
+    ) -> Result<impl CommandExt + 'a> {
+        let powershell = require_option(self.path.as_ref(), t!("Powershell is not installed").to_string())?;
+        let executor = &mut ctx.run_type();
+
+        // Always execute directly without sudo on Unix systems
+        let mut command = executor.execute(powershell);
```
GideonBear: This is usually inline
```suggestion
        // Always execute directly without sudo on Unix systems
        let mut command = ctx.run_type().execute(powershell);
```

src/steps/powershell.rs
```diff
@@ -77,18 +77,51 @@ impl Powershell {
...
+        // Update user-scope modules WITHOUT sudo (to avoid root ownership issues)
+        println!("Updating user-scope modules...");
+        let user_result = self.build_command_without_sudo(ctx, &["-Command", "Update-Module", "-Scope", "CurrentUser"])?
+            .status_checked();
+        
+        match user_result {
+            Ok(_) => println!("✓ User-scope modules updated successfully"),
+            Err(e) => {
+                println!("⚠ Failed to update user-scope modules: {}", e);
+            }
```
GideonBear: This should fail the step
```suggestion
        self.build_command_without_sudo(ctx, &["-Command", "Update-Module", "-Scope", "CurrentUser"])?
            .status_checked()?;
```

src/steps/powershell.rs
```diff
@@ -77,18 +77,51 @@ impl Powershell {
...
+        // Update system-scope modules WITH sudo (if available)
+        println!("Updating system-scope modules...");
+        let system_result = self.build_command_internal(ctx, &["-Command", "Update-Module", "-Scope", "AllUsers"])?
+            .status_checked();
+        
+        match system_result {
+            Ok(_) => println!("✓ System-scope modules updated successfully"),
+            Err(e) => {
+                println!("⚠ Failed to update system-scope modules: {}", e);
+            }
```
GideonBear: This should fail the step
```suggestion
        self.build_command_internal(ctx, &["-Command", "Update-Module", "-Scope", "AllUsers"])?
            .status_checked()?;
```

src/steps/powershell.rs
```diff
@@ -77,18 +77,51 @@ impl Powershell {
...
+            .status_checked();
+        
+        match user_result {
+            Ok(_) => println!("✓ User-scope modules updated successfully"),
+            Err(e) => {
+                println!("⚠ Failed to update user-scope modules: {}", e);
+            }
         }
-        if ctx.config().yes(Step::Powershell) {
-            cmd_args.push("-Force");
```
GideonBear, line 88: You're completely missing the `-Force` option.

src/steps/powershell.rs
```diff
@@ -77,18 +77,51 @@ impl Powershell {
...
+        let executor = &mut ctx.run_type();
+
+        // Always execute directly without sudo on Unix systems
+        let mut command = executor.execute(powershell);
+        command.args(Self::common_args()).args(additional_args);
+        Ok(command)
+    }
 
-        if ctx.config().verbose() {
-            cmd_args.push("-Verbose");
```
GideonBear, line 85: You're completely missing the `-Verbose` option.

GideonBear (changes requested):

src/steps/powershell.rs
```diff
@@ -77,18 +77,55 @@ impl Powershell {
...
+        let mut command = executor.execute(powershell);
+        command.args(Self::common_args()).args(additional_args);
+        Ok(command)
+    }
+
+    #[cfg(not(windows))]
     pub fn update_modules(&self, ctx: &ExecutionContext) -> Result<()> {
-        print_separator(t!("Powershell Modules Update"));
-        let mut cmd_args = vec!["Update-Module"];
+        print_separator("Powershell Modules Update");
```
atoerien: Removed the `t!` translation macro, the new messages should probably also get translations as well

niStee:
Please test in on MacOs and Linux. I just tested it on Windows 11:
```
PS C:\git\topgrade\target\debug> .\topgrade.exe --only powershell       

── 20:15:51 - Powershell Modules Update ────────────────────────────────────────

── 20:15:54 - Summary ──────────────────────────────────────────────────────────
Powershell Modules Update: OK
```

atoerien:
Can confirm it is running the two commands correctly on both macOS and Linux.

However I was just doing some more testing and it turns out that `Update-Module` doesn't exactly work as I thought initially: no matter what `-Scope` value you use, it will always install updates for all installed modules. The `-Scope` just determines where to put the new modules. So if you install a module to `/usr/local/share/powershell/Modules/` with `Install-Module -Scope AllUsers`, and then update it with just `Update-Module` (`-Scope CurrentUser` is the default), then the updated version of that module will be in `~/.local/share/powershell/Modules/` instead.

In other words, in the topgrade step, after `pwsh -NoProfile -Command Update-Module -Scope CurrentUser` is run, **all** modules will have been updated, including those installed to `/usr/local/share/powershell/Modules/`, with the new files stored in `~/.local/share/powershell/Modules/`. And that then means that the `sudo pwsh -NoProfile -Command Update-Module -Scope AllUsers` command will never actually do anything, because all modules have already been updated.

I can't find a way to tell `Update-Module` to only updated packages installed with a specific scope, so I think it would make the most sense for topgrade to only run `pwsh -NoProfile -Command Update-Module`, which is almost the same as before, only now it's **not** using sudo. The `-Scope CurrentUser` isn't needed since that's the default for PowerShellGet v2.0.0 and up, and PowerShellCore (pwsh) comes with v2.2.5. On Windows, it's probably better to keep using sudo/gsudo/whatever if possible, because the default Windows PowerShell 5.1 has PowerShellGet v1.0.0.1 which defaults to `-Scope AllUsers` and would therefore need administrator privileges.

src/steps/powershell.rs
SteveLauC: This should be revered, we don't want to break the i18n support

src/steps/powershell.rs
```diff
@@ -78,17 +79,44 @@ impl Powershell {
...
         if ctx.config().verbose() {
             cmd_args.push("-Verbose");
         }
-        if ctx.config().yes(Step::Powershell) {
-            cmd_args.push("-Force");
+
+        #[cfg(not(windows))]
+        {
+            // Update user-scope modules WITHOUT sudo (to avoid root ownership issues)
+            println!("Updating user-scope modules...");
```
SteveLauC: This and `println!("Updating system-scope modules...");` are user-facing messages, which need translations

SteveLauC (commented):
I know nothing about powershell, thanks to everyone who reviewed this PR.

SteveLauC:
> Can confirm it is running the two commands correctly on both macOS and Linux.
> 
> However I was just doing some more testing and it turns out that `Update-Module` doesn't exactly work as I thought initially: no matter what `-Scope` value you use, it will always install updates for all installed modules. The `-Scope` just determines where to put the new modules. So if you install a module to `/usr/local/share/powershell/Modules/` with `Install-Module -Scope AllUsers`, and then update it with just `Update-Module` (`-Scope CurrentUser` is the default), then the updated version of that module will be in `~/.local/share/powershell/Modules/` instead.
> 
> In other words, in the topgrade step, after `pwsh -NoProfile -Command Update-Module -Scope CurrentUser` is run, **all** modules will have been updated, including those installed to `/usr/local/share/powershell/Modules/`, with the new files stored in `~/.local/share/powershell/Modules/`. And that then means that the `sudo pwsh -NoProfile -Command Update-Module -Scope AllUsers` command will never actually do anything, because all modules have already been updated.
> 
> I can't find a way to tell `Update-Module` to only updated packages installed with a specific scope, so I think it would make the most sense for topgrade to only run `pwsh -NoProfile -Command Update-Module`, which is almost the same as before, only now it's **not** using sudo. The `-Scope CurrentUser` isn't needed since that's the default for PowerShellGet v2.0.0 and up, and PowerShellCore (pwsh) comes with v2.2.5. On Windows, it's probably better to keep using sudo/gsudo/whatever if possible, because the default Windows PowerShell 5.1 has PowerShellGet v1.0.0.1 which defaults to `-Scope AllUsers` and would therefore need administrator privileges.

Since I have no experience with pwsh, I will adopt the solution you guys choose:)

niStee:
> Can confirm it is running the two commands correctly on both macOS and Linux.
> 
> However I was just doing some more testing and it turns out that `Update-Module` doesn't exactly work as I thought initially: no matter what `-Scope` value you use, it will always install updates for all installed modules. The `-Scope` just determines where to put the new modules. So if you install a module to `/usr/local/share/powershell/Modules/` with `Install-Module -Scope AllUsers`, and then update it with just `Update-Module` (`-Scope CurrentUser` is the default), then the updated version of that module will be in `~/.local/share/powershell/Modules/` instead.
> 
> In other words, in the topgrade step, after `pwsh -NoProfile -Command Update-Module -Scope CurrentUser` is run, **all** modules will have been updated, including those installed to `/usr/local/share/powershell/Modules/`, with the new files stored in `~/.local/share/powershell/Modules/`. And that then means that the `sudo pwsh -NoProfile -Command Update-Module -Scope AllUsers` command will never actually do anything, because all modules have already been updated.
> 
> I can't find a way to tell `Update-Module` to only updated packages installed with a specific scope, so I think it would make the most sense for topgrade to only run `pwsh -NoProfile -Command Update-Module`, which is almost the same as before, only now it's **not** using sudo. The `-Scope CurrentUser` isn't needed since that's the default for PowerShellGet v2.0.0 and up, and PowerShellCore (pwsh) comes with v2.2.5. On Windows, it's probably better to keep using sudo/gsudo/whatever if possible, because the default Windows PowerShell 5.1 has PowerShellGet v1.0.0.1 which defaults to `-Scope AllUsers` and would therefore need administrator privileges.

@AThePeanut4  Please test it again, especially if it now runs as expected on MacOS.

AThePeanut4 (approved):
LGTM, tested working on both macOS and Linux.

SteveLauC:
Hi @ao-xing, why did you close this PR? I didn't review it does not mean I do not want it, it's simply because I do not have too much time for open source

ao-xing:
Thanks for asking. It was merged into https://github.com/topgrade-rs/topgrade/pull/1197 so i closed it.

PR #1197, opened 2025-06-23
Fix various powershell issues

AThePeanut4:
## What does this PR do

This is a combination of previous PRs #1187, #1192 and #1196. While testing them, I had to make some tweaks to fix issues I found, and then couldn't help myself from doing some cleanup and refactoring in `powershell.rs`. I've also added two minor additional fixes.

@SteveLauC if you want to rather merge the individual PRs separately then I don't mind, I just think it's easier to understand and review when it's all in one place. I've added the change from each previous PR as a separate commit, with `Co-authored-by` tags so it shows the relevant people as co-authors - the idea would be to fast-forward merge the PR rather than squash it.

@ao-xing and @niStee, I hope it's fine that I've merged both your changes into one PR here. I have tested this myself on both Windows, Linux and macOS, but if you could also do a quick test that would be much appreciated.

Going through each commit:

- 574bda2a47857fc06391de518e35f50d621cf66f: this is #1187.
- b75a2011dd529512e947c1c91f669324dcacb08a: this is the cleanup and refactoring I did on `powershell.rs`. Basically what I did was replace `execute_with_command` and `build_command_internal` with `build_command_internal` and `build_command`, that take a single command string instead of a list of arguments. Then all the uses of `Command::new` in the file can be replaced with either `build_command_internal` or `build_command`. I also removed the unnecessary `windows` sub-module and inlined the code into the `Powershell` impl just above, which is windows-only anyway.
- 889c8fb27b5cacc6d53f3b3100d1386690d5919b: this is #1192. What I changed was that I added logic to remember whether we were using `powershell` or `pwsh`, and then only use `sudo` if running `powershell`. The reason for this is because even though Windows's file permission inheritance means that using `sudo` doesn't cause the modules in `C:\Users\user\Documents\PowerShell\Modules` to become un-deletable like it does on macOS, the owner is still set to "Administrators" rather than your user account. Only in normal Windows PowerShell is sudo needed, because that uses `-Scope AllUsers` and installs in `C:\Program Files\WindowsPowerShell\Modules`.
[...]

Issue #1114, opened 2025-04-10
Warn when running Topgrade as superuser on non-Windows systems


GideonBear:
Ref #1106

## I want to suggest some general feature
Topgrade should warn when running as superuser on non-Windows systems.
There should also be a configuration option to disable this warning.
### Why not on Windows?
On Windows, the default `sudo` experience is pretty bad, as it is uncached without installing `gsudo` or similar. Therefore it is acceptable to run Topgrade as superuser (Administrator).

SteveLauC:
Maybe this warning should be enabled on all the UNIX systems?

GideonBear:
> Maybe this warning should be enabled on all the UNIX systems?

Oh yes, all non-Windows systems is probably better

Issue #598, opened 2023-11-03
An option/config to echo dry-run commands in wet runs


SteveLauC:
### Discussed in https://github.com/topgrade-rs/topgrade/discussions/597

> Originally posted by @jiangyilism on November  2, 2023

Currently if we want to check/view what commands topgrade executes we run it with `--dry-run` .
Please consider adding an option/config to print each command like `--dry-run` while doing wet runs.

`--verbose` works but it prints too much. It would be great if we have `--verbose` be less verbose than `--log-filter=debug`. In my opinion `--verbose` is for users while `debug` is more for developers. But that is another story...

If it is not worth adding a new command, then maybe add this behavior to `--log-filter=info`. But personally, I prefer another new log level (e.g. `--log-filter=echo`) since `info` is still too verbose for me.

SteveLauC:
would you like to tell me why would you want this in wet run if we can already get this info from a dry run?

--------


> Please consider adding an option/config to print each command like --dry-run while doing wet runs.

Could you please show me what it should look like? Something like this?

```sh
$ topgrade

Topgrade is going to update the following tools:

1. xx
2. xxx

Updateing:
----------

── 16:17:08 - XXXXXXXXXXXX ─────────────────────────

xxx

── 16:17:16 - XXXXXXXXXXXX ─────────────────────────

xxx
```

> If it is not worth adding a new command, then maybe add this behavior to --log-filter=info. But personally, I prefer another new log level (e.g. --log-filter=echo) since info is still too verbose for me.

We can not add new log levels as they are defined by the [upstream crate](https://docs.rs/log/0.4.20/log/enum.LevelFilter.html#method.to_level) unless we drop it and implement our log crate, which is not what I want

jiangyilism:
For example, this is a part of my `topgrade` output:

```
── 17:33:59 - rustup ───────────────────────────────────────────────────────────
info: syncing channel updates for 'stable-x86_64-unknown-linux-gnu'
info: latest update on 2023-11-16, rust version 1.74.0 (79e9716c9 2023-11-13)
info: downloading component 'cargo'
info: downloading component 'clippy'
...

```
and this is the corresponding `upgrade --dry-run` output:

```
── 17:30:52 - rustup ───────────────────────────────────────────────────────────
Dry running: /home/grape/.local/share/cargo/bin/rustup update
```

What I wish for is something like (the only diff is the 2nd line):

```
── 17:33:59 - rustup ───────────────────────────────────────────────────────────
[...]

Issue #530, opened 2023-07-30
Topgrade silently skips `apt` update if `sudo` is not installed


KizzyCode:
## Erroneous Behavior
`topgrade` silently skips the `apt-get` system update step on Debian Bookworm if `sudo` is not installed; even if `topgrade` is run as `root`.

## Expected Behavior
If the current user is `root`, `topgrade` should run `apt-get` (or other commands) without `sudo` to perform the update. Alternatively, `topgrade` should raise a warning or error instead of silently skipping this step.

## Steps to reproduce
- Install a minimal Debian Bookworm without `sudo` (i.e. from the netinstall image)
- Install `rust` via rustup
- Install `topgrade` via `cargo`

## Problem persists without calling from topgrade
- [x] Yes
- [x] No

`Yes` if I copy the **exact** command with `sudo` (since `sudo` is not installed), `No` if I just execute `apt-get update`.

## Did you run topgrade through `Remote Execution`
- [x] Yes
- [x] No

If yes, does the issue still occur when you run topgrade directlly in your
remote host

- [x] Yes
- [ ] No

## Additional Details
- Operation System/Version: Debian Bookworm

- Installation: `topgrade` via `cargo` via `rustup`

- Topgrade version (`topgrade -V`): `Topgrade 12.0.2`

## Verbose Output (`topgrade -v`)

<details>
<pre>
DEBUG Configuration at /root/.config/topgrade.toml
DEBUG Loaded configuration: ConfigFile { include: None, misc: Some(Misc { pre_sudo: None, sudo_command: None, git_repos: None, predefined_git_repos: None, disable: None, ignore_failures: None, remote_topgrades: None, remote_topgrade_path: None, ssh_arguments: None, git_arguments: None, tmux_arguments: None, set_title: None, display_time: None, assume_yes: None, yay_arguments: None, aura_aur_arguments: None, aura_pacman_arguments: None, no_retry: None, run_in_tmux: None, cleanup: None, notify_each_step: None, accept_all_windows_updates: None, skip_notify: None, bashit_branch: None, only: None, no_self_update: None }), pre_commands: None, post_commands: None, commands: Some({"restic": "restic self-update"}), python: None, composer: None, brew: None, linux: None, git: None, windows: None, npm: None, yarn: None, vim: None, firmware: None, vagrant: None, flatpak: None, distrobox: None }
DEBUG Version: 12.0.2
DEBUG OS: x86_64-unknown-linux-gnu
[...]

PR #1208, opened 2025-06-26
Sudo fixes and related refactoring

AThePeanut4:
## What does this PR do

This started out as a fix for an issue I was getting on Windows after updating to v16.0.4, where the PowerShell step fails because it was using `sudo` (I have gsudo) and the full path to `pwsh.exe` on my system contains a space. After a while I found the cause - https://github.com/gerardog/gsudo/issues/297 - with a solution being to use the `-d` argument. But even without that bug, it still makes sense to use `-d` because it prevents gsudo from running the command in a shell, which is in our case unnecessary (it does this by default so you can sudo internal shell commands).

I could have just checked for gsudo and added the `-d` argument in the powershell step, but then the problem could come up again if some other step needs sudo and has a binary with a space in the name. So what I did was change all the steps to go through `execute_elevated` rather than calling sudo directly. But while doing that I realised that several steps already pass arguments to sudo, and they all do it without checking the sudo kind, meaning that if the sudo isn't standard `sudo`, the steps would fail.

To deal with this, I ended up adding "generic" sudo options, that can be specified if a step needs them. These can then be translated into the correct arguments for the specific sudo kind in use, or error thrown if the option is unsupported. A potential improvement here could be to try emulate some options if they are unsupported, by e.g. prepending the command with `env`, at least on Unix platforms. On Windows this would probably be too difficult to do elegantly, and probably wouldn't be that useful anyway.

I discovered while testing that the `pre_sudo` option doesn't work when topgrade is run in powershell and with gsudo, because it uses `echo` which is aliased to `Write-Output` and that blocks waiting for input if no arguments are given. A better command to use is `rem` in cmd, which is a "comment" command and does nothing.

I also discovered that `sudo_command = "sudo"` in the config doesn't work to force the use of Windows Sudo if gsudo is installed and masquerading as sudo. The easiest fix for this is to just hardcode the path for sudo on Windows, which is fine imo because it ships with the OS so should always be at the same path. The alternative would be to use `which_all` and go through each match, and checking whether it's actually gsudo. But that has the disadvantage that some other sudo implementation masquerading as `sudo.exe` would be treated as if it was Windows Sudo.

Then the rest of the PR is basically just refactoring again, I'm sorry 😭 

- 12c12caf05bc11a521607083174e2968006d5135: I noticed that 99% of the uses for `ctx.run_type()` was to just do `ctx.run_type().execute()`, and that seems a bit silly to me. Logically I think it makes just as much sense to have `execute()` directly on `ExecutionContext`, so I moved everything over.
- 4ef6ccf411ef4dfcfcf78c2c1811b42825e1e5ab: After that, `RunType` becomes just a glorified boolean, so I replaced it with just a `dry_run` bool.
- 60a3210d467a9f7bbb1339d81ab156a460186100: This just simplifies the apt frontend detection code in `upgrade_debian()`, and extracts it out so it can be reused in `should_skip_needrestart()`.
- 99c01cf7523308983562f4a61a88ff885ce698b9: This removes the unnecessary `Yarn::yarn` field - `Yarn::new` is only ever called once with the result of `require("yarn")`, so `command` and `yarn` will always be the same.

## Standards checklist

- [x] The PR title is descriptive (well idk, but not sure what else to put there)


GideonBear:
image
Love to see it! Will review soon :)

src/error.rs
```diff
@@ -68,6 +70,26 @@ impl Display for StepFailed {
     }
 }
 
+#[derive(Error, Debug)]
+pub struct UnsupportedSudo<'a> {
+    pub sudo_kind: SudoKind,
+    pub flag: &'a str,
+}
```
GideonBear: I don't like that this exists. What is the user supposed to do with this?
I don't have a much better alternative for you, but I don't know if this is a good idea.

src/steps/node.rs
```diff
@@ -93,14 +92,10 @@ impl NPM {
...
         let args = ["update", self.global_location_arg()];
         if use_sudo {
-            let sudo = require_option(ctx.sudo().clone(), get_require_sudo_string())?;
-            ctx.run_type()
-                .execute(sudo)
-                .arg(&self.command)
-                .args(args)
-                .status_checked()?;
+            let sudo = ctx.require_sudo()?;
+            sudo.execute(ctx, &self.command)?.args(args).status_checked()?;
```
GideonBear, line 96: Can't this be a method on `ctx`, like `execute_elevated` was?
```rust
fn execute_elevated(&self, command: &str) -> Result<Command> {
    self.require_sudo()?.execute(self, command)
}
```
```diff
-let sudo = ctx.require_sudo()?;
-sudo.execute(ctx, &self.command)?.args(args).status_checked()?;
+ctx.execute_elevated(&self.command)?.args(args).status_checked()?;
```

src/steps/os/archlinux.rs
```diff
@@ -294,39 +285,34 @@ impl ArchPackageManager for Aura {
...
             cmd.arg("-Syu")
                 .args(ctx.config().aura_pacman_arguments().split_whitespace());
             if ctx.config().yes(Step::System) {
                 cmd.arg("--noconfirm");
             }
             cmd.status_checked()?;
         } else {
-            let sudo = crate::utils::require_option(
-                ctx.sudo().as_ref(),
-                t!("Aura(<0.4.6) requires sudo installed to work with AUR packages").to_string(),
```
GideonBear, line 316: Is this removed in the locale file as well?

src/steps/os/dragonfly.rs
```diff
@@ -1,36 +1,36 @@
...
-    {
-        println!(t!(
-            "The package audit was successful, but vulnerable packages still remain on the system"
-        ));
-    }
-    Ok(())
+    sudo.execute(ctx, "/usr/local/sbin/pkg")?
+        .args(["audit", "-Fr"])
+        .status_checked_with(|status| {
+            if !status.success() {
```
GideonBear, line 28: Not your code, but can we check for a specific error code here? There should be a single one meaning what we expect, and others for incorrect arguments, SIGINT, etc.

src/steps/os/openbsd.rs
```diff
@@ -49,18 +46,13 @@ pub fn upgrade_packages(ctx: &ExecutionContext) -> Result<()> {
...
-        ctx.run_type()
-            .execute(sudo)
-            .args(["/usr/sbin/pkg_delete", "-ac"])
-            .status_checked()?;
+        sudo.execute(ctx, "/usr/sbin/pkg_delete")?.arg("-ac").status_checked()?;
     }
 
-    let mut args = vec!["/usr/sbin/pkg_add", "-u"];
+    let mut command = sudo.execute(ctx, "/usr/sbin/pkg_add")?;
+    command.arg("-v");
```
GideonBear: `-u` became `-v` here. Accidental?

src/steps/os/windows.rs
```diff
@@ -20,12 +20,8 @@ pub fn run_chocolatey(ctx: &ExecutionContext) -> Result<()> {
...
     let mut command = match ctx.sudo() {
-        Some(sudo) => {
-            let mut command = ctx.run_type().execute(sudo);
-            command.arg(choco);
-            command
-        }
-        None => ctx.run_type().execute(choco),
+        Some(sudo) => sudo.execute(ctx, &choco)?,
+        None => ctx.execute(choco),
     };
```
GideonBear, line 25: This seems like a bad idea. Why change the behavior just because sudo is or isn't available?
(Again not your code, sorry)

src/sudo.rs
```diff
@@ -21,36 +22,56 @@ pub struct Sudo {
...
     pub fn new(kind: SudoKind) -> Option<Self> {
-        which(kind.as_ref()).map(|path| Self { path, kind })
+        kind.which().map(|path| Self { path, kind })
+    }
+
+    /// Gets the path to the `sudo` binary. Do not use this to execute `sudo` directly - either use
+    /// [`Sudo::elevate`], or if you need to specify arguments to `sudo`, use [`Sudo::elevate_opts`].
+    /// This way, sudo options can be specified generically and the actual arguments customized
+    /// depending on the sudo kind.
+    #[allow(unused)]
```
GideonBear, line 124: Why have this in the first place if it's disallowed *and* unused?
If we want to keep this, can you add this as a disallowed method for clippy?

src/sudo.rs
```diff
@@ -21,36 +22,56 @@ pub struct Sudo {
...
+    /// Set the HOME environment variable to the target user's home directory.
+    pub set_home: bool,
+    /// Run the command as a user other than the root user.
+    pub user: Option<&'a str>,
+}
 
+#[cfg(not(target_os = "windows"))]
+const DETECT_ORDER: [SudoKind; 5] = [
+    SudoKind::Doas,
+    SudoKind::Sudo,
```
GideonBear, line 95: I would prefer splitting this into `WindowsSudo` and `UnixSudo` (behind `cfg(target_os)`), to make sure we don't accidentally forget the `target_os` check somewhere.

src/sudo.rs
```diff
@@ -114,24 +142,132 @@ impl Sudo {
...
+                }
+                SudoKind::Gsudo => {
+                    // By default, gsudo runs all commands inside a shell, so it's effectively
+                    // always "interactive". If interactive is *not* specified, we add `-d`
+                    // to run outside of a shell - see below.
+                }
+                _ => {
+                    return Err(UnsupportedSudo {
+                        sudo_kind: self.kind,
+                        flag: "-i",
```
GideonBear: Why is this `-i` and not `interactive`? IMO better to put the generic name in the message. (here and elsewhere.)

src/sudo.rs
```diff
@@ -21,36 +22,56 @@ pub struct Sudo {
...
-    }
+#[derive(Clone, Debug, Default)]
+/// Generic sudo options, translated into flags to pass to `sudo`. Depending on the sudo kind, OS
+/// and system config, some options might be specified by default or unsupported.
+pub struct SudoExecuteOpts<'a> {
+    /// Run the command "interactively", i.e. inside a login shell.
+    pub interactive: bool,
+    /// Preserve environment variables across the sudo call. If an empty list is given, preserves
+    /// all existing environment variables.
+    pub preserve_env: Option<&'a [&'a str]>,
```
GideonBear: Empty as "all" seems kinda error-prone. `enum` of `List(NonEmpty<String>)` and `All`?

src/sudo.rs
```diff
@@ -114,24 +142,132 @@ impl Sudo {
...
+    pub fn execute_opts<S: AsRef<OsStr>>(
+        &self,
+        ctx: &ExecutionContext,
+        command: S,
+        opts: SudoExecuteOpts,
+    ) -> Result<Executor> {
+        let mut cmd = ctx.execute(&self.path);
+
+        if opts.interactive {
+            match self.kind {
```
GideonBear, line 211: Can we make these `match`es exhaustive and drop the `_ => `?

src/steps/os/unix.rs
```diff
@@ -316,26 +323,27 @@ pub fn run_brew_formula(ctx: &ExecutionContext, variant: BrewVariant) -> Result<
...
-                .status_checked()?;
+            let sudo = ctx.require_sudo()?;
+            sudo.execute_opts(
+                ctx,
+                &binary_name,
+                SudoExecuteOpts {
+                    set_home: true,
+                    user: Some(&user.name),
+                    ..Default::default()
+                },
```
GideonBear: Is this a normal way to do this?
Having to specify `Some` and `..Default::default()` is a bit unwieldy IMO, I would suggest two alternatives:
1. Builder pattern on `SudoExecuteOpts`, pass that into `execute_opts`
```rust
sudo.execute_opts(
    ctx,
    &binary_name,
    SudoExecuteOpts::new()
        .set_home()
        .user(&user.name),
)?
    .current_dir("/tmp")
    .arg("update")
    .status_checked()?;
```
2. Remove `execute_opts`, do builder on `Sudo` instead:
```rust
sudo.clone()
    .set_home()
    .user(&user.name)
    .execute(ctx, &binary_name)?
    .current_dir("/tmp")
    .arg("update")
    .status_checked()?;
```
Then store the opts in `Sudo` itself. The clone is a bit suboptimal, but the resulting code is very clean. We can also make `require_sudo` return `Sudo` instead of `&Sudo`, however this is not necessary if you don't use the opts.

src/steps/os/unix.rs
```diff
@@ -74,19 +74,36 @@ impl BrewVariant {
...
+                command
+            }
+            BrewVariant::MacArm if cfg!(target_arch = "x86_64") => {
+                let mut command = Command::new("arch");
+                command.arg("-arm64e").arg(self.binary_name());
+                command
+            }
+            _ => Command::new(self.binary_name()),
+        }
+    }
```
GideonBear: I don't get this, what is this for? Can you comment it?

src/steps/os/unix.rs
```diff
@@ -547,16 +550,20 @@ pub fn run_nix_self_upgrade(ctx: &ExecutionContext) -> Result<()> {
...
-        ctx.execute_elevated(&nix, true)?
-            .args(nix_args)
-            .arg("upgrade-nix")
-            .status_checked()
+        let sudo = ctx.require_sudo()?;
+        sudo.execute_opts(
+            ctx,
+            &nix,
+            SudoExecuteOpts {
+                interactive: true,
```
GideonBear: Why did you add this option here? Can you comment it?

src/steps/os/linux.rs
```diff
@@ -509,49 +504,51 @@ fn upgrade_gentoo(ctx: &ExecutionContext) -> Result<()> {
...
+
+    if let Some(apt_fast) = which("apt-fast") {
+        Ok((AptFast, apt_fast))
+    } else if let Some(mist) = which("mist") {
+        Ok((Mist, mist))
+    } else if Path::new("/usr/bin/nala").exists() {
+        Ok((Nala, Path::new("/usr/bin/nala").to_path_buf()))
+    } else {
+        Ok((AptGet, require("apt-get")?))
+    }
```
GideonBear, line 526: No priority list here?

src/sudo.rs
GideonBear: (e.g. if someone populates the list programmatically, and some code paths lead to an empty list)

src/error.rs
atoerien: Yeah I also wasn't sure about it. Initially I had it just error with `SkipStep`, but by default that doesn't show any message at all. I thought that was bad because the error would generally happen after the separator had been printed which would mean the step would print nothing and so appear to succeed, when in fact it failed.

But yes there's not really much for a user to do with it, besides install a different sudo and run topgrade again.

In effect it's doing the same as before, where if you were using e.g. `doas` then `doas -i` would give a "command failed" error, only now with a different error message.

src/steps/node.rs
atoerien: It could certainly be, but I think it's nice to separate out "requiring" from "executing", and having such a method encourages combining them. What I mean is most non-sudo steps (should) look like this:
```rust
let bin = require("binary")?;

print_separator(t!("Separator"));

let mut cmd = ctx.execute(bin)?;
...
```
so that the separator is only printed if all the `require`s succeeded. And finding sudo is basically another "require" which normally goes above the separator.

There is an argument to be made for having the step properly error out if sudo is missing (related: #370 and #530), in which case the `require_sudo` would go after the separator and would return something other than `SkipStep`. Then a combined method like `execute_elevated` would be useful, although I still think it's logically nicer to keep the require and execute separate.

src/steps/os/archlinux.rs
atoerien: Oops I forgot about updating the locale file, forgot to add the unsupported sudo message there as well. Will fix

src/steps/os/dragonfly.rs
atoerien: Yeah I have no idea what the original reasoning behind the code here was - I don't use nor have access to a DragonFly BSD system. Here I just saw the "clippy::disallowed_methods" and tried to just rewrite it verbatim to be a bit cleaner. This should probably be made into an issue?

src/steps/os/openbsd.rs
atoerien: Yep, thanks for the spot. I knew I would have made a typo somewhere 😭

src/steps/os/windows.rs
atoerien: I assume it's because when it was written, sudo for windows wasn't a thing, and so sudo generally wouldn't be available on windows. If you then force the sudo requirement, the step would fail even if you run topgrade itself as administrator.

I think the best solution would be to add a dummy "Sudo" implementation that does nothing (just runs commands directly) if topgrade itself was run as admin/root. This could probably come with a warning on Unix systems saying that running topgrade as root is unsupported (#1114)

src/sudo.rs
atoerien: The reason for leaving it there was basically to have the "Use elevate_opts" docstring somewhere - avoiding someone doing what was done before and adding an `AsRef<OsStr> for Sudo` impl or similar. But it could still be useful to get the path to the sudo binary at some point, maybe for logging or something. That's why I did `allow(unused)` rather than entirely disallowing it.

src/steps/node.rs
GideonBear: Fair point!

src/sudo.rs
atoerien: I did do something similar initially, I added a `WinSudo`, but then I changed it because it would break config files with `sudo_command = "sudo"` on windows. However now that you mention it, `sudo_command = "sudo"` doesn't work on windows currently anyway, so it doesn't matter :) 

Actually nevermind we can just rename the variants with a `#[serde(rename = "sudo")]`. I'll add the WinSudo back then.

src/sudo.rs
atoerien: Yeah I don't particularly mind, I think I just did it to have shorter error messages initially 🤣

src/sudo.rs
atoerien: Agreed, will do

src/sudo.rs
atoerien: Yeah that's probably better, will force whoever adds a new sudo to implement the opts if possible.

src/steps/os/unix.rs
atoerien: I suppose the `..Default::default()` is a bit weird. Of those two, I'd definitely lean towards the first - will fix.

src/steps/os/unix.rs
atoerien: Huh I swear I added a comment, will add one then. It's basically the same as `execute`, but it will always run the command even with dry run enabled. The name is a bit weird, couldn't really think of anything better. It's basically meant for e.g. checks and config queries - read-only operations that should still run when dry running.

src/steps/os/unix.rs
atoerien: This is what it was before, I don't know why - I don't use nix (although that is something I've been wanting to look into for quite a while). The second argument to `execute_elevated()` specifies `interactive`.

src/steps/os/linux.rs
atoerien: Not sure I understand what you mean? Do you mean the `DETECT_ORDER` I added in `sudo.rs`? That was basically just because there needed to be a different order on windows and unix.

src/steps/os/unix.rs
GideonBear: `execute_always` maybe? `execute_atomic`? `execute_pure`?

src/steps/os/linux.rs
GideonBear: Missed that, indeed doesn't make much sense here.

src/steps/os/unix.rs
GideonBear: Ooh that's the true, it's fine then. Thx

src/steps/os/windows.rs
GideonBear: Oh yeah, missed that this was windows-only code. Thx

src/sudo.rs
GideonBear: You could also do something with removing the `Option` and using `[]` instead of `None`, and an enum for if it's all of some/none.
Don't know what the best solution is here, that's up to you

src/steps/os/openbsd.rs
GideonBear: Copy-pasting > typing ;)

themadprofessor:
Not sure about replacing `RunType` with a `bool`.
Not saying we should do this, but we could implement #598 as `RunType::Damp` for example. (More I think about it, the worse `Damp` feels)
There's also https://blakesmith.me/2019/05/07/rust-patterns-enums-instead-of-booleans.html
More than willing to conceed since it is a private field and it's not exposed.

src/steps/os/unix.rs
themadprofessor: `force_execute`? Then again, do we want to run anything if we're dry running?

GideonBear:
> Not sure about replacing `RunType` with a `bool`. Not saying we should do this, but we could implement #598 as `RunType::Damp` for example. (More I think about it, the worse `Damp` feels) There's also [blakesmith.me/2019/05/07/rust-patterns-enums-instead-of-booleans.html](https://blakesmith.me/2019/05/07/rust-patterns-enums-instead-of-booleans.html) More than willing to conceed since it is a private field and it's not exposed.

Hmm you have a point there, maybe we should look at how we're gonna implement that. I think I would be for a config option that is used in the actual executor of the command (`status_checked`), but looks like that needs something like a `RunType` as well.
Let's wait with making a decision about this until we've decided on how to implement #598.

src/steps/os/dragonfly.rs
GideonBear: #1210

src/steps/os/unix.rs
atoerien: We definitely want to still run commands if we're dry running - in many steps there are checks for the tool version (e.g. uv), or for a list of things to then loop over and update (e.g. pulling docker containers). If those checks don't run then the dry run output won't really be very useful, since it would only be able to list the first "check" command and not the commands that do the actual updating.

As for the names, idk I kinda like `execute_internal` 🤣 The alternatives to me imply something about _how_ the command will be run, that it will be executed "atomically" or "forcibly" somehow, which doesn't really make sense. It's the command itself that's "atomic", but there could potentially be a case where you want to run a "non-atomic" command even with dry run enabled, so that isn't always true.

src/sudo.rs
atoerien: I ended up doing this in a slightly weird way, basically so that "sudo" in the config file would work on windows.

Initially I tried putting the individual `WinSudo` and `Sudo` variants under `#[cfg(target_os)]` checks, but that meant all the `match kind` statements also needed `#[cfg]` checks which was super annoying. So instead I've left both `Sudo` and `WinSudo` there on both platforms, with `#[serde]` directives to have serde instantiate the correct one and `#[expect(unused)]` that will throw a lint error if `Sudo` is used on windows or vice versa.

GideonBear: That works, but can you add a comment noting that a new variant will have to be added to both, so that we don't accidentally add it to just one of the definitions?

src/steps/os/unix.rs
GideonBear: It depends on your definition of dry-running... Anyway with the comment it's good enough for me.
Now that I think about it though, I would like this to be `ctx.execute_always` or whatever, especially considering #1217 and future functionality. But this is not the only place where that occurs, so I'll open a separate issue for that.

GideonBear: (You can keep it like this if that wasn't clear. I'll fix it in the future together with other places where it occurs.)

GideonBear:
@AThePeanut4 prepare for reverting the `run_type: RunType` -> `dry_run: bool` change when #1217 is merged (which I think it will). `ctx.run_type().execute()` -> `ctx.execute()` is still more than welcome!

src/error.rs
GideonBear: If the behavior stays the same it's fine I guess. But I'll open an issue to discuss further.

src/sudo.rs
```diff
@@ -21,36 +22,108 @@ pub struct Sudo {
...
+    pub user: Option<&'a str>,
+}
+
+impl<'a> SudoExecuteOpts<'a> {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Run the command "interactively", i.e. inside a login shell.
+    #[allow(unused)]
```
GideonBear, line 57: These methods are all used, right? Can we remove `#[allow(unused)]` on all of them?

GideonBear (changes requested):

src/sudo.rs
atoerien: Unfortunately no - they're not all used on all platforms. Specifically, none of them are used on windows (because no windows program really needs "sudo" options anyway).

atoerien:
> @AThePeanut4 prepare for reverting the `run_type: RunType` -> `dry_run: bool` change when #1217 is merged (which I think it will). `ctx.run_type().execute()` -> `ctx.execute()` is still more than welcome!

No problem, I'll remove the commit from here 👍

src/sudo.rs
GideonBear: Makes sense!

src/steps/powershell.rs
```diff
@@ -75,13 +75,10 @@ impl Powershell {
     /// Builds a "primary" powershell command (uses dry-run if required):
     /// {powershell} -NoProfile -Command {cmd}
     fn build_command<'a>(&self, ctx: &'a ExecutionContext, cmd: &str, use_sudo: bool) -> Result<impl CommandExt + 'a> {
-        let executor = &mut ctx.run_type();
         let mut command = if use_sudo && ctx.sudo().is_some() {
-            let mut cmd = executor.execute(ctx.sudo().as_ref().unwrap());
-            cmd.arg(&self.path);
-            cmd
+            ctx.sudo().as_ref().unwrap().execute(ctx, &self.path)?
```
GideonBear: Why no `ctx.require_sudo().execute` here?

atoerien: Previous behaviour (even before #1197) was to fallback to not using sudo if it was unavailable. `require_sudo` would error in that case.

I think the reason was probably the same as here https://github.com/topgrade-rs/topgrade/pull/1208#discussion_r2172093503. A dummy sudo implementation like I mentioned there should allow us to use `require_sudo` in both places (and probably elsewhere as well).

src/steps/powershell.rs
```diff
@@ -75,13 +75,10 @@ impl Powershell {
...
         } else {
-            executor.execute(&self.path)
+            ctx.execute(&self.path)
```
GideonBear, line 81: (side note, shouldn't this be:)
```suggestion
            ctx.execute(&self.path)?;
```

src/steps/powershell.rs
GideonBear: Couldn't you do something like:
```rust
if use_sudo && let Ok(sudo) = ctx.require_sudo() {
    sudo.execute(&self.path)?;
}
```
To avoid an ugly `.as_ref().unwrap()`?
(or even `ctx.require_sudo_optional` which returns `Result<Option<Sudo>>`)

GideonBear: Makes sense, but can you use `match Some(sudo) = ctx.sudo()` like in the other place, to avoid the `.as_ref().unwrap()`?

atoerien: Did think of that, but forgot that you could use an if guard in the match arm. PR updated.

GideonBear:
PR is approved in this state, but blocked by #1217. Just need to rebase after that's merged (which requires moving the new `RunType` functionality).

atoerien:
@GideonBear Could we not merge this first and rebase #1217 afterwards :)

GideonBear:
> @GideonBear Could we not merge this first and rebase #1217 afterwards :)

Well, originally it was because if #1217 were to be denied, we could still do the `run_type: RunType` -> `dry_run: bool` change. But should we even do that in that case?
Mostly I thought #1217 would be merged fairly quickly (@SteveLauC said "I will take a look at this tomorrow" but that never happened)

atoerien:
> Well, originally it was because if #1217 were to be denied, we could still do the `run_type: RunType` -> `dry_run: bool` change. But should we even do that in that case?

In that case we still don't need to change `run_type` to a `bool` no, I have no problem with leaving it as an enum. Enums are theoretically more idiomatic Rust after all, and they do have advantages over a plain `bool` as mentioned above.

GideonBear:
Sure, let's merge this then

PR #1217, opened 2025-06-30
WIP: Add damp run type

### LINKED ISSUE #598: An option/config to echo dry-run commands in wet runs

themadprofessor:
## What does this PR do
Add the "damp" run type which both logs commands and runs them.
Closes #598
One possible implementation as discussed in #1208 

```
── 10:42:43 - System update ────────────────────────────────────────────────────
Executing /usr/sbin/paru -Pw
no new news
Executing /usr/sbin/paru --pacman pacman -Syu
With env: "PATH"="/usr/bin:/usr/sbin:......"
```
Trimmed down the `env` since its both long and containes names.
Since it can be so long, might be better to have it in the `debug` logging?

Also added `translate.sh` which can be used for easy i18n:
```shell
./translate.sh "String to translate"
```
outputs:
```yaml
"String to translate":
  en: "String to translate"
  lt: "translated string"
  es: "translated string"
  fr: "translated string"
  zh_CN: "translated string"
  zh_TW: "translated string"
  de: "translated string"
```

## Standards checklist

       <sub>Not the OpenBSD changes</sub>
 
## For new steps

- [x] *Optional:* The `--run-mode dry` option works with this step
  the underlying command


.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -13,13 +13,15 @@ assignees: ''
 * Which operating systems are supported by this tool?
 * What should Topgrade do to figure out if the tool needs to be invoked?
 * Which exact commands should Topgrade run?
-* Does it have a `--dry-run` option? i.e., print what should be done and exit
+* Does it have a `--run-type dry` option? i.e., print what should be done and exit
```
GideonBear: This is referring to `--dry-run` as an argument to the external program, which is probably called `--dry-run` and not `--run-type dry`.
```suggestion
* Does it have a `--dry-run` option? i.e., print what should be done and exit
```

BREAKINGCHANGES_dev.md
```diff
@@ -1,3 +1,8 @@
 1. The `jet_brains_toolbox` step was renamed to `jetbrains_toolbox`. If you're
    using the old name in your configuration file in the `disable` or `only`
    fields, simply change it to `jetbrains_toolbox`.
+
+2. The "dry run" command line argument has been replaced with the run mode
+   argument:
+    * `-n` => `-m dry`
+    * `--dry-run` => `--run-type dry`
```
GideonBear: I personally do not agree with removing `--dry-run`. I would keep it as an alias for `--run-type dry` because
1. `--dry-run` is a standard name that people know and expect
2. There's not much benefit from removing it compared to the breakage it causes

locales/app.yml
```diff
@@ -1208,30 +1208,14 @@ _version: 2
...
-  zh_TW: "會升級 OpenBSD 套件"
-  de: "Würde OpenBSD-Pakete aktualisieren"
+"Checking if /etc/motd contains -current or -beta":
+  en: "Checking if /etc/motd contains -current or -beta"
+  lt: "Tikrinimas, jei /etc/motd yra -srovė arba -beta"
+  es: "Comprobación de si /etc/motd contiene -current o -beta"
+  fr: "Vérification si /etc/ Motd contient -current ou -beta"
+  zh_CN: "检查 /etc/motd 是否包含 -current 或 -beta"
+  zh_TW: "检查 /etc/motd 是否包含 -current 或 -beta"
+  de: "Überprüfen Sie, ob /etc/motd -current oder -beta enthält"
```
GideonBear: This is phrased as a question ("(please) check if /etc/motd contains -current or -beta"), note "Sie" (meaning "you")
Better would be: "Überprüfen, ob /etc/motd -current oder -beta enthält"
```suggestion
  de: "Überprüfen, ob /etc/motd -current oder -beta enthält"
```

src/config.rs
```diff
@@ -805,9 +806,9 @@ pub struct CommandLineArgs {
     #[arg(short = 'c', long = "cleanup")]
     cleanup: bool,
 
-    /// Print what would be done
-    #[arg(short = 'n', long = "dry-run")]
-    dry_run: bool,
+    /// Pick between just running commands, running and logging commands, and just logging commands
+    #[arg(short = 'm', long = "run-type", value_enum, default_value_t)]
```
GideonBear: Why `-m`? Looks like `-r` is available.

src/executor.rs
```diff
@@ -274,3 +304,21 @@ impl CommandExt for Executor {
...
+    if env.len() != 0 {
+        println!(
+            "With env: {}",
+            env.filter(|(_, val)| val.is_some())
+                .map(|(key, val)| format!("{:?}={:?}", key, val.unwrap()))
+                .collect::<Vec<_>>()
+                .join(" ")
+        )
+    }
+}
```
GideonBear: Why do we need this when we have `Command::dry_run`? Can changes not be incorporated into `Command::dry_run`?

src/executor.rs
```diff
@@ -253,6 +275,10 @@ impl CommandExt for Executor {
     fn output_checked_with(&mut self, succeeded: impl Fn(&Output) -> Result<(), ()>) -> Result<Output> {
```
GideonBear, line 220: Can we comment on `ExecutorChild::Wet` that it represents both wet and damp outputs?

GideonBear (changes requested):

GideonBear:
Looks like a good solution, thanks!

src/config.rs
themadprofessor: For a brief moment, I forgot it was called `run-type` and thought it was `run-mode`. `-r` is better.

src/executor.rs
themadprofessor: Only `DryCommand` has `dry_run()`.

GideonBear: I don't like the duplication and disparity between `log_real_command` and `dry_run`. I would prefer then extracting this into a single function.
Also, `dry_run` uses `shell_words::join`, which should be used here as well.

locales/app.yml
themadprofessor: Google translate strikes again

.github/ISSUE_TEMPLATE/feature_request.md
themadprofessor: Ah, should have read the full file, my bad.

locales/app.yml
GideonBear: > Google translate strikes again

You sure? The version without "Sie" comes straight from Google Translate for me.

BREAKINGCHANGES_dev.md
themadprofessor: Unless I'm misunderstanding clap, it doesn't allow for aliases to set values.
Trying to alias `--dry-run` with `--run-type` leads to `--dry-run` needing a value.
We could wrap the `CommandLineArgs::parse()`, so if `--dry-run` is set, we override the `--run-type` value?

src/executor.rs
themadprofessor: That's fair. Do we want it to log how `dry_run` does (include the dir), how `log_real_command` does (include the env), or include both dir and env?

locales/app.yml
themadprofessor: Well, it was generated using the new script, which calls google translate.

BREAKINGCHANGES_dev.md
GideonBear: > it doesn't allow for aliases to set values.

This is how I did it in an application of mine:
```rust
#[derive(Args, Debug, Clone)]
pub struct TopLevelArgs {
    /// The log level to use.
    #[arg(long, short, default_value = "info")]
    log_level: String,

    /// Output debug logs. Alias for `--log-level debug`.
    #[arg(long, short)]
    verbose: bool,
}

impl TopLevelArgs {
    fn effective_log_level(&self) -> &str {
        if self.verbose {
            "debug"
        } else {
            &self.log_level
        }
    }
}
```

Just implement a method for the args that checks `if dry_run`, and then use that method instead when constructing the original `RunType`.

GideonBear: Correction, that method already exists (`run_type`)

themadprofessor:
```rust
    /// Load the command line arguments
    pub fn load() -> Self {
        // Currently, clap doesn't allow for aliases to set values, so we can't alias
        // --dry-run to --run-type dry
        let mut args = CommandLineArgs::parse();
        if args.dry_run {
            args.run_type = RunType::Dry;
        }

        args
    }
```
was what I was thinking. Either way ends up in the same place.

src/executor.rs
GideonBear: Since the directory is only added if it's `Some`, that's probably fine to add. If the env is also only added if it's explicitly added to the command construction (looks like it is, right?) then we can add both for completeness.

BREAKINGCHANGES_dev.md
GideonBear: I prefer putting it in the already existing `run_type`. Defaults and stuff are already in those getter functions.

src/executor.rs
themadprofessor: Yeah, only added env is printed. Probably should have included an example of that in the initial comment.

GideonBear (approved):

themadprofessor:
The real question is this the way we want to approach #598? Or do we want to use the log filters from the original ticket?

GideonBear:
Idk about @SteveLauC but IMO this is a better way than with log filters.

atoerien:
Alright fine you can have your enum :)

I would however strongly prefer to not use the word "Damp", but I can't think of anything that makes sense given the context 🤣

src/steps/os/macos.rs
```diff
@@ -44,7 +45,7 @@ pub fn run_mas(ctx: &ExecutionContext) -> Result<()> {
 pub fn upgrade_macos(ctx: &ExecutionContext) -> Result<()> {
     print_separator(t!("macOS system update"));
 
-    let should_ask = !(ctx.config().yes(Step::System) || ctx.config().dry_run());
+    let should_ask = !(ctx.config().yes(Step::System) || ctx.run_type() == RunType::Dry);
```
atoerien: There are a lot of other places where this kind of check is done with `.dry()` - should we prefer one or the other?

themadprofessor: Good spot. I'd say `.dry()` is better.

src/executor.rs
```diff
@@ -1,35 +1,33 @@
...
+use std::process::{Child, Command, ExitStatus, Output};
+use strum::EnumString;
 use tracing::debug;
 
-use crate::command::CommandExt;
-use crate::error::DryRun;
-
 /// An enum telling whether Topgrade should perform dry runs or actually perform the steps.
-#[derive(Clone, Copy, Debug)]
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Deserialize, Default, EnumString, ValueEnum)]
```
GideonBear:
```suggestion
#[derive(Clone, Copy, Debug, Deserialize, Default, EnumString, ValueEnum)]
```
This is not needed anymore (I checked) and removing this encourages using `.dry()`. I can't see a situation where you would need to differentiate between `Wet` and `Damp` (and we can always add another method for that then).

locales/app.yml
```diff
@@ -1208,30 +1224,14 @@ _version: 2
...
-  zh_TW: "會升級 OpenBSD 套件"
-  de: "Würde OpenBSD-Pakete aktualisieren"
+"Checking if /etc/motd contains -current or -beta":
+  en: "Checking if /etc/motd contains -current or -beta"
+  lt: "Tikrinimas, jei /etc/motd yra -srovė arba -beta"
+  es: "Comprobación de si /etc/motd contiene -current o -beta"
+  fr: "Vérification si /etc/motd contient -current ou -beta"
+  zh_CN: "检查 /etc/motd 是否包含 -current 或 -beta"
+  zh_TW: "检查 /etc/motd 是否包含 -current 或 -beta"
+  de: "Überprüfen, ob /etc/motd -current oder -beta enthält"
```
atoerien, line 1324: The `-current` and `-beta` strings should probably remain in English in all translations, as the English strings are what it's searching for in `/etc/motd`. This is one of the scenarios where `translate.sh` unfortunately won't work, using i.e. ChatGPT is needed.

atoerien:
> Trimmed down the env since its both long and containes names.
> Since it can be so long, might be better to have it in the debug logging?

I don't think anyone has commented on this yet. I think it could be logged as a debug message? The env I think is only really useful if you want to re-run the command outside of topgrade, as the env could potentially change the command's behaviour. And imo you should probably be using `-v` if you're doing something like this.

src/executor.rs
```diff
@@ -149,55 +150,71 @@ impl Executor {
...
+            Executor::Dry(_) => Ok(()),
+        }
+    }
+
+    fn log_command(&self) {
+        match self {
+            Executor::Wet(_) => (),
+            Executor::Damp(c) => {
+                log_command(
+                    "Executing {program_name} {arguments}",
```
atoerien:
```suggestion
                    "Executing: {program_name} {arguments}",
```
Doesn't match the key in `app.yml`, results in "Executing {program_name} {arguments}" messages.

src/executor.rs
```diff
@@ -149,55 +150,71 @@ impl Executor {
...
+                    c.get_program(),
+                    c.get_args(),
+                    c.get_envs(),
+                    c.get_current_dir(),
+                );
             }
+            Executor::Dry(c) => log_command(
+                "Dry running {program_name} {arguments}",
```
atoerien:
```suggestion
                "Dry running: {program_name} {arguments}",
```

AThePeanut4 (changes requested):

.github/workflows/check_config_creation_if_not_exists.yml
```diff
@@ -15,10 +15,10 @@ jobs:
...
-    - uses: actions/checkout@v5.0.0
-    - run: |
+      - uses: actions/checkout@v5.0.0
+      - run: |
           CONFIG_PATH=~/.config/topgrade.toml;
           if [ -f "$CONFIG_PATH" ]; then rm $CONFIG_PATH; fi
-          cargo build;
-          TOPGRADE_SKIP_BRKC_NOTIFY=true ./target/debug/topgrade --dry-run --only system;
+          cargo build; 
+          TOPGRADE_SKIP_BRKC_NOTIFY=true ./target/debug/topgrade --run-type dry --only system;
```
GideonBear: Can this stay `--dry-run`?

.github/workflows/check_config_creation_if_not_exists.yml
```diff
@@ -15,10 +15,10 @@ jobs:
...
     steps:
-    - uses: actions/checkout@v5.0.0
-    - run: |
+      - uses: actions/checkout@v5.0.0
+      - run: |
           CONFIG_PATH=~/.config/topgrade.toml;
           if [ -f "$CONFIG_PATH" ]; then rm $CONFIG_PATH; fi
-          cargo build;
-          TOPGRADE_SKIP_BRKC_NOTIFY=true ./target/debug/topgrade --dry-run --only system;
+          cargo build;
```
GideonBear:
```suggestion
          cargo build;
```

BREAKINGCHANGES_dev.md
```diff
@@ -1,3 +1,3 @@
 1. The `jet_brains_toolbox` step was renamed to `jetbrains_toolbox`. If you're
    using the old name in your configuration file in the `disable` or `only`
-   fields, simply change it to `jetbrains_toolbox`.
+   fields, simply change it to `jetbrains_toolbox`.
```
GideonBear, line 3: Please re-add trailing newline

src/config.rs
```diff
@@ -17,11 +11,17 @@ use regex::Regex;
 use regex_split::RegexSplit;
 use rust_i18n::t;
 use serde::Deserialize;
+use std::fs::{write, File};
+use std::io::Write;
+use std::path::{Path, PathBuf};
+use std::process::Command;
+use std::{env, fmt, fs};
```
GideonBear: Please move these back, let's keep the diff small

src/execution_context.rs
```diff
@@ -1,44 +1,42 @@
...
 use crate::{config::Config, executor::Executor};
+use clap::ValueEnum;
+use color_eyre::eyre::Result;
+use rust_i18n::t;
+use serde::Deserialize;
+use std::env::var;
+use std::ffi::OsStr;
+use std::process::Command;
+use std::sync::{LazyLock, Mutex};
+use strum::EnumString;
```
GideonBear: Please move (the `use std` lines, the non-`std` lines can stay here) these back

src/execution_context.rs
```diff
@@ -1,44 +1,42 @@
...
 
 /// An enum telling whether Topgrade should perform dry runs or actually perform the steps.
-#[derive(Clone, Copy, Debug)]
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Deserialize, Default, EnumString, ValueEnum)]
```
GideonBear: You removed `Eq, PartialEq` earlier, but they're back. Lost commits?

src/executor.rs
```diff
@@ -1,20 +1,21 @@
...
 use std::path::Path;
 use std::process::{Child, Command, ExitStatus, Output};
-
-use color_eyre::eyre::Result;
-use rust_i18n::t;
-use tracing::debug;
-
-use crate::command::CommandExt;
-use crate::error::DryRun;
+use tracing::{debug, enabled, Level};
```
GideonBear: Again

GideonBear:
Thanks for the rebase @themadprofessor!

GideonBear (approved):

PR #1230, opened 2025-07-17
Fixes for #1188; `custom_commands` broken

GideonBear:
@themadprofessor You had some mistakes in #1188 which we both didn't spot:
* `CustomCommands` is broken and running `pre_commands` instead.
* `NixHelper` and `Remotes` are missing from `default_steps`
* `Pkg` is missing from `default_steps` on `target_os = "android"`

A few other changes:
* I added a few more `*self` where I could.
* Combined `step-enum-sorted` and `step-match-sorted` into `custom-checks`
* Added `default_steps` custom check to check if it contains every step
* Refactored `default_steps` to contain less overlapping `cfg` calls

@themadprofessor please review when you can.


GideonBear:
Never mind, I cannot merge this myself. @SteveLauC could you review and merge as soon as possible?

src/step.rs
```diff
@@ -706,20 +713,20 @@ pub(crate) fn default_steps() -> Vec<Step> {
...
         Lure,
         Waydroid,
         AutoCpufreq,
         CinnamonSpices,
     ]);
 
-    #[cfg(target_os = "freebsd")]
-    steps.push(Audit);
+    #[cfg(any(target_os = "linux", target_os = "macos"))]
+    steps.push(BrewFormula);
```
atoerien: I believe this would mean BrewFormula runs before BrewCask on macOS? BrewFormula is the step that runs `brew update` to update homebrew itself, so it should run before any other homebrew steps.

src/step.rs
```diff
@@ -672,16 +675,20 @@ pub(crate) fn default_steps() -> Vec<Step> {
...
-    steps.extend_from_slice(&[BrewFormula, BrewCask, Macports, Xcodes, Sparkle, Mas]);
+    steps.extend_from_slice(&[BrewCask, Macports, Xcodes, Sparkle, Mas]);
 
-    #[cfg(target_os = "dragonfly")]
-    steps.extend_from_slice(&[Pkg, Audit]);
+    #[cfg(any(target_os = "freebsd", target_os = "dragonfly"))]
+    steps.push(Audit);
 
-    #[cfg(any(target_os = "freebsd", target_os = "openbsd"))]
+    #[cfg(any(target_os = "freebsd", target_os = "openbsd", target_os = "dragonfly"))]
```
atoerien: This also swaps the order of Pkg and Audit on Dragonfly BSD, and makes Audit run before Pkg on FreeBSD. I don't use either of these systems, but from what `pkg audit` seems to do, I assume you'd want to run it after updating your system. The order before #1188 was Pkg, System, Audit on FreeBSD and Pkg, Audit on Dragonfly BSD.

atoerien (commented):
IMO, I prefer the way step ordering was defined before #1188, with separate blocks for each OS, even if they have overlap. It makes it easier to see at a glance which steps will be run for each OS.

So I think the system-specific section of `default_steps()` should look more like this:
```rust
#[cfg(windows)]
steps.extend_from_slice(&[Wsl, WslUpdate, Chocolatey, Scoop, Winget, System, MicrosoftStore]);

#[cfg(target_os = "macos")]
steps.extend_from_slice(&[BrewFormula, BrewCask, Macports, Xcodes, Sparkle, Mas, System]);

#[cfg(target_os = "dragonfly")]
steps.extend_from_slice(&[Pkg, Audit]);

#[cfg(target_os = "freebsd")]
steps.extend_from_slice(&[Pkg, System, Audit]);

#[cfg(target_os = "openbsd")]
steps.extend_from_slice(&[Pkg, System]);

#[cfg(target_os = "android")]
steps.push(System);

#[cfg(target_os = "linux")]
steps.extend_from_slice(&[
    System,
    ConfigUpdate,
    AM,
    AppMan,
    DebGet,
    Toolbx,
    Snap,
    Pacstall,
    Pacdef,
    Protonup,
    Distrobox,
    DkpPacman,
    Firmware,
    Restarts,
    Flatpak,
    BrewFormula,
    Lure,
    Waydroid,
    AutoCpufreq,
    CinnamonSpices,
]);
```
So even though the `System` step is duplicated several times, it is really easy to see the list of steps that any particular system will run.

GideonBear:
> IMO, I prefer the way step ordering was defined before https://github.com/topgrade-rs/topgrade/pull/1188, with separate blocks for each OS, even if they have overlap. It makes it easier to see at a glance which steps will be run for each OS.

I agree actually, I did not realize that #1188 added any overlap, I thought that was there before.

I noticed "android" should run `Pkg` (#1188 forgot that), not `System`, but I mostly used your example. Thx!

themadprofessor:
Good spots. Having the `target_os` steps split up is definitely a better choice. All looks good to me.

atoerien (approved):
LGTM 👍

GideonBear:
@AThePeanut4 @themadprofessor I have a favor to ask...
Since @SteveLauC seems to be unavailable, could one of you open a new PR based on this branch, so I can merge it? That way I don't have to wait on a review from @SteveLauC. You've both reviewed it already, so I'm not concerned about skipping that review, more concerned about the main branch being broken for such a long time.
Here's how you can do it:
```bash
# On your fork
git remote add GideonBear "https://github.com/GideonBear/topgrade"
git fetch GideonBear
git switch GideonBear/fix-custom-commands --detach
git switch -c fix-custom-commands-2
git push --set-upstream origin fix-custom-commands-2
```
Then create a pr from the new branch.

GideonBear:
Superseded by #1332, because of the inability to merge this without 2nd maintainer approval.

PR #1248, opened 2025-08-06
i18n(app.yml): fix fr language

Gudsfile:
## What does this PR do

Fix for French translations. 

## Standards checklist

## For new steps

  the underlying command


locales/app.yml
```diff
@@ -418,7 +418,7 @@ _version: 2
   en: "Cannot find any of %{binary_names} in PATH"
   lt: "Nepavyksta rasti jokių %{binary_names} PATH sąraše"
   es: "No se puede encontrar ninguno de %{binary_names} en PATH"
-  fr: "Impossible de trouver %{binary_names} dans PATH"
```
GideonBear, line 421: From what Google Translate tells me, `le` is equivalent to `the` here, changing this from
`Could not find %{binary_names} in PATH`
to
`Could not find %{binary_names} in the PATH`
Which is inconsistent with the other translations.
(fully written out would be `Could not find ... in the PATH environment variable`)

Also I would probably change this to `Impossible de trouver un des {binary_names} dans PATH` anyway, to make this actually read as `any of`/`one of` instead of just `Could not find`.

GideonBear (changes requested):

locales/app.yml
Gudsfile: In French the sentence with "**le**" or **without** "**le**" doesn't have the same meaning. French uses determiners ("**le**", "**la**", "**l’**", "**les**", etc.) much more frequently than English, they are almost always used.

- `dans PATH` means in the word "path"
  For example: `Il n'y a pas de B dans phrase` means `There is no B in the word "phrase"`.
- `dans le PATH` means in PATH (whatever the PATH refers to)
    For example: `Il n'y a pas de B dans la phrase` means `There is no B in the sentence`.

So, `dans PATH` is not correct here. `le PATH` is understandable, but as you say, `la variable PATH` (the PATH variable) is more correct. It's not perfect, but I think it's sufficiently correct and understandable.

Do you prefer keeping `dans PATH` for consistency between translations or having a more fluid and natural translation? I'm fine with either.

---

You're right for the "any of", I suggest translating it as: `Impossible de trouver l'un des {binary_names} dans le PATH` (I just added "**l'**" to your suggestion).

---

It's hard to explain a language 🤯  I just realized that.

EDIT: I haven't changed anything yet, sorry for the new proofreading request.

GideonBear: A natural translation has priority, what I was mostly worried about was that sometimes the best way to describe something is in a way that the language technically does not allow; like `Could not find ... in PATH`, where `PATH` is simply short for `the PATH environment variable`. (I was worried you were trying to make the text linguistically correct, losing the programming nuance.)
Thanks for the detailed explanation, and yes I think `Impossible de trouver l'un des {binary_names} dans le PATH` is the best then!

Gudsfile: Ok I understand what you mean. It's been corrected. Thank.

GideonBear:
Thanks for the detailed explanation, I learned something today :blush:

Issue #1251, opened 2025-08-08
Some GitHub Actions are out of date


GideonBear:
We need to do some major version upgrades for Actions. We're major versions behind on some (e.g. download-artifact).

GideonBear:
`download-artifact` was just bumped by #1263 but I'll need to check if everything is up-to-date.

GideonBear:
- [x] actions/attest-build-provenance	https://github.com/actions/attest-build-provenance	v2.4.0	v2.4.0
- [x] actions/cache	https://github.com/actions/cache	v4.2.3	v4.2.4
- [x] actions/checkout	https://github.com/actions/checkout	v5.0.0	v5.0.0
- [x] actions/dependency-review-action	https://github.com/actions/dependency-review-action	da24556b548a50705dd671f47852072ea4c105d9	v4.7.1
- [x] actions/download-artifact	https://github.com/actions/download-artifact	v5.0.0	v5.0.0
- [x] (ARCHIVED) actions-rs/cargo	https://github.com/actions-rs/cargo	844f36862e911db73fe0815f00a4a2602c279505	null
- [x] (ARCHIVED) actions-rs/toolchain	https://github.com/actions-rs/toolchain	16499b5e05bf2e26879000db0c1d13f7e13fa3af	v1.0.6
- [x] actions/upload-artifact	https://github.com/actions/upload-artifact	v4.6.2	v4.6.2
- [x] github/codeql-action/upload-sarif	https://github.com/github/codeql-action/upload-sarif	v3.29.8	v3.29.9
- [x] Homebrew/actions/bump-packages	https://github.com/Homebrew/actions/bump-packages	24a0b15df658487e137fcd20fba32757d41a9411	84cd7c5
- [x] Homebrew/actions/setup-homebrew	https://github.com/Homebrew/actions/setup-homebrew	24a0b15df658487e137fcd20fba32757d41a9411	84cd7c5
- [x] microsoft/DevSkim-Action	https://github.com/microsoft/DevSkim-Action	4b5047945a44163b94642a1cecc0d93a3f428cc6	v1.0.16
- [x] ossf/scorecard-action	https://github.com/ossf/scorecard-action	05b42c624433fc40578a4040d5cf5e36ddca8cde	v2.4.2
- [x] PyO3/maturin-action	https://github.com/PyO3/maturin-action	e10f6c464b90acceb5f640d31beda6d586ba7b4a	v1.49.3
[...]

PR #1257, opened 2025-08-09
chore(ci): Dependabot, workflow security

niStee:
Follow-up to #1250. Removes harden-runner, adds least-privilege permissions, switches to shellcheck-py in pre-commit, and tunes Dependabot (groups minor/patch).

Resolve: [#GHSA-5v97-m3rg-j227](https://github.com/topgrade-rs/topgrade/security/advisories/GHSA-5v97-m3rg-j227)

fyi: @GideonBear

### DIFF
--- a/build-all.sh
+++ b/build-all.sh
@@ -1,4 +1,5 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
+
 build_function() {
     rustup update
     cargo install cross
@@ -20,7 +21,7 @@ build_function() {
 
 package_function() {
 
-    cd build
+    cd build || exit 1
     mkdir x86_64-unknown-linux-gnu/
     mkdir x86_64-unknown-linux-musl/
     mkdir x86_64-unknown-freebsd/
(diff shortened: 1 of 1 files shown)


GideonBear:
This does not resolve #1251; that is about major updates that were previously dependabot-ignored by @SteveLauC

.github/dependabot.yml
```diff
@@ -2,9 +2,38 @@
...
+      interval: "monthly"
+      day: "monday"
+      time: "06:00"
+      timezone: "UTC"
+    versioning-strategy: increase
+    open-pull-requests-limit: 2
+    labels: ["dependencies", "cargo", "manifest-update"]
+    commit-message:
+      prefix: "deps(cargo)"
+      include: "scope"
```
GideonBear: Nit: can you add a trailing newline?

.github/dependabot.yml
```diff
@@ -2,9 +2,38 @@
...
+      timezone: "UTC"
+    versioning-strategy: lockfile-only
+    open-pull-requests-limit: 5
+    labels: ["dependencies", "cargo", "safe-update"]
+    groups:
+      cargo-minor-patch:
+        update-types: ["minor", "patch"]
+  
+  # Less frequent manifest bumps (can include breaking changes)
+  - package-ecosystem: cargo
```
GideonBear: I don't think these have to be separate. Can you make one that does both? Weekly is fine, and please remove `open-pull-requests-limit`. Please set it to `versioning-strategy: increase`

.github/workflows/create_release_assets.yml
```diff
@@ -19,6 +19,9 @@ on:
         required: true
         type: string
 
+permissions:
+  contents: read
+
```
GideonBear, line 24: I am not an expert on GHA, but doesn't the `create_release_assets` need write access?

GideonBear (changes requested):

.github/workflows/create_release_assets.yml
```diff
@@ -20,7 +20,7 @@ on:
         type: string
 
 permissions:
-  contents: read
+  contents: write
```
GideonBear: Nevermind; I see it now. It's read access to the *contents*, it only writes to the release assets.
Also I didn't mean "please change this", but more "I am not sure if this is correct, and would like to verify if this is correct before merging"

.github/dependabot.yml
```diff
@@ -5,35 +5,21 @@ updates:
...
+    labels: ["dependencies", "cargo"]
     commit-message:
       prefix: "deps(cargo)"
-      include: "scope"
\ No newline at end of file
+      include: "scope"
+    groups:
+      cargo-minor-patch:
+        update-types: ["minor", "patch"]
+
```
GideonBear: ~~Trailing newline still seems messed up~~

GideonBear (changes requested):

GideonBear:
Thanks!

PR #1143, opened 2025-05-02
windows update, use explicit reboot policy

samhug:
Before this change, the `Install-WindowsUpdate` powershell command would sometimes prompt the user for reboot. See below
```
── 08:15:08 - Windows Update ───────────────────────────────────────────────────
DEBUG Executing command `C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe -NoProfile -Command 'Get-Module -ListAvailable PSWindowsUpdate'`
The installer will request to run as administrator, expect a prompt.
DEBUG Executing command `C:\Program Files\gsudo\Current\gsudo.exe 'C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe' -NoProfile 'Install-WindowsUpdate -Verbose' -AcceptAll`
VERBOSE: MyPC (5/2/2025 8:15:10 AM): Connecting to Microsoft Update server. Please wait...
VERBOSE: Found [2] Updates in pre search criteria
VERBOSE: Found [2] Updates in post search criteria

X ComputerName Result     KB          Size Title
- ------------ ------     --          ---- -----
1 MyPC         Accepted   KB5056580   72MB 2025-04 Cumulative Update Preview for .NET Framework 3.5 and 4.8.1 for Windows 11, version 23H2 for x64 (KB5056580)
1 MyPC         Accepted               23MB SATO CORPORATION - Printer - 8.4.16.27981
VERBOSE: Accepted [2] Updates ready to Download
2 MyPC         Downloaded KB5056580   72MB 2025-04 Cumulative Update Preview for .NET Framework 3.5 and 4.8.1 for Windows 11, version 23H2 for x64 (KB5056580)
2 MyPC         Downloaded             23MB SATO CORPORATION - Printer - 8.4.16.27981
VERBOSE: Downloaded [2] Updates ready to Install
3 MyPC         Installed  KB5056580   72MB 2025-04 Cumulative Update Preview for .NET Framework 3.5 and 4.8.1 for Windows 11, version 23H2 for x64 (KB5056580)
3 MyPC         Failed                 23MB SATO CORPORATION - Printer - 8.4.16.27981
VERBOSE: Installed [1] Updates
Reboot is required. Do it now? [Y / N] (default is 'N')

```
## What does this PR do

Pass the `-IgnoreReboot` option to `Install-WindowsUpdate` by default, and add a config option to pass the `-AutoReboot` option instead.

## Standards checklist

- [x] The PR title is descriptive
- [x] I have read `CONTRIBUTING.md`
[...]

PR #1274, opened 2025-08-19
Firmware updates: not prompting for reboot when `assume_yes`

lucaspar:
## What does this PR do

One of the reasons `assume_yes` is used is to reduce user prompts to a minimum. This prevents this prompt from `fwupdmgr` when updates are applied:

```
An update requires a reboot to complete. Restart now? [y|N]: 
```

## Standards checklist


GideonBear:
I disagree that assume_yes should imply ignoring reboots. Instead I would prefer a solution like #1143 separating it into another configuration value, where ignoring the reboot can be the default.

However having a single option for both is probably preferred, do you agree? We should look at things like needrestart as well to unify this, having to set only a single configuration value for asking or ignoring.

I don't have the time for that right now, but I will look at it later. Thanks!

lucaspar:
`fwupdmgr` doesn't offer an alternative for auto-reboot; our options are A. pass `--no-reboot-check` or B. require manual intervention.

Choosing B when `assume_yes` is enabled is not a good option IMO.

GideonBear:
> `fwupdmgr` doesn't offer an alternative for auto-reboot; our options are A. pass `--no-reboot-check` or B. require manual intervention.

Does `fwupdmgr` create `/var/run/reboot_required` when you select `N` or pass `--no-reboot-check`?

GideonBear:
I just realized that firmware upgrades might actually be a different case than `needrestart` and Windows Updates.
During the reboot the firmware is actually flashed (I think, and an LLM tells me), which means that if you lose power during the reboot you might brick hardware. This means **the user should be completely aware a firmware update is happening when they reboot**. So I think `--no-reboot-check` should actually be **only under a completely separate option (something like `fwupd_ignore_reboot`)**. ("should" completely based on my assumptions)
This is not at all the case for `needrestart` since Linux kernel updates and the like don't happen during a restart. Windows updates do, but are 1. basically safe to kill ("don't turn off your computer" is basically a myth, especially if you compare it to the real danger of mobo UEFI flashes), 2. probably more known by users, and 3. always clearly show "do not turn off your computer", while I wouldn't be surprised if a firmware upgrade is just a black screen.

Lmk what you think!

lucaspar:
yes, that makes sense to me

GideonBear:
If you want, you can edit this PR to do add this as a config option (like `[fwupd] ignore_reboot`) with clear documentation that this could be dangerous.

GideonBear:
Opened #1299

PR #1286, opened 2025-08-26
feat: add "show_skipped" option in config file #1280

### LINKED ISSUE #1280: add `--show-skipped` in config file
I'm missing one thing in the configuration file, the `--show-skipped` flag, we have almost all flags in the configuration file, but we are missing this one. I think this flag is very useful, I discover a bunch of tools while reading the output, and I always pass this flag for my topgrade's calls.

I don't know the difficult to add a new option in the config file, but I think is not that hard

rafaScalet:
## What does this PR do
Closes #1280

## Standards checklist

## For new steps

  the underlying command

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1488,11 +1488,13 @@ impl Config {
     }
 
     pub fn show_skipped(&self) -> bool {
-        self.config_file
-            .misc
-            .as_ref()
-            .and_then(|misc| misc.show_skipped)
-            .unwrap_or(false)
+        self.opt.show_skipped
+            || self
+                .config_file
+                .misc
+                .as_ref()
+                .and_then(|misc| misc.show_skipped)
+                .unwrap_or(false)
     }
 
     pub fn open_remotes_in_new_terminal(&self) -> bool {


GideonBear:
FYI I've edited your message to include `closes #1280` so GitHub recognizes it and links it to the issue.

config.example.toml
```diff
@@ -44,6 +44,10 @@
 # Do not ask to retry failed steps (default: false)
 # no_retry = true
 
+# Show all skipped steps at the end of a run (default: false)
+# Ignored if the "only" option is specified
```
GideonBear: 47: matches the CLI help
48: nit; it doesn't actually get ignored
```suggestion
# Show the reason for skipped steps (default: false)
# This has no effect if the "only" option is specified
```

src/config.rs
```diff
@@ -1486,7 +1488,11 @@ impl Config {
     }
 
     pub fn show_skipped(&self) -> bool {
-        self.opt.show_skipped
+        self.config_file
+            .misc
+            .as_ref()
+            .and_then(|misc| misc.show_skipped)
+            .unwrap_or(false)
```
GideonBear: This breaks `--show-skipped`. `--show-skipped` doesn't do anything anymore.
See https://github.com/topgrade-rs/topgrade/blob/e32a58f6ff56481fbcf8d313c3533bb73f94c45a/src/config.rs#L972 for an example

GideonBear (changes requested):

config.example.toml
rafaScalet: Thank you for the review, I didn't notice that the descriptions in config file matches the CLI help options

src/config.rs
rafaScalet: I think that what I do now resolve this problem. I add the same logic applied in `cleanup` function for `show_skipped` function

rafaScalet: And now, I have tested with and without the `--show-skipped` flag for the CLI

GideonBear (approved):

GideonBear:
Thanks!

PR #1306, opened 2025-09-08
feat: support for pkgfile

rafaScalet:
## Standards checklist

## For new steps

- [X] *Optional:* The `--dry-run` option works with this step
  the underlying command

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -284,6 +284,7 @@
 # Offer to update firmware; if false just check for and display available updates
 # upgrade = true
 
+
 [vagrant]
 # Vagrant directories
 # directories = []
--- a/src/steps/os/linux.rs
+++ b/src/steps/os/linux.rs
@@ -687,10 +687,9 @@ pub fn run_pkgfile(ctx: &ExecutionContext) -> Result<()> {
     let sudo = ctx.require_sudo()?;
     let pkgfile = require("pkgfile")?;
 
-    let mut cmd = sudo.execute(ctx, pkgfile)?;
     print_separator("pkgfile");
 
-    cmd.arg("--update").status_checked()
+    sudo.execute(ctx, pkgfile)?.arg("--update").status_checked()
 }
 
 pub fn run_packer_nu(ctx: &ExecutionContext) -> Result<()> {


config.example.toml
```diff
@@ -284,6 +284,11 @@
 # Offer to update firmware; if false just check for and display available updates
 # upgrade = true
 
+[pkgfile]
+# Enable the pkgfile step. PKGFILE is updated in the background by a service that is disabled by default
```
GideonBear: Typo? I thought the service was enabled by default.

locales/app.yml
```diff
@@ -1018,9 +1018,9 @@ _version: 2
   en: 'File "/etc/os-release" does not exist or is empty'
   lt: 'Failas "/etc/os-release" neegzistuoja arba yra tuščias'
   es: 'El archivo "/etc/os-release" no existe o está vacío'
-  fr: "Le fichier \"/etc/os-release\" n'existe pas ou est vide"
+  fr: 'Le fichier "/etc/os-release" n''existe pas ou est vide'
   zh_CN: "文件 “/etc/os-release” 不存在或为空"
-  zh_TW: '「/etc/os-release」不存在或為空'
+  zh_TW: "「/etc/os-release」不存在或為空"
```
GideonBear: Are these autoformatter changes? Please revert.

config.example.toml
rafaScalet: No, this tool provides a service for systemd, but is disabled by default, you can read more about [here](https://wiki.archlinux.org/title/Pkgfile#Automatic_updates).

If this isn't your expected behavior, I can change this step to be enabled by default, but with an option to disable it.

locales/app.yml
rafaScalet: Yes, it is an autoformatted change, my LSP for yaml auto format that, I will revert

config.example.toml
GideonBear: Yes, it can be enabled by default then. The whole config option can be removed then, since it can be disabled using the `disable` option.
`mandb` does have its service enabled by default, right?

rafaScalet: Ok, so I Will fix that, and yes, the mandb has a service that IS enabled by default

rafaScalet: In most distros the mandb IS enabled by default, but in distros like nixos and artix I think it is disabled by default, but I'm not sure, but even so, it's a small number of distros that don't enabled it

GideonBear: Mandb should be opt-in if it is usually done automatically.

config.example.toml
```diff
@@ -284,7 +284,6 @@
 # Offer to update firmware; if false just check for and display available updates
 # upgrade = true
 
-
```
GideonBear: Please revert

locales/app.yml
```diff
@@ -1312,3 +1312,11 @@ _version: 2
...
   zh_CN: "nh 无法找到任何已配置的 flake"
   de: "nh kann keine konfigurierten flakes finden"
+"Pkgfile isn't enabled":
+  en: "Pkgfile isn't enabled"
+  lt: "Pkgfile nėra įjungtas"
+  es: "Pkgfile no está habilitado"
+  fr: "Pkgfile n'est pas activé"
+  zh_CN: "Pkgfile 未启用"
+  zh_TW: "Pkgfile 未啟用"
+  de: "Pkgfile ist nicht aktiviert"
```
GideonBear: Unused, please revert

src/steps/os/linux.rs
```diff
@@ -683,6 +683,16 @@ pub fn run_pacstall(ctx: &ExecutionContext) -> Result<()> {
...
 }
 
+pub fn run_pkgfile(ctx: &ExecutionContext) -> Result<()> {
+    let sudo = ctx.require_sudo()?;
+    let pkgfile = require("pkgfile")?;
+
+    let mut cmd = sudo.execute(ctx, pkgfile)?;
+    print_separator("pkgfile");
+
+    cmd.arg("--update").status_checked()
```
GideonBear:
```suggestion
    print_separator("pkgfile");

    sudo.execute(ctx, pkgfile)?.arg("--update").status_checked()
```

GideonBear (approved):

GideonBear:
Thanks!

PR #1327, opened 2025-09-18
use login shell when executing topgrade

### LINKED ISSUE #1097: Missing sub-steps from WSL step
## Erroneous Behavior
When running the WSL step, topgrade doesn't run certain steps (e.g. `rustup`). When running topgrade manually inside WSL, all steps get executed as expected.

## Expected Behavior
Topgrade executing all steps that would get executed when running topgrade inside WSL manually.

## Steps to reproduce
Use `rustup` script to install rust in WSL. Install `topgrade` and `cargo-install-update` through cargo in WSL. When running topgrade in WSL, it will run the `rustup` and `cargo` steps. When running the `wsl` step in windows, only the `cargo` step in WSL is run.

## Possible Cause (Optional)
When searching for the `topgrade` binary within WSL, topgrade uses a login shell (`bash -lc ...`). But when executing topgrade in WSL, it doesn't do that (`bash -c  ...`).

- [x] Yes, because the WSL command (`C:\WINDOWS\system32\wsl.exe -d Debian bash -c 'TOPGRADE_PREFIX=Debian exec /home/{user}/.cargo/bin/topgrade'`)  is missing either `--shell-type login` as an arg to WSL, or `-i` / `-l` as an arg to bash
- [ ] No

- [ ] Yes
- [x] No

remote host
- [ ] Yes
- [ ] No

```toml

```

## Additional Details
- Operating System/Version
 Windows 10 22H2
 WSL: Debian 12

- Installation
  Windows: scoop
  WSL: cargo

  topgrade 16.0.3

<details>
<pre>

</pre>
</details>

SteveCoding125:
## What does this PR do

Closes #1097

Changes how topgrade is executed in the `WSL` step by making the shell a login shell.

When topgrade (in windows) searches for `topgrade` in wsl, it uses a login shell (`bash -lc ...`). When actually executing, it doesn't (`bash -c ...`). Because of that certain shell setup files are not loaded and topgrade (in wsl) cannot find / execute all steps.

## Standards checklist

## For new steps

  the underlying command

### DIFF
--- a/src/steps/os/windows.rs
+++ b/src/steps/os/windows.rs
@@ -149,12 +149,12 @@ fn upgrade_wsl_distribution(wsl: &Path, dist: &str, ctx: &ExecutionContext) -> R
     //
     // ```rust
     // command
-    //  .args(["-d", dist, "bash", "-c"])
+    //  .args(["-d", dist, "bash", "-lc"])
     //  .arg(format!("TOPGRADE_PREFIX={dist} exec {topgrade}"));
     // ```
     //
     // creates a command string like:
-    // > `C:\WINDOWS\system32\wsl.EXE -d Ubuntu bash -c 'TOPGRADE_PREFIX=Ubuntu exec /bin/topgrade'`
+    // > `C:\WINDOWS\system32\wsl.EXE -d Ubuntu bash -lc 'TOPGRADE_PREFIX=Ubuntu exec /bin/topgrade'`
     //
     // Adding the following:
     //
@@ -163,7 +163,7 @@ fn upgrade_wsl_distribution(wsl: &Path, dist: &str, ctx: &ExecutionContext) -> R
     // ```
     //
     // appends the next argument like so:
-    // > `C:\WINDOWS\system32\wsl.EXE -d Ubuntu bash -c 'TOPGRADE_PREFIX=Ubuntu exec /bin/topgrade' -v`
+    // > `C:\WINDOWS\system32\wsl.EXE -d Ubuntu bash -lc 'TOPGRADE_PREFIX=Ubuntu exec /bin/topgrade' -v`
     // which means `-v` isn't passed to `topgrade`.
     let mut args = String::new();
     if ctx.config().verbose() {
@@ -171,7 +171,7 @@ fn upgrade_wsl_distribution(wsl: &Path, dist: &str, ctx: &ExecutionContext) -> R
     }
 
     command
-        .args(["-d", dist, "bash", "-c"])
+        .args(["-d", dist, "bash", "-lc"])
         .arg(format!("TOPGRADE_PREFIX={dist} exec {topgrade} {args}"));
 
     if ctx.config().yes(Step::Wsl) {


GideonBear:
I recall (at least one) issues related to this

SteveCoding125:
Do you mean #1097? I mention this in the commit message, but not the pr describtion.

GideonBear:
> Do you mean #1097? I mention this in the commit message, but not the pr describtion.

Yes, ah I see it in the commit now. I will edit the PR description to contain the issue as well, this makes it clear to GitHub that they are linked.

GideonBear:
From `man bash`:

> INVOCATION
>  A login shell is one whose first character of argument zero is a -, or one started with the --login option.
> 
>  An  interactive shell is one started without non-option arguments (unless -s is specified) and without the -c option, whose standard input and error are both connected to terminals (as determined by isatty(3)), or one started
>  with the -i option.  PS1 is set and $- includes i if bash is interactive, allowing a shell script or a startup file to test this state.
> 
>  The following paragraphs describe how bash executes its startup files.  If any of the files exist but cannot be read, bash reports an error.  Tildes are expanded in filenames as described below under Tilde  Expansion  in  the
>  EXPANSION section.
> 
>  When  bash  is  invoked  as an interactive login shell, or as a non-interactive shell with the --login option, it first reads and executes commands from the file /etc/profile, if that file exists.  After reading that file, it
>  looks for ~/.bash_profile, ~/.bash_login, and ~/.profile, in that order, and reads and executes commands from the first one that exists and is readable.  The --noprofile option may be used when the shell is started to inhibit
>  this behavior.
> 
>  When an interactive login shell exits, or a non-interactive login shell executes the exit builtin command, bash reads and executes commands from the file ~/.bash_logout, if it exists.
> 
>  When an interactive shell that is not a login shell is started, bash reads and executes commands from /etc/bash.bashrc and ~/.bashrc, if these files exist.  This may be inhibited by using the --norc option.  The --rcfile file
>  option will force bash to read and execute commands from file instead of /etc/bash.bashrc and ~/.bashrc.
> 
>  When bash is started non-interactively, to run a shell script, for example, it looks for the variable BASH_ENV in the environment, expands its value if it appears there, and uses the expanded value as the name of  a  file  to
>  read and execute.  Bash behaves as if the following command were executed:
>         if [ -n "$BASH_ENV" ]; then . "$BASH_ENV"; fi
>  but the value of the PATH variable is not used to search for the filename.
> 
>  If  bash is invoked with the name sh, it tries to mimic the startup behavior of historical versions of sh as closely as possible, while conforming to the POSIX standard as well.  When invoked as an interactive login shell, or
>  a non-interactive shell with the --login option, it first attempts to read and execute commands from /etc/profile and ~/.profile, in that order.  The --noprofile option may be used to inhibit this behavior.  When  invoked  as
>  an  interactive shell with the name sh, bash looks for the variable ENV, expands its value if it is defined, and uses the expanded value as the name of a file to read and execute.  Since a shell invoked as sh does not attempt
>  to read and execute commands from any other startup files, the --rcfile option has no effect.  A non-interactive shell invoked with the name sh does not attempt to read any other startup files.  When invoked as sh,  bash  en‐
>  ters posix mode after the startup files are read.
> 
>  When  bash  is started in posix mode, as with the --posix command line option, it follows the POSIX standard for startup files.  In this mode, interactive shells expand the ENV variable and commands are read and executed from
>  the file whose name is the expanded value.  No other startup files are read.
> 
>  Bash attempts to determine when it is being run with its standard input connected to a network connection, as when executed by the historical remote shell daemon, usually rshd, or the secure shell daemon sshd.  If bash deter‐
>  mines it is being run non-interactively in this fashion, it reads and executes commands from /etc/bash.bashrc and ~/.bashrc, if these files exist and are readable.  It will not do this if invoked as sh.  The --norc option may
>  be used to inhibit this behavior, and the --rcfile option may be used to force another file to be read, but neither rshd nor sshd generally invoke the shell with those options or allow them to be specified.
> 
>  If the shell is started with the effective user (group) id not equal to the real user (group) id, and the -p option is not supplied, no startup files are read, shell functions are not inherited from the environment, the SHEL‐
>  LOPTS, BASHOPTS, CDPATH, and GLOBIGNORE variables, if they appear in the environment, are ignored, and the effective user id is set to the real user id.  If the -p option is supplied at invocation, the startup behavior is the
>  same, but the effective user id is not reset.

Could you explain why we want to use `-l` instead of `-i` here?

SteveCoding125:
I don't have a preference for any one of those settings. I just used `-l` since we already use it [here](https://github.com/topgrade-rs/topgrade/blob/ccefd0a43a83592d5e5d8f3974d608d97ec96b02/src/steps/os/windows.rs#L135) for checking if topgrade is installed in WSL.
For me (Debian in WSL), both `-i` and `-l` would fix #1097, since `~/.profile` sources `~/.bashrc` by default.
`~/.profile` also adds `~/bin` and `~/.local/bin` to `$PATH`. I think this would make it more likely to find **all** installed programs, that could be upgraded. I have not checked how `~/.profile` behaves on other distros by default.

I checked a few programs to see how they behave:
- `pipx` sets its paths in both `~/.profile` and `~/.bashrc`
- `cargo` does the same
- `go` suggests to use `~/.profile` in the install instructions
- `bun` uses `~/.bashrc` in its install script

GideonBear (approved):

GideonBear:
Thanks!

Issue #1297, opened 2025-09-04
Does not wait for Chocolatey before proceeding


selfagency:
## Erroneous Behavior

Chocolatey gets launched in a new terminal window but Topgrade doesn't wait for it to finish, and moves on to the next step. In my case, Scoop runs right after Chocolatey, which can result in two installers for the same app running because both Chocolatey and Scoop will attempt to update them.

## Expected Behavior

For Topgrade to wait for Chocolatey to finish running before moving on.

## Steps to reproduce

Install Chocolatey and Scoop on Windows 11.

## Possible Cause (Optional)

Topgrade isn't expecting Chocolatey to run in a new window and therefore isn't able to determine whether it's finished running.

## Problem persists without calling from topgrade

- [ ] Yes
- [X] No

## Did you run topgrade through `Remote Execution`

- [ ] Yes
- [X] No

If yes, does the issue still occur when you run topgrade directlly in your
remote host

- [ ] Yes
- [ ] No

## Configuration file (Optional)

```toml
# Include any additional configuration file(s)
# [include] sections are processed in the order you write them
# Files in $CONFIG_DIR/topgrade.d/ are automatically included before this file
[include]
# paths = ["/etc/topgrade.toml"]


[misc]
# Run `sudo -v` to cache credentials at the start of the run
# This avoids a blocking password prompt in the middle of an unattended run
# (default: false)
# pre_sudo = false

# Sudo command to be used
# sudo_command = "sudo"

# Disable specific steps - same options as the command line flag
disable = ["vscode", "system"]

# Ignore failures for these steps
# ignore_failures = ["powershell"]

# List of remote machines with Topgrade installed on them
# remote_topgrades = ["toothless", "pi", "parnas"]

# Path to Topgrade executable on remote machines
# remote_topgrade_path = ".cargo/bin/topgrade"

# Arguments to pass to SSH when upgrading remote systems
# ssh_arguments = "-o ConnectTimeout=2"

# Arguments to pass tmux when pulling Repositories
# tmux_arguments = "-S /var/tmux.sock"

# Do not set the terminal title (default: true)
# set_title = true

# Display the time in step titles (default: true)
# display_time = true

# Don't ask for confirmations (no default value)
# assume_yes = true

# Do not ask to retry failed steps (default: false)
# no_retry = true

# Run inside tmux (default: false)
# run_in_tmux = true

# Cleanup temporary or old files (default: false)
# cleanup = true

# Send a notification for every step (default: false)
[...]

Issue #1025, opened 2025-02-01
[Windows 11] Topgrade uses Sudo when run as administrator


GideonBear:
## I want to suggest some general feature
In Windows 11, Topgrade should not use Sudo when run as administrator, but run the command directly.

## More information
In Windows 10 running as administrator worked fine, but in Windows 11 I have to press "Yes" on Sudo's prompt multiple times (Windows Update, Microsoft Store). In Windows 10 some steps would simply fail if not run as administrator, but this is no better.

SteveLauC:
Emm, I am no Windows expert, so I try to mirror my UNIX knowledge here.

If `sudo` is needed, shouldn't Topgrade use it? I see "press "Yes" on Sudo's prompt" as a necessary step to make `sudo` work. On Linux, sudo privilege will be [cached](https://unix.stackexchange.com/q/77356/498440), wouldn't Windows sudo handle that?

GideonBear:
> Emm, I am no Windows expert, so I try to mirror my UNIX knowledge here.
> 
> If `sudo` is needed, shouldn't Topgrade use it? I see "press "Yes" on Sudo's prompt" as a necessary step to make `sudo` work. On Linux, sudo privilege will be [cached](https://unix.stackexchange.com/q/77356/498440), wouldn't Windows sudo handle that?

I don't see why Sudo is needed when Topgrade itself is ran as administrator, as it can just spawn processes with administrator permissions. What mostly confuses me is that Topgrade seemed to be supposed to be run as Administrator in Windows 10 (as that was the only way to get (IIRC) Windows Update to work), but in Windows 11 suddenly uses Sudo regardless of privilege state.
Anyway referencing #1015 as well (although that's on Linux), I think this behavior is unacceptable, regardless of the method to solve it. Winget installers sometimes ask me to press "Yes" as well, but that is more of a Winget problem and not really an issue; but having to do it every time, even when nothing is upgraded, is just annoying, and requires me to monitor my Topgrade run constantly.
Please don't interpret my “unacceptable” negatively though, I understand that you're not developing this for Windows primarily, and I still think Topgrade is an amazing tool, even (maybe even especially) on Windows :)

[...]

PR #1341, opened 2025-09-26
Detect when running as root/elevated, and skip using sudo in this case

### LINKED ISSUE #1114: Warn when running Topgrade as superuser on non-Windows systems

AThePeanut4:
## What does this PR do

Adds a warning message if running as root, followed by a yes/no confirmation prompt.

Also adds a "null" sudo, which just runs the command directly. If Topgrade is running as root/elevated, this is used instead of calling `Sudo::detect()`.

On Windows, it doesn't print the warning, as running commands as Administrator usually doesn't cause the same kind of problems as on Unix systems. I've used the single-function `is_elevated` crate to do the elevated check here.

## Notes

- I haven't excluded the "null" sudo from being set in the config file, so you could use `sudo_command = "null"` to get the null sudo even when not running as administrator. Not sure why you'd ever want to do that, but maybe it could be useful somewhere.
- Also if a `sudo_command` is set in the config file, that is always used even when running as root/administrator. I wasn't sure whether to do it this way or have it override the config file.
- I also added a yes/no prompt after the "don't run as root" warning message. Should there be some way of skipping this prompt, a command-line flag maybe, or is that unnecessary? The prompt was mainly to give i.e. a first-time user the opportunity to quit before any steps are inadvertently run as root.

## Standards checklist

Closes #1114

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -27,6 +27,7 @@ use self::error::StepFailed;
 use self::error::Upgraded;
 #[allow(clippy::wildcard_imports)]
 use self::steps::{remote::*, *};
+use self::sudo::{Sudo, SudoKind};
 #[allow(clippy::wildcard_imports)]
 use self::terminal::*;
 use self::utils::{install_color_eyre, install_tracing, is_elevated, update_tracing};
@@ -146,10 +147,16 @@ fn run() -> Result<()> {
         }
     }
 
+    let sudo = match config.sudo_command() {
+        Some(kind) => Sudo::new(kind),
+        None if elevated => Sudo::new(SudoKind::Null),
+        None => Sudo::detect(),
+    };
+    debug!("Sudo: {:?}", sudo);
+
     #[cfg(target_os = "linux")]
     let distribution = linux::Distribution::detect();
 
-    let sudo = config.sudo_command().map_or_else(sudo::Sudo::detect, sudo::Sudo::new);
     let run_type = execution_context::RunType::new(config.dry_run());
     let ctx = execution_context::ExecutionContext::new(
         run_type,
--- a/src/main.rs
+++ b/src/main.rs
@@ -29,7 +29,7 @@ use self::error::Upgraded;
 use self::steps::{remote::*, *};
 #[allow(clippy::wildcard_imports)]
 use self::terminal::*;
-use self::utils::{install_color_eyre, install_tracing, update_tracing};
+use self::utils::{install_color_eyre, install_tracing, is_elevated, update_tracing};
 
 mod breaking_changes;
 mod command;
@@ -134,6 +134,18 @@ fn run() -> Result<()> {
         }
     }
 
+    let elevated = is_elevated();
+
+    #[cfg(unix)]
+    if !config.allow_root() && elevated {
+        print_warning(t!(
+            "Topgrade should not be run as root, it will run commands with sudo or equivalent where needed."
+        ));
+        if !prompt_yesno(&t!("Continue?"))? {
+            exit(1)
+        }
+    }
+
     #[cfg(target_os = "linux")]
     let distribution = linux::Distribution::detect();
 
@@ -157,7 +169,7 @@ fn run() -> Result<()> {
     if !should_skip() && first_run_of_major_release()? {
         print_breaking_changes();
 
-        if prompt_yesno("Confirmed?")? {
+        if prompt_yesno(&t!("Continue?"))? {
             write_keep_file()?;
         } else {
             exit(1);
--- a/src/sudo.rs
+++ b/src/sudo.rs
@@ -38,8 +38,8 @@ pub enum SudoPreserveEnv<'a> {
 /// default or unsupported.
 #[derive(Clone, Debug, Default)]
 pub struct SudoExecuteOpts<'a> {
-    /// Run the command "interactively", i.e. inside a login shell.
-    pub interactive: bool,
+    /// Run the command inside a login shell.
+    pub login_shell: bool,
     /// Preserve environment variables across the sudo call.
     pub preserve_env: SudoPreserveEnv<'a>,
     /// Set the HOME environment variable to the target user's home directory.
@@ -53,10 +53,10 @@ impl<'a> SudoExecuteOpts<'a> {
         Self::default()
     }
 
-    /// Run the command "interactively", i.e. inside a login shell.
+    /// Run the command inside a login shell.
     #[allow(unused)]
-    pub fn interactive(mut self) -> Self {
-        self.interactive = true;
+    pub fn login_shell(mut self) -> Self {
+        self.login_shell = true;
         self
     }
 
@@ -221,11 +221,11 @@ impl Sudo {
     ) -> Result<Executor> {
         // null sudo is very different, do separately
         if let SudoKind::Null = self.kind {
-            if opts.interactive {
+            if opts.login_shell {
                 // TODO: emulate running in a login shell with su/runuser
                 return Err(UnsupportedSudo {
                     sudo_kind: self.kind,
-                    option: "interactive",
+                    option: "login_shell",
                 }
                 .into());
             }
@@ -248,20 +248,19 @@ impl Sudo {
         // self.path is only None for null sudo, which we've handled above
         let mut cmd = ctx.execute(self.path.as_ref().unwrap());
 
-        if opts.interactive {
+        if opts.login_shell {
             match self.kind {
                 SudoKind::Sudo => {
                     cmd.arg("-i");
                 }
                 SudoKind::Gsudo => {
-                    // By default, gsudo runs all commands inside a shell, so it's effectively
-                    // always "interactive". If interactive is *not* specified, we add `-d`
-                    // to run outside of a shell - see below.
+                    // By default, gsudo runs all commands inside a shell. If login_shell
+                    // is *not* specified, we add `-d` to run outside of a shell - see below.
                 }
                 SudoKind::Doas | SudoKind::WinSudo | SudoKind::Pkexec | SudoKind::Run0 | SudoKind::Please => {
                     return Err(UnsupportedSudo {
                         sudo_kind: self.kind,
-                        option: "interactive",
+                        option: "login_shell",
                     }
                     .into());
                 }
@@ -269,7 +268,7 @@ impl Sudo {
             }
         } else if let SudoKind::Gsudo = self.kind {
             // The `-d` (direct) flag disables shell detection, running the command directly
-            // rather than through the current shell, making it "non-interactive".
+            // rather than through the current shell.
             // Additionally, if the current shell is pwsh >= 7.3.0, then not including this
             // gives errors if the command to run has spaces in it: see
             // https://github.com/gerardog/gsudo/issues/297
--- a/src/sudo.rs
+++ b/src/sudo.rs
@@ -17,7 +17,7 @@ use crate::utils::which;
 #[derive(Clone, Debug)]
 pub struct Sudo {
     /// The path to the `sudo` binary.
-    path: PathBuf,
+    path: Option<PathBuf>,
     /// The type of program being used as `sudo`.
     kind: SudoKind,
 }
@@ -105,8 +105,8 @@ impl Sudo {
     /// Get the `sudo` binary for this platform.
     pub fn detect() -> Option<Self> {
         for kind in DETECT_ORDER {
-            if let Some(path) = kind.which() {
-                return Some(Self { path, kind });
+            if let Some(sudo) = Self::new(kind) {
+                return Some(sudo);
             }
         }
         None
@@ -114,7 +114,13 @@ impl Sudo {
 
     /// Create Sudo from SudoKind, if found in the system
     pub fn new(kind: SudoKind) -> Option<Self> {
-        kind.which().map(|path| Self { path, kind })
+        match kind {
+            SudoKind::Null => Some(Self {
+                path: None, // no actual binary for null sudo
+                kind,
+            }),
+            _ => kind.which().map(|path| Self { path: Some(path), kind }),
+        }
     }
 
     /// Gets the path to the `sudo` binary. Do not use this to execute `sudo` directly - either use
@@ -122,8 +128,8 @@ impl Sudo {
     /// This way, sudo options can be specified generically and the actual arguments customized
     /// depending on the sudo kind.
     #[allow(unused)]
-    pub fn path(&self) -> &Path {
-        self.path.as_ref()
+    pub fn path(&self) -> Option<&Path> {
+        self.path.as_deref()
     }
 
     /// Elevate permissions with `sudo`.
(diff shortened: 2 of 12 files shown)


atoerien:
@niStee Regarding https://github.com/topgrade-rs/topgrade/issues/1297#issuecomment-3259621349, I decided to just print a warning message and get the user to change the sudo configuration, rather than changing it ourselves. The problem is that the configuration is global - so even if you do restore the old mode after topgrade exists, it is globally set to "inline" mode while topgrade is running, which is theoretically less secure and so the user might not want that.

Also changing the mode without the user doing it directly requires a working sudo anyway (`sudo config --enable normal` needs to run as administrator, so you need to do `sudo sudo config --enable normal`). So it's really only useful to automatically change between the 'In a new window' and 'Inline' modes.

GideonBear:
> I'm also happy to move any of the refactoring into separate PR(s) if needed. Most of it was just moving things around, and the winapi -> windows migration was like a 5 line change.

Yeah, I would appreciate it if you split this up into a couple PRs. This is a bit hard to review.
Note that I don't mind a PR depending on another one, if that's easier for you! You don't have to submit them all at once.

src/main.rs
```diff
@@ -135,10 +136,28 @@ fn run() -> Result<()> {
         }
     }
 
+    let elevated = is_elevated();
+
+    #[cfg(unix)]
+    if elevated {
+        print_warning(t!(
+            "Topgrade does not need to be run as root, it will run commands with sudo or equivalent where needed."
```
GideonBear: "Does not need to", sure; but maybe also "shouldn't be"?
"Topgrade does not need to, and shouldn't be, run as root, it will [...]"

src/main.rs
```diff
@@ -135,10 +136,28 @@ fn run() -> Result<()> {
...
+        ));
+        if !prompt_yesno(&t!("Continue?"))? {
```
GideonBear, line 145: There should be a way to disable this prompt. `allow_root`? This should also be mentioned in the message then.

To be clear: originally with "warn" I meant `warn!()` and just continue on. Stopping and asking for confirmation is probably better though.

src/sudo.rs
```diff
@@ -205,6 +211,30 @@ impl Sudo {
...
+                }
+                .into());
+            }
+            if opts.user.is_some() {
+                return Err(UnsupportedSudo {
+                    sudo_kind: self.kind,
+                    option: "user",
+                }
+                .into());
+            }
```
GideonBear: Could you add an explanation as to why we want to fail here? Or in other words, why running as super-user doesn't have the same functionality as `sudo --interactive` or `sudo --user`?

src/sudo.rs
```diff
@@ -343,6 +378,8 @@ pub enum SudoKind {
     Pkexec,
     Run0,
     Please,
+    /// A "no-op" sudo, used when topgrade itself is running as Administrator
+    Null,
 }
 
 #[derive(Clone, Copy, Debug, Display, Deserialize)]
```
GideonBear: Forgive me if I've asked this before, but why don't we put `#[cfg(target_os = "windows")]` on only the `Sudo` and `WinSudo`?

src/sudo.rs
```diff
@@ -373,10 +412,14 @@ impl SudoKind {
...
             SudoKind::Run0 => "run0",
             SudoKind::Please => "please",
+            SudoKind::Null => "<null>", // dummy sudo, no actual binary
         }
     }
 
     fn which(self) -> Option<PathBuf> {
-        which(self.binary_name())
+        match self {
+            SudoKind::Null => Some(self.binary_name().into()), // no actual binary for null sudo
```
GideonBear: This will cause weird errors if we use `which`. What does `None` represent currently, can we use `None` instead?

GideonBear (changes requested):

src/main.rs
atoerien: In that case I'd say it should just be "Topgrade should not be run as root, ...". Will change.

src/main.rs
atoerien: Should it be in the config file, a command-line parameter, or both? I guess might as well just do both, why not.

I also initially just had it as a warning without the prompt, but then I thought that a user might want to cancel after seeing the warning so that no steps actually get run as root. An alternative I thought of was to just sleep for a few seconds to give the user a chance to hit `Ctrl-C`, but I decided to go for the prompt in the end.

src/sudo.rs
atoerien: `sudo -i` runs the command through a login shell, so the shell startup files are sourced (`.profile`, `.bashrc`, etc.). `sudo -u` runs the command as a different (i.e. not root) user.

Running the command directly (with Topgrade running as root) doesn't do either of those. We should be able to emulate both using external commands though, I just haven't implemented that.

I'll add a comment.

src/sudo.rs
atoerien: The "best" way would be this:
```rust
#[derive(Clone, Copy, Debug, Display, Deserialize)]
#[serde(rename_all = "lowercase")]
#[strum(serialize_all = "lowercase")]
pub enum SudoKind {
    Doas,
    #[cfg(not(target_os = "windows"))]
    Sudo,
    #[cfg(target_os = "windows")]
    #[serde(rename = "sudo")]
    WinSudo,
    Gsudo,
    Pkexec,
    Run0,
    Please,
    /// A "no-op" sudo, used when topgrade itself is running as root
    Null,
}
```
But that means we need `#[cfg(...)]`s every time `Sudo` or `WinSudo` is used, which makes things quite messy.

Another alternative, which I assume is what you're referring to, is this:
```rust
#[derive(Clone, Copy, Debug, Display, Deserialize)]
#[serde(rename_all = "lowercase")]
#[strum(serialize_all = "lowercase")]
pub enum SudoKind {
    Doas,
    #[cfg(not(target_os = "windows"))]
    Sudo,
    #[cfg(target_os = "windows")]
    #[expect(unused, reason = "Sudo is unix-only")]
    #[serde(skip)]
    Sudo,
    #[cfg(not(target_os = "windows"))]
    #[expect(unused, reason = "WinSudo is windows-only")]
    #[serde(skip)]
    WinSudo,
    #[cfg(target_os = "windows")]
    #[serde(rename = "sudo")]
    WinSudo,
    Gsudo,
    Pkexec,
    Run0,
    Please,
    /// A "no-op" sudo, used when topgrade itself is running as root
    Null,
}
```

That avoids the `#[cfg(...)]`s everywhere, and is shorter overall than the current version, though it is perhaps a bit more complicated to read.

I'm happy to change to either though.

src/sudo.rs
atoerien: If `SudoKind::which()` returns `None`, then `Sudo::new(SudoKind::Null)` will always return `None`:
https://github.com/topgrade-rs/topgrade/blob/8fff8ca649fcf2407d9ed9c721b005986335cc25/src/sudo.rs#L115-L118

The problem is that the assumption is that all `SudoKind`s have a binary: `SudoKind::binary_name()` returns a `&str`, `Sudo` has a `path: PathBuf` field. But `SudoKind::Null` obviously doesn't have a binary.

We can certainly move the special case up into `Sudo::new()`. `Sudo::path` would still have to be "fake" though.

niStee:
Would it be feasible and desirable to use 'sudo' just once to elevate privileges? In Unix, you enter the 'sudo' password once and then run all subsequent commands with elevated privileges. Wouldn't this be a good option? Currently, we have to use the UAC prompt for each step when running a command with 'sudo', even when the 'yes' flag is set.

src/sudo.rs
GideonBear: Indeed; you could use `bash -i`. But that can come later

src/sudo.rs
GideonBear: Yes, I think I prefer the second version to avoid the duplication. A few tips to make it maybe easier to read:
1. Use `cfg(windows)` instead of `cfg(target_os = windows)`
2. Put `Sudo` and `WinSudo` at the top
3. Add line breaks
4. Add comments
```rust
#[derive(Clone, Copy, Debug, Display, Deserialize)]
#[serde(rename_all = "lowercase")]
#[strum(serialize_all = "lowercase")]
pub enum SudoKind {
    #[cfg(not(windows))] // On non-Windows:
    Sudo,
    #[cfg(windows)] // On Windows:
    #[expect(unused, reason = "Sudo is unix-only")]
    #[serde(skip)]
    Sudo,

    #[cfg(not(windows))] // On non-Windows:
    #[expect(unused, reason = "WinSudo is windows-only")]
    #[serde(skip)]
    WinSudo,
    #[cfg(windows)] // On Windows:
    #[serde(rename = "sudo")]
    WinSudo,

    Doas,
    Gsudo,
    Pkexec,
    Run0,
    Please,
    /// A "no-op" sudo, used when topgrade itself is running as root
    Null,
}
```

src/sudo.rs
GideonBear: Oh yeah, `None` means not found.
Wouldn't it be better to make `Sudo.path` (and with it, `SudoKind::binary_name()`) an `Option<PathBuf>`?
Or in other words, remove the assumption that all `SudoKind`s have a binary?

GideonBear:
> Would it be feasible and desirable to use 'sudo' just once to elevate privileges? In Unix, you enter the 'sudo' password once and then run all subsequent commands with elevated privileges. Wouldn't this be a good option? Currently, we have to use the UAC prompt for each step when running a command with 'sudo', even when the 'yes' flag is set.

Do you mean on Windows, or on Unix?

src/utils.rs
```diff
@@ -169,6 +169,22 @@ pub fn hostname() -> Result<String> {
         .map(|output| output.stdout.trim().to_owned())
 }
 
+#[cfg(target_family = "unix")]
```
```suggestion
#[cfg(unix)]
```

src/utils.rs
```diff
@@ -169,6 +169,22 @@ pub fn hostname() -> Result<String> {
...
+pub fn is_elevated() -> bool {
+    let euid = nix::unistd::Uid::effective();
+    debug!("Running with euid: {euid}");
+    euid.is_root()
+}
+
+#[cfg(target_family = "windows")]
```
```suggestion
#[cfg(windows)]
```

atoerien:
> Would it be feasible and desirable to use 'sudo' just once to elevate privileges? In Unix, you enter the 'sudo' password once and then run all subsequent commands with elevated privileges. Wouldn't this be a good option? Currently, we have to use the UAC prompt for each step when running a command with 'sudo', even when the 'yes' flag is set.

@niStee This is #1025. It's definitely desirable, but unfortunately not really feasible, see my comment https://github.com/topgrade-rs/topgrade/issues/1025#issuecomment-3012931172. On Unix `sudo` handles that, it caches credentials and doesn't ask for a password multiple times (unless that functionality is disabled). For Topgrade to emulate that behaviour will be very complicated, it's really something that Microsoft should add to Windows Sudo instead.

src/sudo.rs
```diff
@@ -322,61 +365,62 @@ impl Sudo {
...
+    #[serde(skip)]
+    WinSudo,
+
+    // On unix, Sudo is skipped and unused
+    #[cfg(windows)] // On Windows:
     #[expect(unused, reason = "Sudo is unix-only")]
     #[serde(skip)]
     Sudo,
+    // and "sudo" in the config file means WinSudo.
+    #[cfg(windows)] // On Windows:
```
GideonBear:
```suggestion
    #[cfg(windows)]
```

src/sudo.rs
```diff
@@ -322,61 +365,62 @@ impl Sudo {
...
+    #[cfg(not(windows))]
+    Sudo,
+    // and WinSudo is skipped, making it unused.
+    #[cfg(not(windows))]
+    #[expect(unused, reason = "WinSudo is windows-only")]
+    #[serde(skip)]
+    WinSudo,
+
+    // On unix, Sudo is skipped and unused
+    #[cfg(windows)] // On Windows:
```
GideonBear: Was this intentional? This reads a bit weird.
The new comments are much better!
```suggestion
    #[cfg(windows)]
```

src/sudo.rs
```diff
@@ -205,7 +217,33 @@ impl Sudo {
         command: S,
         opts: SudoExecuteOpts,
     ) -> Result<Executor> {
-        let mut cmd = ctx.execute(&self.path);
+        // null sudo is very different, do separately
+        if let SudoKind::Null = self.kind {
+            if opts.interactive {
+                // TODO: emulate running in a login shell with su/runuser
```
GideonBear, line 225: Nit: there's a distinction between login shells and interactive shells
```suggestion
                // TODO: emulate running in an interactive shell with su/runuser
```

src/sudo.rs
```diff
@@ -322,61 +365,62 @@ impl Sudo {
...
     Pkexec,
     Run0,
     Please,
+    /// A "no-op" sudo, used when topgrade itself is running as root
+    Null,
 }
 
 impl SudoKind {
-    fn binary_name(self) -> &'static str {
+    fn binary_name(self) -> Option<&'static str> {
```
GideonBear, line 414:
```suggestion
    /// Returns `None` for `SudoKind::Null`.
    fn binary_name(self) -> Option<&'static str> {
```

src/sudo.rs
```diff
@@ -205,7 +217,33 @@ impl Sudo {
...
+            }
+
+            // NOTE: we ignore preserve_env and set_home, using
+            // no sudo effectively preserves these by default
+
+            // run command directly
+            return Ok(ctx.execute(command));
+        }
+
+        let mut cmd = ctx.execute(self.path.as_ref().unwrap());
```
GideonBear, line 249:
```suggestion
		// `self.path` is only `None` for `SudoKind::Null`, which we just handled
        let mut cmd = ctx.execute(self.path.as_ref().unwrap());
```

src/sudo.rs
atoerien: Oops :see_no_evil:

src/sudo.rs
atoerien: There is - but (unfortunately?) `sudo -i`, which is what the `interactive` option does, means a *login* shell:
```
-i, --login
        Run  the shell specified by the target user's password database entry as a login shell.  This means that login-specific resource files such as .profile, .bash_profile, or .login will be read by
        the shell.  If a command is specified, it is passed to the shell as a simple command using the -c option.  The command and any args are concatenated, separated by spaces,  after  escaping  each
        character  (including  white space) with a backslash (‘\’) except for alphanumerics, underscores, hyphens, and dollar signs.  If no command is specified, an interactive shell is executed.  sudo
        attempts to change to that user's home directory before running the shell.  The command is run with an environment similar to the one a user  would  receive  at  log  in.   Most  shells  behave
        differently  when  a  command is specified as compared to an interactive session; consult the shell's manual for details.  The Command environment section in the sudoers(5) manual documents how
        the -i option affects the environment in which a command is run when the sudoers policy is in use.
```
(from `man sudo`)

Maybe `login` or `login_shell` would have been a better name for the `interactive` option?

GideonBear: Oh no...
Bash has `-i` (interactive) and `--login` (login). Let's rename `interactive` to `login_shell` then.

Issue #1136, opened 2025-04-19
WSL Integration Fails: Error with WSL Distribution Detection and Topgrade Upgrade (in commit a2f57e4769115291bd78b33c1252930bc0ce6e9a aka main)


niStee:
### Description

When running topgrade on Windows with WSL enabled, the update procedure fails during the upgrade step. The verbose output provides multiple clues:

- The command to list distributions returns ["Debian", ""]—indicating an empty/discrepant WSL distribution name.
- The command to execute topgrade in the Debian distribution is formed as follows:  
  `C:\WINDOWS\system32\wsl.exe -d Debian bash -c 'TOPGRADE_PREFIX=Debian exec /home/nils/.cargo/bin/topgrade -v'`  
  This command fails with exit code 1.
- Later, a command attempts to locate topgrade by running:  
  `C:\WINDOWS\system32\wsl.exe -d '' bash -lc 'which topgrade'`  
  which clearly fails because an empty distribution name was provided, giving the error:  
  "There is no distribution with the supplied name. Error code: Wsl/Service/WSL_E_DISTRO_NOT_FOUND."
- The Windows side then continues to execute `C:\WINDOWS\system32\wsl.exe --update`, which confirms that the latest WSL version is installed, but then WSL panics with the message:  
  `WSL already reported`

This behavior seems to be triggered when the WSL distribution list includes an empty string, causing topgrade to pass an incorrect distribution name to WSL.

---

### Erroneous Behavior

- Topgrade detects WSL distributions as ["Debian", ""] instead of filtering out an empty distribution name.
- This empty value is later used when trying to run commands (e.g., `which topgrade`), leading to an error.
- The primary update command fails within the Debian instance, and subsequent operations cause WSL to panic.

---

### Expected Behavior

- Topgrade should correctly list and filter available WSL distributions, excluding empty or invalid entries.
- The upgrade command should only invoke valid WSL distributions (for example, using "Debian") and succeed in executing the topgrade binary within WSL.
- WSL error messages related to an empty distribution should not be encountered.

---

### Steps to Reproduce

1. On a Windows system, install and configure WSL with at least one distribution (e.g., Debian).
2. Build topgrade via cargo.
3. Run topgrade in verbose mode from Windows (e.g., using PowerShell or CMD) such that it internally calls WSL, as shown by:  
   `.\target\debug\topgrade.exe --verbose`
4. Observe the verbose output where:
[...]

PR #1344, opened 2025-09-27
Fix "WSL already reported" panic

### LINKED ISSUE #1246: Panic "WSL already reported" due to duplicate step reporting
## Description
Topgrade panics with "WSL already reported" when both the WSL step (running topgrade inside WSL distributions) and the WSL Update step (updating WSL itself) attempt to report using the same identifier.

## Error Message
```
Message:  WSL already reported
Location: src\report.rs:37
```

## Root Cause
The debug logs show two separate WSL-related steps:
1. **WSL step**: Runs `topgrade` inside each WSL distribution 
2. **WSL Update step**: Updates WSL itself using `wsl --update`

Both steps appear to be using the same reporting key/identifier, causing a collision in the reporting system when the second step tries to report its results.

## Environment
- **OS**: Windows with WSL (Debian)
- **Topgrade version**: Built from main branch
- **Command**: `.\target\debug\topgrade.exe --verbose`

## Steps to Reproduce
1. Run Topgrade on Windows with WSL installed
2. Let both WSL steps execute (WSL distribution updates + WSL system update)
3. Observe panic after WSL update completes

## Debug Evidence
From the verbose output:
```
DEBUG Step "WSL"  # First WSL step (runs topgrade in distributions)
...
DEBUG Step "WSL"  # Second WSL step appears
── 02:28:15 - Update WSL ───  # WSL Update step runs
Checking for updates.
The most recent version of Windows Subsystem for Linux is already installed.
Message:  WSL already reported
```

## Expected Behavior
Both WSL-related steps should be able to report their results independently without conflicts.

## Actual Behavior
The second WSL-related step (WSL Update) cannot report because the first WSL step has already used the same reporting identifier.

## Suggested Fix
The two WSL steps need distinct reporting identifiers:
- WSL distributions step: "WSL" or "WSL Distributions"  
- WSL system update step: "WSL Update" or "WSL System"

## Code Location
- Panic occurs in `src\report.rs:37`
- WSL steps likely defined in `src\steps\os\windows.rs`

## Workaround
```toml
[misc]
disable = ["wsl", "wsl_update"]
```

## Additional Context
- This is a **regression** - the issue existed before, was fixed, but has reappeared
- The WSL functionality itself works correctly - only the reporting system has the bug
- Both steps complete their actual work successfully before the panic occurs

AThePeanut4:
## What does this PR do

Fix an assertion failure caused by two steps using "WSL" as the key, when using a debug build.

The assertion is here:
https://github.com/topgrade-rs/topgrade/blob/3961ef61c809b7476e818ac36f7a4a0a4691b1b5/src/report.rs#L37

While investigating I noticed that `Report` was basically just a thin wrapper around a `Vec`, which seemed unnecessary to me. So this also removes that entirely, storing the `Vec` directly in the `Runner` and moving `push_result()` to `Runner` (and removes the function's unnecessary use of `Option` and `Into<...>` as well).

## Standards checklist

Fixes #1246


GideonBear:
@niStee do #1246 or #1136 happen on this branch? Or does this fix both?

src/main.rs
```diff
@@ -6,7 +6,6 @@ use std::path::PathBuf;
 use std::process::exit;
 use std::time::Duration;
 
-use crate::breaking_changes::{first_run_of_major_release, print_breaking_changes, should_skip, write_keep_file};
```
GideonBear, line 9: Please move this back

src/step.rs
```diff
@@ -664,7 +664,7 @@ impl Step {
             WslUpdate =>
             {
                 #[cfg(windows)]
-                runner.execute(*self, "WSL", || windows::update_wsl(ctx))?
+                runner.execute(*self, "WSL Update", || windows::update_wsl(ctx))?
```
GideonBear: Should match separator
```suggestion
                runner.execute(*self, "Update WSL", || windows::update_wsl(ctx))?
```

GideonBear (changes requested):

niStee:
> @niStee do #1246 or #1136 happen on this branch? Or does this fix both?

I'll check this later.
I'm not at home at the moment.

atoerien:
I was able to reproduce #1246 (just run a debug build of Topgrade on Windows, with WSL installed), so it does fix that for me. #1136 looks completely unrelated, so I don't think this should fix it, but I haven't looked into it.

src/main.rs
atoerien: :(

src/step.rs
atoerien: Ah I forgot about the separator text. Will fix.

niStee:
> I was able to reproduce #1246 (just run a debug build of Topgrade on Windows, with WSL installed), so it does fix that for me. #1136 looks completely unrelated, so I don't think this should fix it, but I haven't looked into it.

I believe https://github.com/topgrade-rs/topgrade/issues/1136 must be fixed in https://github.com/topgrade-rs/topgrade/blob/fec08a5ad125a309fbb504c16dc06d89304ab9de/src/steps/os/windows.rs#L125-L132
like https://github.com/niStee/topgrade/blob/a7b315a2c04143c442b5413d1cb94c0f4cc41ef9/src/steps/os/windows/mod.rs#L131-L138

GideonBear:
Thanks! :heart:

atoerien:
> I believe #1136 must be fixed in
> 
> https://github.com/topgrade-rs/topgrade/blob/fec08a5ad125a309fbb504c16dc06d89304ab9de/src/steps/os/windows.rs#L125-L132
> 
> like https://github.com/niStee/topgrade/blob/a7b315a2c04143c442b5413d1cb94c0f4cc41ef9/src/steps/os/windows/mod.rs#L131-L138

@niStee I haven't been able to reproduce the bug, but if the change you link there fixes it then I think go ahead and make a PR for it. The only potential issue I can see is if you *intentionally* had a WSL distribution with an empty name, or if you had one with leading/trailing spaces in the name (`" foo"`) to cause issues with the `trim()`.

PR #918, opened 2024-09-27
feat(git): add support to only fetch repositories

Stef16Robbe:
## What does this PR do
add `fetch_only` to config in case user wants to only fetch, not already pull changes

## Standards checklist

- [x] The PR title is descriptive.
- [x] I have read `CONTRIBUTING.md`
- [x] *Optional:* I have tested the code myself
 
## For new steps

- [ ] *Optional:* Topgrade skips this step where needed
- [ ] *Optional:* The `--dry-run` option works with this step
- [ ] *Optional:* The `--yes` option works with this step if it is supported by 
  the underlying command

If you developed a feature or a bug fix for someone else and you do not have the
means to test it, please tag this person here.

PR #1371, opened 2025-10-21
feat(git): add option to fetch instead of pull repositories

### LINKED ISSUE #806: Git support fetch only
An option to only fetch git repos instead of pulling them would be appreciated. I do not want my git working directory changed by topgrade. But a git fetch would be useful so I can instantly pull in the future.

Stef16Robbe:
Continuation of https://github.com/topgrade-rs/topgrade/pull/918
Closes #806

New PR since I deleted my old fork :D. Also includes improvements based on original comments.

## What does this PR do

add fetch_only to config in case user wants to only fetch, not already pull changes

## Standards checklist

## For new steps

  the underlying command

### DIFF
--- a/src/steps/git.rs
+++ b/src/steps/git.rs
@@ -314,7 +314,7 @@ impl RepoStep {
         if is_fetch_only {
             command.args(["fetch", "--recurse-submodules"]);
         } else {
-            command.args(["pull", "--ff-only"]);
+            command.args(["pull", "--ff-only", "--recurse-submodules"]);
         }
 
         if let Some(extra_arguments) = ctx.config().git_arguments() {
@@ -322,19 +322,7 @@ impl RepoStep {
         }
 
         let output = command.output().await?;
-        let result = if is_fetch_only {
-            output_checked_utf8(output)
-        } else {
-            let submodule_output = AsyncCommand::new(&self.git)
-                .args(["submodule", "update", "--recursive"])
-                .current_dir(&repo)
-                .stdin(Stdio::null())
-                .output()
-                .await?;
-
-            output_checked_utf8(output).and_then(|()| output_checked_utf8(submodule_output))
-        }
-        .wrap_err_with(|| {
+        let result = output_checked_utf8(output).wrap_err_with(|| {
             let action = if is_fetch_only { "fetch" } else { "pull" };
             format!("Failed to {} {}", action, repo.as_ref().display())
         });


GideonBear:
I added "closes #806" to make sure GitHub understands that it closes that issue

locales/app.yml
```diff
@@ -56,6 +56,14 @@ _version: 2
...
   zh_TW: "正在拉取"
   de: "Abrufen"
+"Fetching":
+  en: "Fetching"
+  lt: "Gaunama"
+  es: "Obteniendo"
+  fr: "Récupération"
+  zh_CN: "正在获取"
+  zh_TW: "正在獲取"
+  de: "Abrufen"
```
GideonBear, line 82: Some of these overlap (fr, zh*, de). Is there some translated Git documentation that you can reference?

src/steps/git.rs
```diff
@@ -298,41 +298,61 @@ impl RepoStep {
...
-        let submodule_output = AsyncCommand::new(&self.git)
-            .args(["submodule", "update", "--recursive"])
-            .current_dir(&repo)
-            .stdin(Stdio::null())
-            .output()
-            .await?;
-        let result = output_checked_utf8(pull_output)
-            .and_then(|()| output_checked_utf8(submodule_output))
-            .wrap_err_with(|| format!("Failed to pull {}", repo.as_ref().display()));
+        let pull_or_fetch_output = command.output().await?;
```
GideonBear:
```suggestion
        let output = command.output().await?;
```

src/steps/git.rs
```diff
@@ -298,41 +298,61 @@ impl RepoStep {
...
+
+        let result = if is_fetch_only {
+            output_checked_utf8(pull_or_fetch_output)
+        } else {
```
GideonBear: Particular reason we don't want to `git submodule update --recursive` in `fetch_only`? Is there some kind of `git submodule fetch --recursive`?

src/steps/git.rs
```diff
@@ -366,16 +386,25 @@ impl RepoStep {
...
-            self.repos
-                .iter()
-                .for_each(|repo| println!("{}", t!("Would pull {repo}", repo = repo.display())));
+            self.repos.iter().for_each(|repo| {
+                println!(
+                    "{}",
+                    t!(
+                        "Would {action} {repo}",
+                        action = if is_fetch_only { "fetch" } else { "pull" },
+                        repo = repo.display()
```
GideonBear: Have you verified that this works? The key is `Would fetch {repo}` and `Would pull {repo}`, not `Would {action} {repo}`. I would expect something like this (although this probably doesn't work):

```suggestion
                    t!(
                        format!("Would {} {{repo}}", if is_fetch_only { "fetch" } else { "pull" }),
                        repo = repo.display()
```

GideonBear (changes requested):

config.example.toml
```diff
@@ -239,6 +239,8 @@
 # Arguments to pass Git when pulling Repositories
 # arguments = "--rebase --autostash"
 
+# Whether to perform a `git fetch` instead of `git pull`
+# fetch_only = false
```
GideonBear, line 255: As @SteveLauC said:
> BTW, for all the docs under `[git]`, we need to update them, `pull or fetch`

src/steps/git.rs
```diff
@@ -115,7 +115,7 @@ pub fn run_git_pull(ctx: &ExecutionContext) -> Result<()> {
 
     print_separator(t!("Git repositories"));
 
-    repos.pull_repos(ctx)
+    repos.pull_or_fetch_repos(ctx)
```
GideonBear, line 118: Another comment from @SteveLauC:
> This should be updated as well:
> 
> https://github.com/topgrade-rs/topgrade/blob/f6d2ba4daeb6d68b5896e3b700f0f47e4496bf55/src/steps/git.rs#L30

locales/app.yml
Stef16Robbe: I could have a look but this is just Claude translated

src/steps/git.rs
Stef16Robbe: It was suggested in the original PR. What do we think of `git submodule foreach --recursive git fetch`?

src/steps/git.rs
Stef16Robbe: Good catch, but i'm not too familiar with how that macro works/how to fix your suggestion, so I went with a "dumb" solution for now

src/steps/git.rs
GideonBear: How about just `git fetch --recurse-submodules`? Can we use `git pull --ff-only --recurse-submodules` as well, instead of a seperate `git submodule update --recursive` call? Or is there a difference?

src/steps/git.rs
GideonBear: Totally fine, thanks!

config.example.toml
```diff
@@ -224,21 +224,23 @@
...
 
-# Additional git repositories to pull
+# Additional git repositories to pull or fetch
 # repos = [
 #     "~/src/*/",
 #     "~/.config/something"
 # ]
 
-# Don't pull the predefined git repos
+# Don't pull/fetch the predefined git repos
```
GideonBear: Not your code, but this is a bit confusing
```suggestion
# Set to `false` to skip pulling/fetching the predefined git repos
```

GideonBear:
Aaaah GitHub review is confusing... Sorry if a comment is duplicated or out of place

GideonBear:
Hi, do you have any idea when you're able to work on this again?

src/steps/git.rs
Stef16Robbe: I agree, I just missed this flag on `fetch`.

```
       --recurse-submodules[=(yes|on-demand|no)]
           This option controls if and under what conditions new commits of submodules should be fetched too. When recursing through submodules, git fetch always
           attempts to fetch "changed" submodules, that is, a submodule that has commits that are referenced by a newly fetched superproject commit but are
           missing in the local submodule clone. A changed submodule can be fetched as long as it is present locally e.g. in $GIT_DIR/modules/ (see
           gitsubmodules(7)); if the upstream adds a new submodule, that submodule cannot be fetched until it is cloned e.g. by git submodule update.

           When set to on-demand, only changed submodules are fetched. When set to yes, all populated submodules are fetched and submodules that are both
           unpopulated and changed are fetched. When set to no, submodules are never fetched.

           When unspecified, this uses the value of fetch.recurseSubmodules if it is set (see git-config(1)), defaulting to on-demand if unset. When this option
           is used without any value, it defaults to yes.
```

I think the best option is to leave `yes|on-demand|no` unset so it uses the user's config.

locales/app.yml
Stef16Robbe: I can see in the French git docs it's the same, as for the other languages, I'm lost :D

Stef16Robbe:
I think we're almost there? Only weird thing I noticed is if I run `cargo build --release && ./target/release/topgrade -v --only git_repos`, the verbose logging does not properly print in colour, e.g.:

```
DEBUG Step "Git Repositories" failed:
   0: \x1b[91mFailed to pull xxx\x1b[0m
   1: \x1b[91mssh: Could not resolve hostname xxx: nodename nor servname provided, or not known
      fatal: Could not read from remote repository.

      Please make sure you have the correct access rights
      and the repository exists.\x1b[0m

Location:
   \x1b[35msrc/steps/git.rs\x1b[0m:\x1b[35m335\x1b[0m
```

locales/app.yml
GideonBear: Ah well, they don't have as good translations as I thought

src/steps/git.rs
```diff
@@ -298,41 +298,53 @@ impl RepoStep {
...
+        command.stdin(Stdio::null()).current_dir(&repo);
 
-        command
-            .stdin(Stdio::null())
-            .current_dir(&repo)
-            .args(["pull", "--ff-only"]);
+        if is_fetch_only {
+            command.args(["fetch", "--recurse-submodules"]);
+        } else {
+            command.args(["pull", "--ff-only"]);
```
GideonBear: Regarding my comment:
> Can we use `git pull --ff-only --recurse-submodules` as well, instead of a seperate `git submodule update --recursive` call? Or is there a difference?

For consistency and shorter code. The whole `if` block with the `"submodule", "update", "--recursive"` can be removed.
I looked it up, and there's only a small difference between the two, but for Topgrade's purposes using `--recurse-submodules` in the `pull`/`fetch` call is totally fine.
```suggestion
            command.args(["pull", "--ff-only", "--recurse-submodules"]);
```

GideonBear:
> I think we're almost there? Only weird thing I noticed is if I run `cargo build --release && ./target/release/topgrade -v --only git_repos`, the verbose logging does not properly print in colour, e.g.:

Can reproduce on `main` so not a problem caused by this PR. It's only in the debug output, it's just that that gets escaped before printing. Normal errors should be just fine.

src/steps/git.rs
Stef16Robbe: Completely read over that, fixed it!

GideonBear (approved):
Thanks! :heart:

PR #1412, opened 2025-11-01
ci: harden workflows with pinned actions, timeouts, and concurrency controls

niStee:
# What does this PR do?

Hardens existing CI and release workflows with security and reliability best practices.

## Changes:

**Security hardening:**
- Pin all actions to full commit SHAs instead of mutable tags
- Prevents supply-chain attacks per [OpenSSF Scorecard Pinned-Dependencies](https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies)

**Reliability improvements:**
- Add timeouts to all jobs (10min for fast jobs, 60-120min for builds)
- Add concurrency groups to release workflows with cancel-in-progress

**Maintenance improvements:**
- Replace manual `cross` curl install with `taiki-e/install-action` for easier Renovate updates

## Notes:

- Split from original scope per maintainer feedback - new workflows (CodeQL, cargo-deny) will be added separately


.github/workflows/ci.yml
```diff
@@ -130,36 +130,33 @@ jobs:
           - target: x86_64-pc-windows-msvc
             target_name: Windows
             os: windows-latest
-    env:
-      cargo_cmd: ${{ matrix.use_cross == true && 'cross' || 'cargo' }}
-      matrix_target: ${{ matrix.target }}
```
GideonBear, line 135: You're reverting changes here, possibly a bad merge

.github/workflows/ci.yml
```diff
@@ -21,11 +21,14 @@ jobs:
   fmt:
     name: Rustfmt
     runs-on: ubuntu-latest
+    timeout-minutes: 10
```
GideonBear, line 24: What benefit do timeouts give?

.github/workflows/ci.yml
```diff
@@ -21,11 +21,14 @@ jobs:
...
     steps:
       - name: Checkout code
+        # action pin via: git ls-remote https://github.com/actions/checkout refs/tags/v5.0.0 | head -1
```
GideonBear, line 27: What are these comments for?

.github/workflows/release_to_pypi.yml
```diff
@@ -33,13 +38,15 @@ jobs:
...
     runs-on: windows-latest
+    timeout-minutes: 60
     strategy:
       matrix:
         target: [x64, x86]
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
         with:
           persist-credentials: false
+          fetch-depth: 1
```
GideonBear: The default is 1 already, why add this?

.github/workflows/cargo-deny.yml
```diff
@@ -0,0 +1,43 @@
+name: cargo-deny
```
GideonBear: Adding new workflows does not belong in the same PR as hardening existing ones

GideonBear (changes requested):

.github/workflows/ci.yml
niStee: Timeouts prevent runaway jobs from consuming runner resources and exceeding GitHub Actions limits.

**From GitHub Actions documentation** ([jobs.<job_id>.timeout-minutes](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes)):
> "The maximum number of minutes to let a job run before GitHub automatically cancels it. **Default: 360**"

**Benefits:**
- **Resource efficiency**: Prevents jobs from hanging indefinitely if a command stalls, freeing up runner capacity for other projects
- **Fast failure detection**: A 10-minute timeout on `cargo fmt --check` that normally takes seconds will fail quickly if something goes wrong, rather than waiting the full 360-minute default
- **Explicit expectations**: Documents expected runtime, making performance regressions obvious
- **Responsible resource use**: Reduces unnecessary compute usage (better for the environment and GitHub's infrastructure)

For the `fmt` and `custom-checks` jobs, 10 minutes is generous (they typically take <1 minute), but provides a reasonable safety net while still failing much faster than the 6-hour default.

.github/workflows/ci.yml
niStee: The pin comments document the exact command used to obtain each SHA, improving auditability and supply-chain security.

**From GitHub Security Hardening guide** ([docs](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions)):
> "Pin actions to full-length commit SHA...Using the commit SHA of a released action version is **the safest for stability and security**."

**From OpenSSF Scorecard - Pinned-Dependencies check** ([docs](https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies)):
> "Pinned dependencies ensure that checking and deployment are all done with the same software, reducing deployment risks...Pinning also enables reproducibility."
> 
> **Risk Level: Medium** - Unpinned dependencies "can create many problems, including running code that is different from the reviewed code."

**The comments serve three purposes:**
1. **Auditability**: Anyone can verify the SHA matches the intended version
2. **Maintenance**: When updating, you know exactly how the pin was obtained (reproducible)
3. **Security**: Documents the provenance of each pinned dependency

Example: `# action pin via: git ls-remote https://github.com/actions/checkout refs/tags/v4.2.2 | head -1`

.github/workflows/release_to_pypi.yml
niStee: You're absolutely correct—`fetch-depth: 1` is currently the default for `actions/checkout`, and with pinned actions (which this PR implements), it's redundant.

I'll remove the explicit `fetch-depth: 1` from all workflows. Since we're pinning actions to specific SHAs, the default behavior won't change unexpectedly, making the explicit configuration unnecessary.

Thanks for catching this! 👍

.github/workflows/cargo-deny.yml
niStee: You're absolutely right. This PR mixes two distinct concerns:
1. **New workflows** (CodeQL, cargo-deny)
2. **Hardening existing workflows** (timeouts, pinned actions, concurrency groups)

I'll split this into two PRs:
- **PR #1412 (this one, revised):** Harden existing workflows only
- **PR (new):** Add CodeQL and cargo-deny workflows

This will make review easier and allow each set of changes to be discussed and merged independently. I'll update this PR shortly to remove the new workflow files.

GideonBear:
@niStee Are you using LLMs to write/translate/check your comments?
Your comments are extremely verbose and bloated, seeming to point to that.

niStee:
> @niStee Are you using LLMs to write/translate/check your comments?
> Your comments are extremely verbose and bloated, seeming to point to that.


Yes, I use them – for me it’s just a smarter spell‑checker; if the style bothers you, I can try to make future replies shorter and less formal.

GideonBear:
The style isn't the problem. It's the length and verbosity. Take this comment:

> You're absolutely right. This PR mixes two distinct concerns:
> 1. **New workflows** (CodeQL, cargo-deny)
> 2. **Hardening existing workflows** (timeouts, pinned actions, concurrency groups)
> 
> I'll split this into two PRs:
> - **PR #1412 (this one, revised):** Harden existing workflows only
> - **PR (new):** Add CodeQL and cargo-deny workflows
> 
> This will make review easier and allow each set of changes to be discussed and merged independently. I'll update this PR shortly to remove the new workflow files. 

I would've written this as:

> You're right, I'll split it up.

The information you're putting in the comment is mostly duplicated or noisy:

> You're absolutely right. This PR mixes two distinct concerns:
> 1. **New workflows** (CodeQL, cargo-deny)
> 2. **Hardening existing workflows** (timeouts, pinned actions, concurrency groups)

That's just what I said but in more words. What benefit do you think it gives to write it out like this?

> I'll split this into two PRs:
> - **PR #1412 (this one, revised):** Harden existing workflows only
> - **PR (new):** Add CodeQL and cargo-deny workflows

I can see you changing the title, and I can see you opening a new PR; link them if you so wish, but don't duplicate all this information.

> This will make review easier and allow each set of changes to be discussed and merged independently. I'll update this PR shortly to remove the new workflow files. 

You seem to understand this. I asked you to split it up, so I assume you know I understand it as well. Why would you need to clarify this?

It looks like you're not putting any effort into your comments, just letting LLMs generate them. Which is quite disrespectful; you're not putting the effort into writing an extensive comment, instead letting an LLM do that, but you expect me to read the entire thing? You're just wasting my time at that point.

(and if you believe it's good to have long verbose communication, perhaps even interpreting that verbose communication with an LLM, then I disagree, and I do not want that type of communication here)

I really don't mind long comments, but they have to be useful. What LLMs tend to do is generate text that looks great on the surface, but when you actually read it, turns out to be full of obvious information, duplication, and useless styling. Which (again) just wastes people's time! A great way to make sure your comment is actually useful is to **write it yourself**. You'll notice if what you're writing isn't useful information. And if you want to pass it through a spell-checker or an LLM then *fine* (even though I don't really care, I don't mind bad English at all), but don't post any text that you wouldn't have written yourself.

If I am interpreting this wrong, let me know. It's not my intention to "accuse" you of anything, but I feel like I have to be a bit harsh to keep out any bad faith. I cannot keep reviewing your stuff if you don't improve your communication.

PR #1423, opened 2025-11-02
ci: set dependabot cooldown

HastD:
This delays dependabot updates until 7 days after the dependency is updated, which is beneficial for stability and supply-chain security. See https://docs.zizmor.sh/audits/#dependabot-cooldown for more info.


GideonBear:
I am going to switch to Renovate soon, so I will close this for now. Dependabot is not updating cargo deps currently anyway.

PR #1433, opened 2025-11-04
fix(vim): change nvimrc base_dir for windows

### LINKED ISSUE #1091: Not finding nvimrc under windows with xdg variables
## Erroneous Behavior
When using `XDG_CONFIG_HOME` under Windows to change the location of the Neovim config dir, topgrade won't find the `init` file and therefore cannot run the Neovim step.

## Expected Behavior
Topgrade should either follow XDG variables when searching for the `init` file, or have a config option for Neovim to set the path to it manually.

## Steps to reproduce
Run topgrade with `XDG_CONFIG_HOME` defined and an `init.lua` file saved accordingly.

## Possible Cause (Optional)
Topgrade expects the `init` file to be placed at `~/AppData/Local/nvim`, because of using [`crate::WINDOWS_DIRS`](https://github.com/topgrade-rs/topgrade/blob/47fa3ba7de3c2c10dc3ac32fb1338d2c2e0d7c74/src/steps/vim.rs#L35).
Neovim respects XDG variables under Windows. Because of that, the file **might** be at `~/.config/nvim`.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

remote host
- [ ] Yes
- [ ] No

```toml

```

## Additional Details
- Operating System/Version
  Windows 10 22H2

- Installation
   scoop

   topgrade 16.0.2

<details>
<pre>

</pre>
</details>

SteveCoding125:
## What does this PR do

Closes #1091

Adds manual checking of the `XDG_CONFIG_HOME` variable when searching for nvim on windows.

Previously, the etcetera Windows strategy was used directly to get the config dir for nvim under Windows. Since XDG is not an official standard on Windows, the Windows strategy doesn't check these variables. Nvim on the other hand respects XDG even on Windows. Because of that, it can happen that the nvim config is not found under Windows.

## Standards checklist

## For new steps

  the underlying command

### DIFF
--- a/src/steps/vim.rs
+++ b/src/steps/vim.rs
@@ -29,15 +29,32 @@ pub fn vimrc() -> Result<PathBuf> {
 
 fn nvimrc() -> Result<PathBuf> {
     #[cfg(unix)]
-    let base_dir = crate::XDG_DIRS.config_dir();
+    let bases: Vec<PathBuf> = vec![crate::XDG_DIRS.config_dir()];
 
     #[cfg(windows)]
-    let base_dir = crate::WINDOWS_DIRS.cache_dir();
+    let mut bases: Vec<PathBuf> = vec![crate::WINDOWS_DIRS.cache_dir()];
 
-    base_dir
-        .join("nvim/init.vim")
-        .require()
-        .or_else(|_| base_dir.join("nvim/init.lua").require())
+    #[cfg(windows)]
+    {
+        if let Some(xdg) = std::env::var_os("XDG_CONFIG_HOME")
+            .map(PathBuf::from)
+            .filter(|path| path.is_absolute())
+        {
+            bases.insert(0, xdg);
+        }
+    }
+
+    for base_dir in bases {
+        if let Ok(p) = base_dir
+            .join("nvim/init.vim")
+            .require()
+            .or_else(|_| base_dir.join("nvim/init.lua").require())
+        {
+            return Ok(p);
+        }
+    }
+
+    Err(SkipStep(format!("{}", t!("No Neovim config found"))).into())
 }
 
 fn upgrade_script() -> Result<tempfile::NamedTempFile> {
--- a/locales/app.yml
+++ b/locales/app.yml
@@ -958,6 +958,14 @@ _version: 2
   zh_CN: "拉取 %{repo}"
   zh_TW: "拉取 %{repo}"
   de: "Würde %{repo} abrufen"
+"No Neovim config found":
+  en: "No Neovim config found"
+  lt: "Neovim konfigūracija nerasta"
+  es: "No se encontró ninguna configuración de Neovim"
+  fr: "Aucune configuration Neovim trouvée"
+  zh_CN: "未找到 Neovim 配置"
+  zh_TW: "未找到 Neovim 配置"
+  de: "Keine Neovim-Konfiguration gefunden"
 "Node Package Manager":
   en: "Node Package Manager"
   lt: "Node paketų tvarkyklė (npm)"


src/steps/vim.rs
```diff
@@ -32,7 +32,10 @@ fn nvimrc() -> Result<PathBuf> {
     let base_dir = crate::XDG_DIRS.config_dir();
 
     #[cfg(windows)]
-    let base_dir = crate::WINDOWS_DIRS.cache_dir();
+    let base_dir = std::env::var_os("XDG_CONFIG_HOME")
+        .map(PathBuf::from)
+        .filter(|path| path.is_absolute())
```
GideonBear: > Returns true if the Path is absolute, i.e., if it is independent of the current directory.

Why `PathBuf::is_absolute`? I would expect `path.exists()` here.
Please add a comment explaining it

SteveCoding125: The reason from my side was because this is the approach etcetera uses in the [XDG strategy](https://github.com/lunacookies/etcetera/blob/07b3f54999e4f3ff0f027de0144c9349d820ce80/src/base_strategy/xdg.rs#L175-L182).

If we use `exists` we would fall back to `WINDOWS_DIRS.cache_dir()` and have a higher likelihood of finding the file.

Thinking about it now, I think this would be a better approach for us. Do you want me to change this, or do you want parity with etcetera?

GideonBear: Uhhhhh...
Actually, can't we check for the existance of `base_dir.join(blah)` instead of just `base_dir` here? Why don't we just loop over the two possible base dirs and exit when any `init.vim` is found? I assume users don't have two?

SteveCoding125: Do you mean something like this

```rust
fn nvimrc() -> Result<PathBuf> {
    let mut bases: Vec<PathBuf> = Vec::new();

    #[cfg(unix)]
    bases.push(crate::XDG_DIRS.config_dir());

    #[cfg(windows)]
    {
        if let Some(xdg) = std::env::var_os("XDG_CONFIG_HOME")
            .map(PathBuf::from)
            .filter(|path| path.is_absolute() && path.exists())
        {
            bases.push(xdg);
        }

        bases.push(crate::WINDOWS_DIRS.cache_dir());
    }

    for base_dir in bases {
        if let Ok(p) = base_dir
            .join("nvim/init.vim")
            .require()
            .or_else(|_| base_dir.join("nvim/init.lua").require())
        {
            return Ok(p);
        }
    }

    Err(SkipStep(format!("{}", t!("No Neovim config found"))).into())
}
```

I am not sure about the default error, but this would be the general layout.

GideonBear: Yeah, something like that seems like a better idea, to not exclude the possibility of it being in `WINDOWS_DIRS.cache_dir()` just because `XDG_CONFIG_HOME` is set.

SteveCoding125: I have now implemented this version

src/steps/vim.rs
```diff
@@ -29,15 +29,32 @@ pub fn vimrc() -> Result<PathBuf> {
...
 
-    base_dir
-        .join("nvim/init.vim")
-        .require()
-        .or_else(|_| base_dir.join("nvim/init.lua").require())
+    #[cfg(windows)]
+    {
+        if let Some(xdg) = std::env::var_os("XDG_CONFIG_HOME")
+            .map(PathBuf::from)
+            .filter(|path| path.is_absolute() && path.exists())
```
GideonBear: Since we check `base_dir.join(...).require()` we don't need to explicitly check if it exists here
```suggestion
            .filter(|path| path.is_absolute())
```

GideonBear:
Just checking, have you verified the new code still works on your end?

SteveCoding125:
> Just checking, have you verified the new code still works on your end?

I have run `cargo build && cargo fmt && cargo clippy && cargo test` on Windows and Debian 12 WSL, I have tried running the compiled topgrade under Windows with:
- XDG set + file in xdg => runs update
- XDG set + file in cache_dir => runs update
- XDG set + file in both => runs update
- XDG set + file in neitehr => 'config not found'
- XDG not set + file in xdg => 'config not found'
- XDG not set + file in cache_dir => runs update
- XDG not set + file in both => runs update
- XDG not set + file in neither => 'config not found'

I think these are all combinations under windows. I haven't tried running under Linux, but that part should still work.

GideonBear:
Thanks for the PR and extensive testing!

PR #1434, opened 2025-11-04
ci: remove template expansion in code contexts

HastD:
Template expansion in code contexts is a potential source of shell injection vulnerabilities, as the expansion isn't syntax-aware. Inline template expansions in `run:` blocks can be replaced with environment variables containing the value. For more info, see: https://docs.zizmor.sh/audits/#template-injection

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,6 +13,10 @@ env:
 permissions:
   contents: read
 
+defaults:
+  run:
+    shell: bash
+
 jobs:
   fmt:
     name: Rustfmt


.github/workflows/release-plz.yml
```diff
@@ -33,12 +33,13 @@ jobs:
 
       - name: Trigger workflows
         if: steps.release-plz.outputs.releases_created == 'true'
-        run: |
-          gh api repos/${{ github.repository }}/dispatches \
-            -f "event_type=release-created" \
-            -F "client_payload[tag]=${{ fromJSON(steps.release-plz.outputs.releases)[0].tag }}"
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          client_payload_tag: ${{ fromJSON(steps.release-plz.outputs.releases)[0].tag }}
```
GideonBear:
```suggestion
          tag: ${{ fromJSON(steps.release-plz.outputs.releases)[0].tag }}
```

.github/workflows/create_release_assets.yml
```diff
@@ -26,6 +26,8 @@ jobs:
         # https://github.com/topgrade-rs/topgrade/issues/1095
         platform: [ ubuntu-22.04, macos-latest, macos-15-intel, windows-latest ]
     runs-on: ${{ matrix.platform }}
+    env:
+      event_tag: ${{ github.event.client_payload.tag }}
```
GideonBear: Here and all other `env: event_tag: ...` usages. Just a bit easier to read, unless you object?
```suggestion
      tag: ${{ github.event.client_payload.tag }}
```

GideonBear (changes requested):

.github/workflows/create_release_assets.yml
HastD: Good point, done.

GideonBear:
Thanks!

.github/workflows/create_release_assets.yml
```diff
@@ -252,23 +257,23 @@ jobs:
...
+          cross build --release --target "${matrix_target}"
+          cargo deb --target="${matrix_target}" --no-build --no-strip
         if: ${{ matrix.target == 'aarch64-unknown-linux-gnu' || matrix.target == 'armv7-unknown-linux-gnueabihf' }}
         shell: bash
 
       - name: Move Debian-based system package
         run: |
           mkdir -p assets
-          mv target/${{matrix.target}}/debian/*.deb assets
+          mv "target/${matrix_target}/debian/*.deb" assets
```
GideonBear, line 269: Aaagh we both missed this! The `*` is interpreted literally now. I will fix

.github/workflows/create_release_assets.yml
```diff
@@ -124,8 +126,8 @@ jobs:
         shell: bash
 
       - name: Upload assets
-        run:
-          gh release upload "${{ github.event.client_payload.tag }}" assets/*
+        run: |
+          gh release upload "${tag}" assets/*
```
GideonBear, line 130: This is also broken, but only on Windows (yay, powershell!).
Adding `defaults: run: shell: bash` in this file as well should fix it.

.github/workflows/create_release_assets.yml
HastD: Oops, yeah, definitely missed that. Thanks for fixing.

.github/workflows/create_release_assets.yml
HastD: Are you going to fix this as well in your PR?

GideonBear: Separate PR, but yes. You can review them both if you want.

PR #1444, opened 2025-11-08
docs: Add AI guidelines to CONTRIBUTING.md

GideonBear:


niStee:
Thanks @GideonBear for setting out these AI guidelines so clearly—I think the strict boundaries you’ve drawn make sense.

I also want to acknowledge your earlier critique of some of my PRs where I leaned too much on LLMs without proper review. That feedback was well‑founded, and it pushed me to create a `copilot-instructions.md` in my fork. The goal of that file is not to change your rules, but to make sure that when AI *is* used, it’s done responsibly and doesn’t compromise quality. For example:
- Propose a short plan before generating code, then keep diffs small and auditable.
- Verify every suggestion with `cargo test`, `cargo clippy`, and `cargo fmt`.
- Document when AI was used in commit messages.
- Avoid sensitive areas like credentials, privilege logic, or shell quoting entirely.
- Treat AI as a helper whose output must always be reviewed line‑by‑line.

GitHub provides documentation on this kind of setup: [Adding repository custom instructions for GitHub Copilot](https://docs.github.com/en/copilot/how-tos/configure-custom-instructions/add-repository-instructions?tool=visualstudio).

My thought is that Topgrade could optionally adopt a similar companion doc—not to expand AI use, but to reinforce responsible practices for contributors who do use it. That way your official guidelines stay strict and simple, while contributors who want more detail have a reference.

Happy to share the draft if you’d like to see how it looks.

GideonBear:
Thanks for the suggestion, but no, for a couple of reasons:
1. Having such a file at all makes the project look like it supports the use of AI explicitly, which I do not, and do not want to associate the project with.
2. It is not my intention to allow people to use LLM tools that attempt to grasp the whole codebase. I will try to clarify this in the contributing file as well.

Issue #933, opened 2024-10-07
feat: allow excluding encrypted file in chezmoi step


rubiin:
## I want to suggest a new step

* Which tool is this about? Where is its repository?
* Which operating systems are supported by this tool?
* What should Topgrade do to figure out if the tool needs to be invoked?
* Which exact commands should Topgrade run?
* Does it have a `--dry-run` option? i.e., print what should be done and exit
* Does it need the user to confirm the execution? And does it provide a `--yes`
  option to skip this step?

## I want to suggest some general feature
Topgrade should...

Allow an option to exclude the encrypted file 

## More information

More context can be found here https://github.com/twpayne/chezmoi/issues/3747

GideonBear:
Ref #831 
This is done by passing `--exclude=encrypted`

rubiin:
does this exist or a pr is needed ?

GideonBear:
> does this exist or a pr is needed ?

Sorry, I meant: "This should be implemented in Topgrade by passing `--exclude=encrypted` to Chezmoi".
A PR is needed.

rubiin:
@GideonBear sorry for the ping. Added a pr could you check this quick

PR #1453, opened 2025-11-10
feat(chezmoi): add `exclude_encrypted` config

rubiin:
## What does this PR do

Fixes https://github.com/topgrade-rs/topgrade/issues/933

## Standards checklist

## For new steps

  the underlying command

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -288,6 +288,7 @@
 
 [chezmoi]
 # Exclude encrypted files from update
+# (default: false)
 # exclude_encrypted = false
 
 [npm]


GideonBear:
To not break existing workflows, this needs to be a configuration option.
```toml
[chezmoi]
# exclude_encrypted = true
```
That would make this a `feat` as well

GideonBear:
(unless you have a good reason why this should/can be enabled for all users)

rubiin:
i will try to implement. I am kind of new to rust and this codebase

rubiin:
@GideonBear  have a look now

src/steps/generic.rs
```diff
@@ -916,9 +916,17 @@ pub fn run_chezmoi_update(ctx: &ExecutionContext) -> Result<()> {
     let chezmoi = require("chezmoi")?;
     HOME_DIR.join(".local/share/chezmoi").require()?;
 
+    let mut executor = ctx.execute(chezmoi);
```
GideonBear: Nit, `executor` is not the best name for it
```suggestion
    let mut cmd = ctx.execute(chezmoi);
```

GideonBear (changes requested):
Looks good!
Please also add an entry to `config.example.toml`, containing a description and the default.

config.example.toml
```diff
@@ -286,6 +286,10 @@
 # winget_use_sudo = true
 
 
+[chezmoi]
+# Exclude encrypted files from update
+# exclude_encrypted = false
```
GideonBear, line 292:
```suggestion
# Exclude encrypted files from update
# (default: false)
# exclude_encrypted = false
```

PR #1473, opened 2025-11-11
fix(openbsd): fix compilation on OpenBSD

Izder456:
## What does this PR do

Fixes compile time error due to incomplete migration to `execution_context` for `RunType`

Before:
```
~/Projects/Misc/topgrade git:(main) $ cargo build 
   Compiling topgrade v16.2.1 (/home/izder456/Projects/Misc/topgrade)
error[E0432]: unresolved import `crate::executor::RunType`
 --> src/steps/os/openbsd.rs:3:5
  |
3 | use crate::executor::RunType;
  |     ^^^^^^^^^^^^^^^^^^^^^^^^ no `RunType` in `executor`
  |
help: consider importing this enum instead
  |
3 - use crate::executor::RunType;
3 + use crate::execution_context::RunType;
  |

error[E0616]: field `config` of struct `ExecutionContext` is private
  --> src/steps/os/openbsd.rs:12:15
   |
12 |     match ctx.config.run_type() {
   |               ^^^^^^ private field
   |
help: a method `config` also exists, call it with parentheses
   |
12 |     match ctx.config().run_type() {
   |                     ++

error[E0599]: no method named `dry_run` found for reference `&Config` in the current scope
  --> src/steps/os/openbsd.rs:28:21
   |
28 |     if ctx.config().dry_run() {
   |                     ^^^^^^^ method not found in `&Config`

Some errors have detailed explanations: E0432, E0599, E0616.
For more information about an error, try `rustc --explain E0432`.
error: could not compile `topgrade` (bin "topgrade") due to 3 previous error
```

## Standards checklist

## For new steps

  the underlying command

### DIFF
--- a/src/steps/os/openbsd.rs
+++ b/src/steps/os/openbsd.rs
@@ -1,10 +1,10 @@
 use crate::command::CommandExt;
-use tracing::debug;
 use crate::execution_context::ExecutionContext;
 use crate::terminal::print_separator;
 use color_eyre::eyre::Result;
 use rust_i18n::t;
 use std::fs;
+use tracing::debug;
 
 fn is_openbsd_current() -> Result<bool> {
     let motd_content = fs::read_to_string("/etc/motd")?;


src/steps/os/openbsd.rs
```diff
@@ -25,9 +25,12 @@ pub fn upgrade_openbsd(ctx: &ExecutionContext) -> Result<()> {
...
-    if ctx.config().dry_run() {
-        println!("{}", t!("Would upgrade the OpenBSD system"));
-        return Ok(());
+    match ctx.config().run_type() {
+        RunType::Dry | RunType::Damp => {
+            println!("{}", t!("Would upgrade the OpenBSD system"));
+            return Ok(());
+        }
+        RunType::Wet => {}
     }
```
GideonBear: Since `sudo.execute` handles the dry running behavior just fine, we can remove this part entirely

```suggestion
```

GideonBear (changes requested):

GideonBear:
Looks like we don't have any openbsd tests. Maybe something to look at.

GideonBear:
@AThePeanut4 Could you review this?

src/steps/os/openbsd.rs
GideonBear: (and also that section in the locale file then)

src/steps/os/openbsd.rs
```diff
@@ -9,7 +9,7 @@ use std::fs;
 fn is_openbsd_current(ctx: &ExecutionContext) -> Result<bool> {
     let motd_content = fs::read_to_string("/etc/motd")?;
     let is_current = ["-current", "-beta"].iter().any(|&s| motd_content.contains(s));
-    match ctx.config.run_type() {
+    match ctx.config().run_type() {
```
atoerien: Do we really need this? It's basically just a dry-run only `println!`.
imo the user doesn't really care whether topgrade is checking `/etc/motd`, a debug message would be fine:
```rust
debug!("OpenBSD is -current/-beta: {is_current}");
```

Otherwise if we do keep the message, this should be `ctx.run_type()` not `ctx.config().run_type()`:
```suggestion
    match ctx.run_type() {
```

src/steps/os/openbsd.rs
atoerien: As before, this should use `ctx.run_type()` instead of `ctx.config().run_type()`:
```suggestion
    match ctx.run_type() {
        RunType::Dry | RunType::Damp => {
            println!("{}", t!("Would upgrade the OpenBSD system"));
            return Ok(());
        }
        RunType::Wet => {}
    }
```
But I do agree that this section can be removed entirely since `.execute()` deals with dry-running just fine.

Izder456:
I've resolved the suggestions in the Code Review. Good to merge?

src/steps/os/openbsd.rs
```diff
@@ -9,9 +10,9 @@ use std::fs;
...
     let motd_content = fs::read_to_string("/etc/motd")?;
     let is_current = ["-current", "-beta"].iter().any(|&s| motd_content.contains(s));
-    match ctx.config.run_type() {
+    match ctx.run_type() {
         RunType::Dry | RunType::Damp => {
-            println!("{}", t!("Checking if /etc/motd contains -current or -beta"));
+            debug!("OpenBSD is -current/-beta: {is_current}");
         }
         RunType::Wet => {}
     }
```
GideonBear: Nit, but there's not really a reason for this to be conditional anymore
```suggestion
    debug!("OpenBSD is -current/-beta: {is_current}");
```

src/steps/os/openbsd.rs
```diff
@@ -9,9 +10,9 @@ use std::fs;
 fn is_openbsd_current(ctx: &ExecutionContext) -> Result<bool> {
     let motd_content = fs::read_to_string("/etc/motd")?;
     let is_current = ["-current", "-beta"].iter().any(|&s| motd_content.contains(s));
-    match ctx.config.run_type() {
+    match ctx.run_type() {
         RunType::Dry | RunType::Damp => {
-            println!("{}", t!("Checking if /etc/motd contains -current or -beta"));
```
GideonBear, line 14: Needs to be removed from the locale file

src/steps/os/openbsd.rs
```diff
@@ -25,11 +26,6 @@ pub fn upgrade_openbsd(ctx: &ExecutionContext) -> Result<()> {
 
     let is_current = is_openbsd_current(ctx)?;
 
-    if ctx.config().dry_run() {
-        println!("{}", t!("Would upgrade the OpenBSD system"));
```
GideonBear, line 29: Needs to be removed from the locale file

GideonBear:
> I've resolved the suggestions in the Code Review. Good to merge?

Almost, thanks! :heart:

src/steps/os/openbsd.rs
Izder456: I don't see this in the locale file.

Izder456:
> Looks like we don't have any openbsd tests. Maybe something to look at.

after a cursory glance on you guys' CI infra, I see it uses the cross compilation mechanisms from rust. but i'm pretty sure OpenBSD isn't a high enough tier of a platform for this so I don't know how it would work. ofc, barring a qemu vm like in [vm-actions](https://github.com/vmactions/openbsd-vm)

src/steps/os/openbsd.rs
GideonBear: Okay wow, then that has always been invalid :facepalm:

src/steps/os/openbsd.rs
```diff
@@ -1,20 +1,17 @@
 use crate::command::CommandExt;
+use crate::debug;
```
GideonBear:
```suggestion
use tracing::debug;
```

GideonBear (approved):

GideonBear:
> > Looks like we don't have any openbsd tests. Maybe something to look at.
> 
> after a cursory glance on you guys' CI infra, I see it uses the cross compilation mechanisms from rust. but i'm pretty sure OpenBSD isn't a high enough tier of a platform for this so I don't know how it would work. ofc, barring a qemu vm like in [vm-actions](https://github.com/vmactions/openbsd-vm?rgh-link-date=2025-11-12T17%3A29%3A08Z)

Thanks, then I'll wave it off for now, given the relatively small part of the codebase that is OpenBSD-only.

PR #1518, opened 2025-11-16
chore(license): switch license variant to GPL-3.0-or-later

### LINKED ISSUE #847: Specify GPL license variant as `GPL-3.0-or-later`
As noted in SPDX GPL3 license documentation, the license text doesn't typically differ between the `-only` and `-or-later` variants, so I'm thinking it would be helpful to users to add a short comment at the bottom of README about which variant we're using (I'm assuming `-or-later` as it's the more convenient one).

tranzystorekk:
## What does this PR do

Closes #847

Changes the license field in `Cargo.toml` to the correct specifier. A quick grep didn't reveal any other places where "license" or "GPL" is mentioned.

We could possibly add a short license info to README to make it more discoverable and clear.

## Standards checklist

## For new steps

  the underlying command


GideonBear (approved):
Thanks!

Issue #1522, opened 2025-11-17
Perform housekeeping for docker


jstelzer:
## Checklist

- [ X] I have searched the issue tracker for relevant or duplicate issues.

## I want to suggest some general feature
Is it worth having an opt-in "docker system prune" be part of the topgrade run? Personally, I find it useful as I actively build docker images and my cache gets bloated fast.

## More information

- [x ] I am able and willing to implement this feature myself

I have a POC working. Cleaning up the code. Rust isn't my primary language so if you're willing to review it I'm willing to share my work.

Right now I have it wired up as opt in:

```
[containers]
system_prune = true
```

The default is false.

GideonBear:
Have you seen that we already run `docker image prune -f` if `cleanup = true` is set? I wonder if it really needs to be opt-in like that then. Is running `docker system prune -f` less safe than `docker image prune`, or in other words, is there a situation in which people could be mad that we're running `docker system prune -f`, assuming that they're okay with running `docker image prune -f`?

Of course I can review your code, feel free to open a PR!

jstelzer:
I did see that. I made it opt in for lack of direction and it can be unwanted. For example, you may not want those dangling volumes pruned because you plan on bringing up a docker-compose stack that uses that saved state. I didn't want to "accidentally all the things" and surprise users. Semantically, what I want to enable is slightly expanded in scope to what is there:

```
Difference Between docker image prune -f and docker system prune

docker image prune -f — Removes only unused images
	•	Deletes dangling images (untagged layers with no references).
	•	Can optionally remove all unused images with --all, but still only touches images.


Bottom line: Image prune is scope-limited. It frees space from unreferenced image layers but leaves the Docker build cache intact—often the biggest disk hog on dev machines.

⸻

docker system prune — Full cleanup of unused Docker objects
	•	Cleans everything not actively in use:
	•	Unused images (including dangling)
	•	Stopped containers
	•	Unused networks
	•	Build cache (if run with --volumes or --all depending on version)
	•	In practice, this is where the huge space savings come from, especially:
	•	Buildkit build caches
	•	Layer caches from repeated builds
	•	Intermediate build artifacts

Bottom line: System prune is holistic cleanup, including the build cache that image prune leaves behind. This is why it frees far more space—often gigabytes.

```

GideonBear:
Sure, it's probably a good idea to keep it opt-in, at least for now

jstelzer:
[...]

PR #1523, opened 2025-11-17
feat(containers): add option to run `system prune`

jstelzer:
## What does this PR do

This PR lets you opt in to having docker cleaned up aggressively.

## Standards checklist

- [ x] The PR title is descriptive
- [ x] I have read `CONTRIBUTING.md`
- [ x] *Optional:* I have tested the code myself
- [ -] If this PR introduces new user-facing messages they are translated

Closes https://github.com/topgrade-rs/topgrade/issues/1522

### DIFF
--- a/src/steps/containers.rs
+++ b/src/steps/containers.rs
@@ -230,7 +230,8 @@ pub fn run_containers(ctx: &ExecutionContext) -> Result<()> {
             error!("Running system prune failed: {}", e);
             success = false;
         }
-    } else if ctx.config().cleanup() { // Only run `image prune` if we don't run `system prune`
+    // Only run `image prune` if we don't run `system prune`
+    } else if ctx.config().cleanup() {
         // Remove dangling images
         debug!("Removing dangling images");
         if let Err(e) = ctx.execute(&crt).args(["image", "prune", "-f"]).status_checked() {


GideonBear:
I've edited your description to include the keyword "closing", which indicates to GitHub that it closes that issue

src/steps/containers.rs
```diff
@@ -233,6 +233,14 @@ pub fn run_containers(ctx: &ExecutionContext) -> Result<()> {
         }
     }
 
+    if ctx.config().containers_system_prune() {
+        // Run system prune to clean up unused containers, networks, and build cache
+        if let Err(e) = ctx.execute(&crt).args(["system", "prune", "-f"]).status_checked() {
```
GideonBear: Nit: since we're not typing this by hand, we can expand it for clarity
```suggestion
        if let Err(e) = ctx.execute(&crt).args(["system", "prune", "--force"]).status_checked() {
```

GideonBear: Do we want to include `-a, --all Remove all unused images not just dangling ones` as well?

src/steps/containers.rs
```diff
@@ -233,6 +233,14 @@ pub fn run_containers(ctx: &ExecutionContext) -> Result<()> {
...
+            error!("Running system prune failed: {}", e);
+            success = false;
```
GideonBear, line 231: I thought you were doing something dumb here but it was already dumb. I will improve this later :/

src/steps/containers.rs
```diff
@@ -233,6 +233,14 @@ pub fn run_containers(ctx: &ExecutionContext) -> Result<()> {
         }
     }
```
GideonBear: This can be moved to be an `else if` block on `if ctx.config().containers_system_prune()`, since `image prune` is a subset of `system prune`; we don't need to run it if we run `system prune` already

config.example.toml
```diff
@@ -341,6 +341,9 @@
 # ignored_containers = ["ghcr.io/rancher-sandbox/rancher-desktop/rdx-proxy:latest", "docker.io*"]
 # Specify the runtime to use for containers (default: "docker", allowed values: "docker", "podman")
 # runtime = "podman"
+# Run 'docker system prune' to clean up unused containers, networks, and build cache
```
GideonBear, line 345:
```suggestion

# Run 'docker system prune' to clean up unused containers, networks, and build cache
```

GideonBear (changes requested):

src/steps/containers.rs
jstelzer: Personally, I often have images present that i'm not using "right now" but I will be launching again. Pulling them back down to relaunch is kind of a waste of bandwidth. IMO.

When I wrote this, I was in the frame of mind of "weekly updates should take care of basic house keeping".

Personally I would rather use docker directly when i'm making decisions about what images to keep and what to remove simply because there wasn't a compose stack running at that moment. But, I'm just one voice.

src/steps/containers.rs
jstelzer: Lol, I  had to read this twice before it clicked... but i see it now. Sec.

src/steps/containers.rs
```diff
@@ -224,7 +224,13 @@ pub fn run_containers(ctx: &ExecutionContext) -> Result<()> {
...
 
-    if ctx.config().cleanup() {
+    if ctx.config().containers_system_prune() {
+        // Run system prune to clean up unused containers, networks, and build cache
+        if let Err(e) = ctx.execute(&crt).args(["system", "prune", "--force"]).status_checked() {
+            error!("Running system prune failed: {}", e);
+            success = false;
+        }
+    } else if ctx.config().cleanup() {
```
GideonBear, line 234:
```suggestion
    } else if ctx.config().cleanup() {  // Only run `image prune` if we don't run `system prune`
```

GideonBear (approved):

src/steps/containers.rs
```diff
@@ -224,7 +224,13 @@ pub fn run_containers(ctx: &ExecutionContext) -> Result<()> {
...
     }
 
-    if ctx.config().cleanup() {
+    if ctx.config().containers_system_prune() {
+        // Run system prune to clean up unused containers, networks, and build cache
+        if let Err(e) = ctx.execute(&crt).args(["system", "prune", "--force"]).status_checked() {
+            error!("Running system prune failed: {}", e);
+            success = false;
+        }
+    } else if ctx.config().cleanup() {  // Only run `image prune` if we don't run `system prune`
```
GideonBear:
```suggestion
    } else if ctx.config().cleanup() { // Only run `image prune` if we don't run `system prune`
```

GideonBear:
Thanks!

PR #1182, opened 2025-06-16
Preserve custom commmand order from config instead of sorting

GideonBear:
(alphabetically)

## What does this PR do
Closes #1181
Technically this is a breaking change.
```console
$ cargo run -- --only custom_commands
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.60s
     Running `target/debug/topgrade --only custom_commands`

── 13:55:04 - B command ────────────────────────────────────────────────────────
This should be first

── 13:55:04 - A command ────────────────────────────────────────────────────────
This should be second

── 13:55:04 - Summary ──────────────────────────────────────────────────────────
B command: OK
A command: OK
```

## Standards checklist

- [x] The PR title is descriptive
- [x] I have read `CONTRIBUTING.md`
- [x] *Optional:* I have tested the code myself
- [ ] If this PR introduces new user-facing messages they are translated

Issue #1181, opened 2025-06-16
Custom commands should not be sorted alphabetically, but preserve their order from the config file


GideonBear:
## Erroneous Behavior

Custom commands are executed in alphabetical order

## Expected Behavior

Custom commands are executed in the order they're defined in the `[commands]` section of the config file.

## Steps to reproduce

```toml
[commands]
"B command" = "echo This should be first"
"A command" = "echo This should be second"
```

## Possible Cause (Optional)

Does the toml parser do this automatically?

## Workarounds
Use `[pre_commands]` and/or `[post_commands]`
Prefix names with numbers (e.g. `"10 My command"`)

GideonBear:
Looking into this. They're stored in a `BTreeMap` which should be ordered.

SteveLauC:
> Custom commands should not be sorted alphabetically, but preserve their order from the config file 

I agree

GideonBear:
I created an issue (https://github.com/toml-rs/toml/issues/954), because I can reproduce this with just `toml`

GideonBear:
Turns out I didn't realize the meaning of a B-tree. `BTreeMap` is always sorted; we should change it to a `IndexMap`. I'll PR

PR #1535, opened 2025-11-19
fix(deps): restore custom commands order

louiswpf:
## What does this PR do
The toml update to 0.9.8 (in commit 9ec8e83f410f) broke custom command ordering because toml 0.9.0+ don't preserve order by default.

Quoting from toml 0.9.0 changelog [1]:
> from_str, Deserializer, etc no longer preserve order, requiring the
> preserve_order feature like Table

Enable preserve_order feature for toml to restore the ordering.

[1] https://github.com/toml-rs/toml/blob/toml-v0.9.8/crates/toml/CHANGELOG.md#compatibility-1

## Standards checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ readme = "README.md"
 home = "=0.5.11"
 etcetera = "=0.10.0"
 serde = { version = "~1.0", features = ["derive"] }
-toml = "=0.9.8"
+toml = { version = "=0.9.8", features = ["preserve_order"] }
 which_crate = { version = "~8.0", package = "which" }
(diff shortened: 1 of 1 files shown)


GideonBear:
Thanks, good spot! Sorry for not reading the changelogs I guess...
For future reference: we need **both** the `preserve_order` feature, and an `IndexMap` for it to be ordered. `preserve_env` is for `toml`s internal representation and `toml`s non-serde `toml::Value::Table` to be an `IndexMap`, but only if we use an `IndexMap` as well, our result is ordered.

Ref #1181 #1182 https://github.com/toml-rs/toml/issues/954

PR #1548, opened 2025-11-20
feat(mise): add support for parallel job configuration in mise

rubiin:
## What does this PR do

adds the jobs flag
https://mise.jdx.dev/cli/upgrade.html#flags

## Standards checklist

## For new steps

  the underlying command

### DIFF
--- a/src/steps/os/unix.rs
+++ b/src/steps/os/unix.rs
@@ -81,7 +81,6 @@ impl BrewVariant {
     /// Execute an "internal" brew command, i.e. one that should always be run
     /// even when dry-running. Basically just a wrapper around [`Command::new`]
     /// that uses `arch` to run using the correct architecture if needed.
-    #[cfg(target_os = "macos")]
     fn execute_internal(self) -> Command {
         match self {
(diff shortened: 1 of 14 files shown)


src/steps/os/unix.rs
```diff
@@ -836,6 +836,11 @@ pub fn run_mise(ctx: &ExecutionContext) -> Result<()> {
         cmd.arg("--bump");
     }
 
+    if ctx.config().mise_jobs() != 4 {
+        cmd.arg("--jobs ");
+        cmd.arg(ctx.config().mise_jobs().to_string());
```
GideonBear: The arg should not have a trailing space; also can be written as
```suggestion
        cmd.args(["--jobs"], ctx.config().mise_jobs().to_string());
```

src/config.rs
```diff
@@ -179,6 +179,7 @@ pub struct Chezmoi {
 pub struct Mise {
     bump: Option<bool>,
     interactive: Option<bool>,
+    jobs: Option<i32>,
```
GideonBear: I assume negative jobs aren't allowed
```suggestion
    jobs: Option<u32>,
```

PR #1556, opened 2025-11-24
fix(auto-cpufreq): fix skipping on systems with merged bin/sbin

PhoenixCausesOof:
Because of https://github.com/topgrade-rs/topgrade/pull/1215, auto-cpufreq is only updated if the path reported by the `which` crate is "/usr/local/bin/auto-cpufreq". In my system, however, the which crate gives "/usr/local/**s**bin/auto-cpufreq" (notice "bin" vs "sbin") and because of that, it assumes I did not install 'auto-cpufreq' with the official installer, which I have.

The fix was to create an exception for the "sbin" path.

## Standards checklist

## For new steps

  the underlying command

### DIFF
--- a/src/steps/os/linux.rs
+++ b/src/steps/os/linux.rs
@@ -1123,11 +1123,9 @@ pub fn run_waydroid(ctx: &ExecutionContext) -> Result<()> {
 }
 
 pub fn run_auto_cpufreq(ctx: &ExecutionContext) -> Result<()> {
-    let auto_cpu_freq = require("auto-cpufreq")?;
+    let auto_cpu_freq = require("auto-cpufreq")?.canonicalize()?;
 
-    if auto_cpu_freq != PathBuf::from("/usr/local/bin/auto-cpufreq")
-        && auto_cpu_freq != PathBuf::from("/usr/local/sbin/auto-cpufreq")
-    {
+    if auto_cpu_freq != PathBuf::from("/usr/local/bin/auto-cpufreq") {
         return Err(SkipStep(String::from(
             "`auto-cpufreq` was not installed by the official installer, but presumably by a package manager.",
         ))


GideonBear:
Hi @PhoenixCausesOof, the `auto-cpufreq-installer` script has this line:
https://github.com/AdnanHodzic/auto-cpufreq/blob/master/auto-cpufreq-installer#L16
I cannot find any reference to `sbin` either. How did you install `auto-cpufreq`? How did it end up in `sbin`? Is there some `sbin`/`bin` unifying symlinking going on?

PhoenixCausesOof:
> Hi @PhoenixCausesOof, the `auto-cpufreq-installer` script has this line: https://github.com/AdnanHodzic/auto-cpufreq/blob/master/auto-cpufreq-installer#L16 I cannot find any reference to `sbin` either. How did you install `auto-cpufreq`? How did it end up in `sbin`? Is there some `sbin`/`bin` unifying symlinking going on?

Huh. You're right. `sbin` was actually a symlink to `bin` (I even managed to find [this](https://fedoraproject.org/wiki/Changes/Unify_bin_and_sbin), and my system is Fedora.). I had assumed the installer used `sbin` because the files' modified date aligned with the use of the installer, but I should've looked deeper. I don't think I can work around this issue without maybe deleting the symlink, although there may be another way that I fail to see. 

Thanks for your patience.

GideonBear:
I think the issue is more "why does `which` detect the path as being in `sbin` and does not resolve the symlink" then...
What is your `echo $PATH`?

PhoenixCausesOof:
```
$ echo $PATH
... /usr/local/sbin /usr/local/bin /usr/sbin /usr/bin ...
```

This has also been addressed by the proposal[^1], but because my system was upgraded to, as opposed to installing, Fedora 42, `PATH` remained untouched, which I presume was to reduce breakage (as part of the transitioning process).

[^1]: https://fedoraproject.org/wiki/Changes/Unify_bin_and_sbin#Upgrade.2Fcompatibility_impact

GideonBear:
Unix `which` also gives the first on the path, even if that's a symlink. I will merge this as temporary fix and open an issue on the `which` crate to see if they can add a feature to skip symlinks. Thanks for the research!

GideonBear:
Nevermind I forgot I can just canonicalize symlinks myself!

GideonBear:
@PhoenixCausesOof Could you test that this works as well?

PhoenixCausesOof:
@GideonBear just tested here. Worked fine for me 👍

PR #1576, opened 2025-11-29
fix: use measure_text_width instead of byte count for header border calculation

jtbrough:
## What does this PR do

The `print_separator` method was using `message.len()` which returns the byte count of the string. This caused incorrect border calculations when the message contains multi-byte UTF-8 characters.

Using `message.chars().count()` properly counts Unicode characters for accurate 80-character header formatting regardless of character encoding.

**Before:**
image

**After:**
image

## Standards checklist

- [✅] The PR title is descriptive
- [✅] I have read `CONTRIBUTING.md`
- [✅] *Optional:* I have tested the code myself
- [n/a] If this PR introduces new user-facing messages they are translated

### DIFF
--- a/src/terminal.rs
+++ b/src/terminal.rs
@@ -8,7 +8,7 @@ use std::time::Duration;
 use chrono::{Local, Timelike};
 use color_eyre::eyre;
 use color_eyre::eyre::Context;
-use console::{style, Key, Term};
+use console::{measure_text_width, style, Key, Term};
 use notify_rust::{Notification, Timeout};
 use rust_i18n::t;
 use tracing::{debug, error};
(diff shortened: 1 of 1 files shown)


GideonBear:
From my testing:
## `main`
image

both recognized as one character too many, causing one character too little padding
## This PR applied
image

* y̆ unchanged? Or correct? Or partly corrected? I can't see and am not gonna rerun it sry; you can probably do the same test
* 😂 recognized as one character too few; I think my terminal prints an emoji as two-wide?

---
> Using `message.chars().count()` properly counts Unicode characters

Incorrect, to quote the docs:

> It’s important to remember that [char](https://doc.rust-lang.org/std/primitive.char.html) represents a Unicode Scalar Value, and might not match your idea of what a ‘character’ is. Iteration over grapheme clusters may be what you actually want. This functionality is not provided by Rust’s standard library, check crates.io instead.

I agree our current `.len()` is bugged, but this doesn't solve everything. Is there some library on crates.io that does terminal-width stuff like this?

jtbrough:
@GideonBear good catch. I think using [console::measure_text_width()](https://docs.rs/console/latest/console/fn.measure_text_width.html) will be a more universally reliable fix here since it will handle multi-space characters (emojis, CJK, ansi codes, etc.) based on their display width.

GideonBear (approved):
Seems to work, thanks!

PR #1586, opened 2025-12-03
fix(archlinux): don't overwrite PATH

mati865:
Overwriting PATH with just `/usr/bin` disregards user settings regarding preferred commands. For example, when sudo-rs is symlinked as `/usr/local/bin/sudo`, some of the commands will still call `/usr/bin/sudo` because of the modified PATH.

This leads to double password prompt: one for sudo and another one for sudo-rs.

---

## What does this PR do

Removes PATH environment variable modification from Arch Linux system update handler. AFAICT, this was the only place where such modification was performed.

## Standards checklist

I have daily driven this change for over two weeks using Paru and the original Pacman, I didn't test other Pacman helpers. 

## For new steps

  the underlying command

### DIFF
--- a/src/steps/os/archlinux.rs
+++ b/src/steps/os/archlinux.rs
@@ -1,5 +1,3 @@
-use std::env::var_os;
-use std::ffi::OsString;
 use std::path::{Path, PathBuf};
 
 use color_eyre::eyre;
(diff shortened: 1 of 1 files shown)


GideonBear:
FFR: The changes in the envvar do not have any effect on the exe (like `yay`) itself, since that path is resolved by `which` (via `require`) before execution.

This could potentially cause breakage; there might be a reason it was added. However there's no comment or explanation in the commit that added it (https://github.com/topgrade-rs/topgrade/commit/fb18af12ac07cb3056dc49276172a06d91ea7e1c, https://github.com/r-darwish/topgrade/pull/776, https://github.com/r-darwish/topgrade/issues/765), so let's remove it since there's a clear issue with it (your `sudo`). Thanks for the PR!

PR #1589, opened 2025-12-05
feat(sudo): propagate --env to sudo commands (#1588)

### LINKED ISSUE #1588: Propagate `--env` to `sudo` commands
Discussed in #1587

## Checklist

`SudoExecuteOps` should put environment variables passed with `--env` in `preserve_env_list` by default. See discussion in #1587.

w568w:
## What does this PR do

1.  Extracts environment variables set via `--env` from `ExecutionContext` within `SudoExecuteOpts::new` to serve as defaults.
2.  Adds tests.
3.  Modifies `preserve_env_list` to `extend_preserve_env_list`, allowing new preserved variables to be added additively.

### Note on `extend_preserve_env_list`

To maintain the original zero-copy behavior for strings and keep the rest of the codebase concise, the signature of `extend_preserve_env_list` is slightly more complex. The implementation could be much simpler if we were willing to use `Vec<String>` to store `SudoPreserveEnv`.

Closes #1588.

## Standards checklist

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -8,8 +8,8 @@ use std::{env, fmt, fs};
 
 use clap::{Parser, ValueEnum};
 use clap_complete::Shell;
-use color_eyre::eyre::Context;
 use color_eyre::eyre::Result;
+use color_eyre::eyre::{Context, OptionExt};
 use etcetera::base_strategy::BaseStrategy;
 use indexmap::IndexMap;
 use merge::Merge;
@@ -845,14 +845,10 @@ pub struct CommandLineArgs {
 }
 
 fn env_args_parser(arg: &str) -> Result<(String, String)> {
-    let parts: Vec<&str> = arg.splitn(2, '=').collect();
-    if parts.len() != 2 {
-        Err(color_eyre::eyre::eyre!(
-            "Environment variable must be in the format NAME=VALUE"
-        ))
-    } else {
-        Ok((parts[0].to_string(), parts[1].to_string()))
-    }
+    let (key, value) = arg
+        .split_once("=")
+        .ok_or_eyre("Environment variable must be in the format NAME=VALUE")?;
+    Ok((key.to_string(), value.to_string()))
 }
 
 impl CommandLineArgs {
--- a/src/sudo.rs
+++ b/src/sudo.rs
@@ -9,6 +9,7 @@ use color_eyre::eyre;
 use color_eyre::eyre::eyre;
 use color_eyre::eyre::Context;
 use color_eyre::eyre::Result;
+use itertools::Itertools;
 use rust_i18n::t;
 use serde::Deserialize;
 use strum::Display;
@@ -23,7 +24,6 @@ use crate::error::UnsupportedSudo;
 use crate::execution_context::ExecutionContext;
 use crate::executor::Executor;
 use crate::terminal::print_separator;
-use crate::utils::join_iter;
 use crate::utils::which;
 
 #[derive(Clone, Debug)]
@@ -410,6 +410,9 @@ impl Sudo {
         }
 
         let mut preserve_env = opts.preserve_env;
+        // The `--env` arguments are set globally in `main.rs`, but sudo by default
+        // does not pass these environment variables through unless explicitly told to.
+        // So we add them here to the preserve_env list.
         let cfg_env_vars = ctx.config().env_variables();
         if !cfg_env_vars.is_empty() {
             let cfg_env_var_keys = cfg_env_vars.iter().map(|(key, _value)| key.as_str());
@@ -439,7 +442,7 @@ impl Sudo {
             },
             SudoPreserveEnv::Some(vars) => match self.kind {
                 SudoKind::Sudo => {
-                    cmd.arg(format!("--preserve-env={}", join_iter(vars.iter(), ",")));
+                    cmd.arg(format!("--preserve-env={}", vars.iter().join(",")));
                 }
                 SudoKind::Run0 => {
                     for env in vars {
@@ -448,7 +451,7 @@ impl Sudo {
                 }
                 SudoKind::Please => {
                     cmd.arg("-a");
-                    cmd.arg(join_iter(vars.iter(), ","));
+                    cmd.arg(vars.iter().join(","));
                 }
                 SudoKind::Doas | SudoKind::WinSudo | SudoKind::Gsudo | SudoKind::Pkexec => {
                     return Err(UnsupportedSudo {
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -1,6 +1,6 @@
 use std::env;
 use std::ffi::OsStr;
-use std::fmt::{Debug, Write};
+use std::fmt::Debug;
 use std::path::{Path, PathBuf};
 use std::process::Command;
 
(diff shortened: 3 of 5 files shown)


src/sudo.rs
```diff
@@ -90,8 +90,13 @@ pub struct SudoExecuteOpts<'a> {
 }
 
 impl<'a> SudoExecuteOpts<'a> {
-    pub fn new() -> Self {
-        Self::default()
+    pub fn new(ctx: &'a ExecutionContext) -> Self {
+        Self::default().extend_preserve_env_list(
```
GideonBear:
```suggestion
        // `--env` arguments are set globally in `main.rs`, but sudo does not pass through environment
        // variables by default. We need to explicitly pass them through here.
        Self::default().extend_preserve_env_list(
```
or something like that

src/sudo.rs
```diff
@@ -90,8 +90,13 @@ pub struct SudoExecuteOpts<'a> {
...
+            ctx.config()
+                .env_variables()
+                .iter()
+                .map(|env_str| env_str.split('=').next().unwrap()),
```
GideonBear: Duplicated across:
https://github.com/topgrade-rs/topgrade/blob/main/src/main.rs#L97-L102
Can we move this to `config.env_variables()` (`-> Map<&str, &str>` or `-> impl Iterator<(&str, &str)>`) and deduplicate?

GideonBear:
> To maintain the original zero-copy behavior for strings and keep the rest of the codebase concise, the signature of `extend_preserve_env_list` is slightly more complex. The implementation could be much simpler if we were willing to use `Vec<String>` to store `SudoPreserveEnv`.

I think it's fine like this, we don't strictly need any clones, and the caller isn't burdened by it.

src/sudo.rs
w568w: Agreed, I'll change `env_variables()` to return an iterator.

GideonBear:
@AThePeanut4 Can I have your review as well?

src/sudo.rs
```diff
@@ -108,10 +111,21 @@ impl<'a> SudoExecuteOpts<'a> {
...
+    where
+        AnyStrIter: IntoIterator<Item = &'a StrOrStrRef>, // Zero-copy reference type, allowing &[&str], &Vec<String>, Vec<&str>, Iter<&str>
+        StrOrStrRef: AsRef<str> + ?Sized + 'a,            // coerced to str, &str, &String
+    {
+        let vars = vars.into_iter().map(|s| s.as_ref());
+        match self.preserve_env {
+            SudoPreserveEnv::All => {}
+            SudoPreserveEnv::Some(ref mut env_list) => env_list.extend(vars),
+            SudoPreserveEnv::None => self.preserve_env = SudoPreserveEnv::Some(vars.collect()),
+        }
```
atoerien: Would it not be easier/simpler to add the user env variables in `Sudo::execute_opts` instead? We have a `ctx` there already, and then we don't need to do any mutation logic in `SudoExecuteOpts`.

Something along these lines:
```rust
let mut preserve_env = opts.preserve_env;
let cfg_env_vars = ctx.config().env_variables();
if !cfg_env_vars.is_empty() {
    match preserve_env {
        SudoPreserveEnv::All => {}
        SudoPreserveEnv::Some(ref mut env_list) => {
            env_list.extend(cfg_env_vars));
        }
        SudoPreserveEnv::None => {
            SudoPreserveEnv::Some(cfg_env_vars);
        }
    }
}
match preserve_env {
...
```
(this almost definitely won't compile, but I think you get the idea)

src/config.rs
```diff
@@ -853,8 +853,13 @@ impl CommandLineArgs {
...
 
-    pub fn env_variables(&self) -> &Vec<String> {
-        &self.env
+    pub fn env_variables(&self) -> impl Iterator<Item = (&str, &str)> {
+        self.env.iter().map(|s| {
+            let mut split = s.splitn(2, '=');
+            let key = split.next().unwrap_or("");
+            let value = split.next().unwrap_or("");
+            (key, value)
+        })
```
atoerien: I think it would be better to eagerly parse the `--env` arguments into key=value pairs using clap, rather than doing it lazily every time `env_variables()` is called. Clap has `value_parser = ...` which we can use:
```rust
pub struct CommandLineArgs {
    ...
    /// Set environment variables
    #[arg(long = "env", value_name = "NAME=VALUE", num_args = 1.., value_parser = parse_env)]
    env: Vec<(String, String)>,
    ...
}

fn parse_env(arg: &str) -> Result<(String, String), Infallible> {
    let mut split = arg.splitn(2, '=');
    let key = split.next().unwrap_or("");
    let value = split.next().unwrap_or("");
    Ok((key.to_string(), value.to_string()))
}
```

GideonBear:
> The one potential issue is that we might want to remove duplicates: e.g. `--env` has `DIFFPROG` and `preserve_env_list(&["DIFFPROG"])`.
> From what I can tell, `sudo` doesn't seem to care, but maybe other implementations do.

Good catch, ~~let's deduplicate it at execution time.~~
Scratch that, let's use a `*Set` instead of a `Vec`, I think that's cleaner, and the order doesn't matter anyway

src/sudo.rs
w568w: > add the user env variables in Sudo::execute_opts instead

Indeed. I originally designed it that way to ensure precedence between code-defined envs and user `--env` flags. But realizing they are additive and commutative, the order actually doesn't matter.

atoerien: It's definitely not a big issue. imo it's just a bit cleaner to have `SudoExecuteOpts` as barebones as possible, like a "keyword arguments" dictionary to `execute_opts()`. `SudoExecuteOpts::new()` should basically only be called "inside" a call to `execute_opts()`.

src/config.rs
```diff
@@ -844,6 +844,17 @@ pub struct CommandLineArgs {
     pub no_self_update: bool,
 }
 
+fn env_args_parser(arg: &str) -> Result<(String, String)> {
+    let parts: Vec<&str> = arg.splitn(2, '=').collect();
+    if parts.len() != 2 {
+        Err(color_eyre::eyre::eyre!(
```
```suggestion
        Err(eyre!(
```

src/config.rs
```diff
@@ -844,6 +844,17 @@ pub struct CommandLineArgs {
     pub no_self_update: bool,
 }
 
+fn env_args_parser(arg: &str) -> Result<(String, String)> {
+    let parts: Vec<&str> = arg.splitn(2, '=').collect();
```
```suggestion
    let parts: Vec<&str> = arg.split_once('=');
```
and then do `.ok_or_eyre`

w568w:
Suggestions applied. Basically rewrote the entire PR, so the commit history looks a bit messy 😅.

src/sudo.rs
```diff
@@ -407,7 +409,18 @@ impl Sudo {
             cmd.arg("-d");
         }
 
-        match opts.preserve_env {
+        let mut preserve_env = opts.preserve_env;
```
GideonBear, line 412: Looks like the explaining comment disappeared?

src/utils.rs
```diff
@@ -151,6 +151,23 @@ pub fn string_prepend_str(string: &mut String, s: &str) {
     *string = new_string;
 }
 
+pub fn join_iter(mut iterator: impl Iterator<Item = impl std::fmt::Display>, separator: &str) -> String {
```
```rust
use itertools::Itertools;
set.iter().join(",");
```
? We have `itertools` in our dependency graph anyways.

w568w: Strange. rust-analyzer was complaining that the crate is missing when I tried to use `itertools`. Might be a cache glitch. 🤨 Anyway I'll add that to the dependency list.

GideonBear: In the dependency graph, it's not a dependency yet; you'll have to `cargo add itertools`

GideonBear: (With "it's in our dependency graph anyways", I mean that it won't be any additional binary size because of that, since it's a transitive dependency already)

w568w: Get it. I'll add it to the dependency.

GideonBear:
Thanks! I'll ask @AThePeanut4 for one last review.

PR #1603, opened 2025-12-10
feat: `--auto-retry [<COUNT>]` flag and `--no-retry` deprecation

uwuclxdy:
## Standards checklist

---

This PR enhances the `--no-retry` flag by implementing `--retry [<COUNT>]` flag to enable automatic retries of failed steps. `--no-retry` was kept to stay compatible with old scripts and is equivalent to `--retry 0`. It also supports permanent configuration from the config file:

```toml
# retry = true
# retry_count = 1
```

### How it works

**if a step fails:**

1. `topgrade` (no flag, no config.toml entry):
    - user will be prompted to do like it was before
2. `topgrade --retry 0` (or set in the config):
    - the step will be automatically skipped
3. `topgrade --retry 5` (or set in the config):
    - the step will be retried 5 times without user confirmation and automatically continue to the next step

### Feedback needed

Should this be made to work with `-y` flag (eg. automatically continue after retrying if `-y` is set, otherwise prompt the user)?

### DIFF
--- a/src/runner.rs
+++ b/src/runner.rs
@@ -154,11 +154,14 @@ impl<'a> Runner<'a> {
                     } else if has_auto_retries_left {
                         attempt += 1;
                     } else {
-                        self.push_result(key, if ignore_failure {
-                            StepResult::Ignored
-                        } else {
-							StepResult::Failure
-						});
+                        self.push_result(
+                            key,
+                            if ignore_failure {
+                                StepResult::Ignored
+                            } else {
+                                StepResult::Failure
+                            },
+                        );
                         break;
                     }
                 }


GideonBear:
It doesn't seem like a good idea to do unattended retries to me. If a step fails, it's expected for the user to do something. Do you have a situation where it is desired to do unattended retries? This seems a bit like an XY-problem.

uwuclxdy:
this is meant for fixing issues which are usually resolved after retrying (for example unstable internet connection which causes a step to fail and pass on 2nd retry for example). of course default behavior hasn't changed and user is still prompted what they want to do if `--retry` is not set. PR implements this for only the people who need it.

GideonBear:
Okay, network conditions is fair; not every tool has this built-in. But don't we want "ask after 2 retries", instead of "skip after 2 retries"? Or possibly both?

uwuclxdy:
yes, this would make a lot of sense. i was thinking about asking after all retries fail by default, skip only if a flag such as `-y` is also set.

GideonBear:
> skip only if a flag such as `-y` is also set.

FYI: I don't think it's a good idea to use `-y` for Topgrade internals at all; it's only for passing through to commands.

uwuclxdy:
agree + `-y` currently doesn't skip failed steps. i made `--no-retry` work with `--retry` so it first retries [count] amount of times, then continues if `--no-retry` is also set. the naming is somewhat confusing and would appreciate a suggestion on how to name the extra flag (or rename --no-retry, remove from help and only keep an alias for it in code so legacy scripts continue to work). the new parameter could be named something like `--skip-failed` ig

GideonBear:
(I was thinking `--retry=no`, `--retry=ask`, `--retry=5` but that doesn't work because `--retry=5` doesn't specify whether to ask or skip after all 5 retries fail.)

I agree `--skip-failed` is a more sensible name for what is currently called `--no-retry`, when we add automatic retries. `--no-retry` can indeed be an undocumented alias (maybe mention `--no-retry` in the description for `--skip-failed`) and then we can remove that in v17. But maybe there's an even better name, like `--no-ask-retry` maybe? `--skip-failed` still seems a bit confusing in combination with `--retry=5`.

One thing I don't understand is your definition of the retry count. You're using `--retry=1` as default, meaning ask immediately after failure; but the first try is not a retry by definition, I would expect `--retry=1` to mean 2 tries in total (1 retry), and the default to be `--retry=0`. Do you disagree?

uwuclxdy:
alright I'll make `--no-retry` undocumented alias for `--no-ask-retry` flag (name can be changed if needed later).

as for `--retry=1`, this is equal to `--retry`, which will retry once. so running without a number of retries specified defaults to 1 retry. this should all be documented but don't know in how much detail should we go to not make it too bloated.

**examples:**
`--retry` / `--retry 1` -> fail, retry once, ask
`--retry 5` -> fail, retry 5 times, ask

`--no-ask-retry` auto continues instead of asking

uwuclxdy:
also (i missed that), yes maybe my code is poorly written but `--retry 1` will run a step twice if it fails. 

this happens here: https://github.com/topgrade-rs/topgrade/pull/1603/commits/5c21ae62316e0ee1b6dbbea05126592e1908eb97#diff-68c6f449503a33c74a15800c0a9c8fd34028465a06edcac92b32bbb9db64bbceR80

edit: `--retry 0` means skip without asking.... which kinda eliminates the need for `--no-ask-retry`? i somehow forgot about that lol. need to document it and just make `--no-retry` undocumented, as it equals to `--retry 0` already.
https://github.com/uwuclxdy/topgrade/blob/5c21ae62316e0ee1b6dbbea05126592e1908eb97/src/config.rs#L1112

uwuclxdy:
this is the behaviour as of the latest commit in this PR:

- `retry = false` (or unset) -> ask on failure
- `retry = true` -> auto retry failed steps (if `retry_count` not set, retry once)
- `retry = true` + `retry_count = 0` or `no_retry = true` -> don't retry, skip failed steps and continue
- `retry = true` + `retry_count = 5`  -> retries a failed step 5 more times

in cli:
- unset -> ask (the same as `retry = false` or unset)
- `--retry` -> auto retry failed steps (if `retry_count` not set, retry once)
- `--retry 0` -> don't retry, skip failed steps and continue
- `--retry 5` -> retries a failed step 5 more times

now, i updated the example config file idk if it's descriptive enough tho

GideonBear:
Is there a reason to have a default for `--retry`? I feel like it only makes it more confusing.

> yes maybe my code is poorly written but `--retry 1` will run a step twice if it fails.

Oh okay, I just skimmed your code and probably misunderstood.

I'm not really a fan of the disparity between the config file and the CLI in your explanation. I had the following in mind:
* `ask_retry = true/false (default true)` and `--no-ask-retry`: if `false`, do not pause to ask for user input after failure/after automatic retries
* `auto_retry: <some number> (default 0)` and `--auto-retry=<some number>`: the amount of times to automatically retry a step when it fails. After failing X+1 times, will resume like normal (go to next step if `ask_retry = false`, and ask if `ask_retry = true`)

I changed `retry` -> `auto-retry` because it makes it more clear that it's about automatic retries, seems better to me when not overloading `--retry` with other meaning.
`stop_on_error` (`--no-stop-on-error`) might also be a good (better?) name for `ask_retry`.

Feel free to argumentate your current proposal, it just doesn't seem very intuitive to me.

uwuclxdy:
your reasoning for this is actually way better than what i could come up with, so this is how it works now:

| config | CLI flag | description | default |
|-------------|----------|-------------|---------|
| `ask_retry = true/false` | `--no-ask-retry` | prompt user on failure | `true` |
| `auto_retry = <number>` | `--auto-retry=<COUNT>` | number of automatic retries per failed step | `0` |
| `no_retry` | `--no-retry` | hidden alias for `--no-ask-retry` | |

- when a step fails, auto retry `auto_retry` times
- after failing a step for `auto_retry + 1` times:
  - `ask_retry = true` -> prompt the user
  - `ask_retry = false` -> automatically continue

src/config.rs
```diff
@@ -104,6 +104,12 @@ pub enum UpdatesAutoReboot {
     Ask,
 }
 
+#[derive(Debug, Clone, Copy)]
+pub struct RetryConfig {
+    pub auto_retry: u16,
+    pub ask_retry: bool,
+}
```
GideonBear: Is this still necessary? Can we add `config.auto_retry() -> u16` and `config.ask_retry()` instead of `config.retry_config()`?

src/config.rs
```diff
@@ -1086,13 +1102,40 @@ impl Config {
...
-        self.opt.no_retry
-            || self
-                .config_file
-                .misc
-                .as_ref()
-                .and_then(|misc| misc.no_retry)
-                .unwrap_or(false)
+        let config = self.retry_config();
+        config.auto_retry == 0 && !config.ask_retry
+    }
```
GideonBear: Where is this used? It looks obsolete.

src/config.rs
```diff
@@ -1086,13 +1102,40 @@ impl Config {
...
+
+    /// Get retry config
+    pub fn retry_config(&self) -> RetryConfig {
```
GideonBear: Like I said, if we split this up, we can make them standard, like `config.cleanup()`

src/runner.rs
```diff
@@ -70,7 +70,59 @@ impl<'a> Runner<'a> {
...
+                                self.push_result(
+                                    key,
+                                    if ignore_failure {
+                                        StepResult::Ignored
+                                    } else {
+                                        StepResult::Failure
+                                    },
+                                );
+                            }
+                        }
```
GideonBear: Heavy duplication here (with lines 158-179), I don't think that is a good idea in the current state. This is probably hard to get right, feel free to reach out if you're having trouble!

uwuclxdy: added

```rust
enum RetryDecision {
    Retry,
    Quit,
    Continue(StepResult),
}
```

let me know how it looks to you

src/config.rs
```diff
@@ -346,7 +352,9 @@ pub struct Misc {
 
     assume_yes: Option<bool>,
 
-    no_retry: Option<bool>,
```
GideonBear, line 360: Should this also be kept for compatibility?

uwuclxdy: what if we write a config migrator that updates the legacy entries to the new format instead?

GideonBear: I want to redo the whole config for v17 (or v18) anyway, including writing a migrator, I don't think it makes sense to do that for this small change. I think it's best to keep a legacy entry for now, and remove and migrate it in v17.

GideonBear: (accidentally submitted this comment already BTW, not done with reviewing yet)

uwuclxdy: > I think it's best to keep a legacy entry for now, and remove and migrate it in v17.

yep that makes sense

src/runner.rs
```diff
@@ -96,31 +127,31 @@ impl<'a> Runner<'a> {
...
+
+                    // Decide whether to prompt the user
+                    let has_auto_retries_left = attempt < max_attempts;
+                    let should_prompt =
+                        interrupted || ignore_failure || (!has_auto_retries_left && retry_config.ask_retry);
+
+                    if should_prompt {
+                        match self.handle_retry_prompt(&key, &e, ignore_failure)? {
+                            RetryDecision::Retry => {
+                                max_attempts += 1;
```
GideonBear: `attempt += 1` is not in this codepath, so I think this is unnecessary?

src/runner.rs
```diff
@@ -96,31 +127,31 @@ impl<'a> Runner<'a> {
...
                             }
-                            break;
+                            RetryDecision::Continue(result) => {
+                                self.push_result(key, result);
+                                break;
+                            }
                         }
-                        ShouldRetry::Yes => (),
+                    } else {
+                        // Store error and retry
```
GideonBear: Store? `s/Store/Ignore/`?

src/config.rs
GideonBear: Not resolved, the method just calls `.auto_retry()` and `.ask_retry()`.

GideonBear:
## auto_retry = 0, ask_retry = true (default)

* max_attempts = 1
* attempt = 1
* has_auto_retries_left = false
* should_prompt = !has_auto_retries_left && retry_config.ask_retry = true
* handle_retry_prompt
  * Retry: max_attempts += 1

Tested, behavior is correct.

## auto_retry = 0, ask_retry = false

* max_attempts = 1
* attempt = 1
* has_auto_retries_left = false
* should_prompt = false
* attempt += 1
* Continues??

Previously this would:
* should_ask = false
* should_retry = ::No
* self.push_result()
* break

But now, it infinitely reruns the step.

image

So the functionality is broken.

The code looks clean though!

uwuclxdy:
this should do it now

GideonBear:
Functionality seems to be correct now!

src/config.rs
```diff
@@ -787,10 +785,18 @@ pub struct CommandLineArgs {
     #[arg(short = 'r', long = "run-type", value_enum, default_value_t)]
     run_type: RunType,
 
-    /// Do not ask to retry failed steps
-    #[arg(long = "no-retry")]
+    /// Do not ask to retry failed steps (same as --no-ask-retry, kept for legacy compatibility)
+    #[arg(long = "no-retry", hide = true)]
     no_retry: bool,
 
+    /// Whether to ask what to do after a step fails (default: true)
```
```suggestion
    /// Do not ask what to do after a step fails
```
On a `--no-*` arg, this is kind of confusing

src/config.rs
```diff
@@ -787,10 +785,18 @@ pub struct CommandLineArgs {
...
+    #[arg(long = "no-ask-retry")]
+    no_ask_retry: bool,
+
+    /// Auto retry failed steps up to COUNT times
```
```suggestion
    /// Automatically retry failed steps up to COUNT times
```

src/runner.rs
```diff
@@ -96,31 +127,32 @@ impl<'a> Runner<'a> {
...
-                                } else {
-                                    StepResult::Failure
-                                },
-                            );
-                            if let ShouldRetry::Quit = should_retry {
+
+                    // Decide whether to prompt the user
+                    let has_auto_retries_left = attempt < max_attempts;
+                    let should_prompt =
+                        interrupted || ignore_failure || (!has_auto_retries_left && retry_config.ask_retry);
```
GideonBear: `ignore_failures` functionality is broken. Should this be `!ignore_failure`?

src/runner.rs
```diff
@@ -96,31 +127,32 @@ impl<'a> Runner<'a> {
...
+                            RetryDecision::Continue(result) => {
+                                self.push_result(key, result);
+                                break;
+                            }
                         }
-                        ShouldRetry::Yes => (),
+                    } else if has_auto_retries_left {
+                        attempt += 1;
+                    } else {
+                        self.push_result(key, StepResult::Failure);
```
GideonBear: Should this be `StepResult::Ignored` `if ignore_failure`?

src/runner.rs
uwuclxdy: yeah oops I missed that

src/runner.rs
```diff
@@ -96,31 +127,36 @@ impl<'a> Runner<'a> {
...
+                            }
+                        }
+                    } else if has_auto_retries_left {
+                        attempt += 1;
+                    } else {
+                        if ignore_failure {
+                            self.push_result(key, StepResult::Ignored);
+                        } else {
+                            self.push_result(key, StepResult::Failure);
                         }
```
GideonBear:
```suggestion
                        self.push_result(key, if ignore_failure {
                            StepResult::Ignored
                        } else {
							StepResult::Failure
						});
```

src/runner.rs
```diff
@@ -96,31 +127,36 @@ impl<'a> Runner<'a> {
...
-                                } else {
-                                    StepResult::Failure
-                                },
-                            );
-                            if let ShouldRetry::Quit = should_retry {
+
+                    // Decide whether to prompt the user
+                    let has_auto_retries_left = attempt < max_attempts;
+                    let should_prompt =
+                        interrupted || !ignore_failure || (!has_auto_retries_left && retry_config.ask_retry);
```
GideonBear: Wait sorry that also doesn't make any sense. Now if `ignore_failure == false`, `ask_retry` and `auto_retry` have no effect. I think this should work (but please review and test):

```suggestion
                        // If interrupted, always prompt
                        interrupted
                            // Otherwise, only prompt when we don't want to auto-retry, are configured
                            //  to prompt, and don't ignore failure for this step (when we ignore_failure,
                            //  we don't prompt regardless of ask_retry)
                            || (!has_auto_retries_left && retry_config.ask_retry && !ignore_failure);
```

src/config.rs
```diff
@@ -107,6 +105,12 @@ pub enum UpdatesAutoReboot {
     Ask,
 }
 
+#[derive(Debug, Clone, Copy)]
+pub struct RetryConfig {
+    pub auto_retry: u16,
+    pub ask_retry: bool,
+}
```
GideonBear: Is this still necessary? Can we just remove this and `config.retry_config()`?

uwuclxdy:
tested, works as intended:

| config | flag | on fail |
|-------------|----------|--------|
| `ask_retry = true/false` | `--no-ask-retry` | continues to next |
| `auto_retry = <number>` | `--auto-retry=<COUNT>` | retries `<COUNT>` times then prompt if no `--no-ask-retry` |
| `no_retry` | `--no-retry` | same as `--no-ask-retry` |

GideonBear:
Tested and working, also in combination with `ignore_failures`!

PR #1630, opened 2025-12-21
ci(release): add OpenBSD release steps

Izder456:
## What does this PR do

Closes #1584

## Standards checklist

- [x] The PR title is descriptive
- [x] I have read `CONTRIBUTING.md`
- [ ] *Optional:* I have tested the code myself
- [ ] If this PR introduces new user-facing messages they are translated

## For new steps

- [ ] *Optional:* Topgrade skips this step where needed
- [ ] *Optional:* The `--dry-run` option works with this step
- [ ] *Optional:* The `--yes` option works with this step if it is supported by
  the underlying command

If you developed a feature or a bug fix for someone else and you do not have the
means to test it, please tag this person here.

@GideonBear I don't have CI permission in my fork, so this is entirely untested on my part. Hopefully this should work...however.

PR #1639, opened 2025-12-22
ci(release): fix OpenBSD release job

GideonBear:
## What does this PR do
@Izder456 I found some more errors(?) in #1630:
1. Adding it to `trigger.needs` makes it block the entire pipeline if it fails, my bad
2. Looks like the filename won't be correct, `self_update` requires a correct triplet. Was there a reason you deviated from the standard `run` block from above?

Please have a look, I don't want to break it if there was a reason you did it differently :)

## Standards checklist

### DIFF
--- a/.github/workflows/create_release_assets.yml
+++ b/.github/workflows/create_release_assets.yml
@@ -348,7 +348,7 @@ jobs:
           FILENAME=topgrade-${tag}-$(default-target)
           mv target/release/topgrade assets
           cd assets
-          tar --format=ustar -czf "$FILENAME.tar.gz" topgrade
+          tar -F ustar -czf "$FILENAME.tar.gz" topgrade
           rm topgrade
           ls .
 


.github/workflows/create_release_assets.yml
```diff
@@ -343,11 +343,12 @@ jobs:
...
         run: |
+          cargo install default-target
           mkdir -p assets
-          FILENAME=topgrade-${tag}-openbsd-x86_64
-          cp target/release/topgrade assets/topgrade
+          FILENAME=topgrade-${tag}-$(default-target)
+          mv target/release/topgrade assets
           cd assets
-          tar -czf "$FILENAME.tar.gz" topgrade
+          tar --format=ustar -czf "$FILENAME.tar.gz" topgrade
```
Izder456: OpenBSD's tar command lacks a `--format` i'm pretty sure. ~~it defaults to this format anyways.~~ the filename is fine, and should have been done that way the first time.

Izder456: Nevermind, https://man.openbsd.org/tar#F. You can use the -F flag for this on OpenBSD. the default is actually `pax`. 

Change this out to use `-F ustar` rather than `--format=ustar` and that should be correct

Izder456: @GideonBear could you review my suggestion?

Izder456:
@GideonBear i've sent suggestions in my review. Could you take a look at them?

GideonBear:
> @GideonBear i've sent suggestions in my review. Could you take a look at them?

I don't see anything. To confirm; you've sent them in this PR?
Are you sure you've submitted the comments, not just queued them up? Are they marked as "Pending"?

Izder456:
> > @GideonBear i've sent suggestions in my review. Could you take a look at them?
> 
> 
> 
> I don't see anything. To confirm; you've sent them in this PR?
> 
> Are you sure you've submitted the comments, not just queued them up? Are they marked as "Pending"?

I don't use GitHub too much so it might be PEBCAK on my part, but yes they are marked as pending.

The reviews were on the code itself, not as a comment in the PR.

To reiterate, the changes i made for the naming were erroneous during my first attempt at making the CI. The correction you made looks fine.

The `--format` flag does not exist in OpenBSD's `tar`, though `-F` does. https://man.openbsd.org/tar#F

So, if you replace `--format ustar` with `-F ustar` it should be good.

GideonBear:
> but yes they are marked as pending

Pending means not sent, that explains why I can't see them. You should be able to post them with the button in the top right corner in the code view.

Izder456 (changes requested):

.github/workflows/create_release_assets.yml
GideonBear:
```suggestion
          tar -F ustar -czf "$FILENAME.tar.gz" topgrade
```

GideonBear:
Thanks for the review!

Issue #1472, opened 2025-11-11
Overhaul the config file structure


atoerien:
## I want to suggest some general feature

The current structure of the config file is rather complicated and unintuitive - see #1193.

I propose a new structure where every step is a new subsection, inspired by [starship's config file](https://starship.rs/config/). Each one would have an enabled key, and those that need more stuff in the config file could have more keys. Custom steps could be `[custom.{name}]`, just like starship. Generic config options (basically those under `[misc]`) could go in the unnamed section at the top of the file.

This would obviously be a big breaking change, but I think it would make the configuration much easier to understand. As part of this, there could also be a website similar to starship's that explains all the config options, rather than having a config.example.toml listing everything.

GideonBear:
Ref #683, which are the `[custom.{name}]` you are talking about.
They can replace `[commands]` entirely.
They should be able to be used with `--only` and `--disable`, which is #392.

GideonBear:
~~There should be a config migration tool built into Topgrade that users can use to (while not preserving comments) migrate their config to v18.~~
Opened #2240

GideonBear:
Ref #515
Ref #1387
Ref #1049 (pay attention to where we want this in the new config file)
Ref #624 ?
Ref #399
Ref #194

GideonBear:
Let's split up the files to one step(group) per file, make the section names and filename exactly the step id, and move the definition of the config substructs to those step files.

GideonBear:
Opt-in steps currently have a manual check for `[step] enabled = true` in the config. These manual checks can be removed, and we can combine that opt-in behavior with `disable` and `only`.

GideonBear:
Ref #748 
I said in #515:
> I quite like the idea of `cleanup = true` just meaning `[*.cache] enable = true`

I also quite like the idea of `assume_yes = true` just meaning `[*] yes = true`.

GideonBear:
Ref #1204, which is either not necessary if we make the config good enough, or should be tackled at the same time

GideonBear:
We need to think about the relation between a step's "id" (used for `--only` and config) and a step's "key" (the pretty name)

GideonBear:
Let's also rethink and hopefully remove all the `Config::some_option` methods, and do the cli/config merging and defaults declaratively instead.

The CLI/config parity needs to be overhauled. Probably remove some CLI flags which should be config-only. And maybe add a few CLI-only flags to the config.

GideonBear:
[...]

PR #1685, opened 2026-01-14
feat(cargo): add `git` and `quiet` options

flying-sheep:
## What does this PR do
Adds two boolean options for the `cargo install-update` command (referred to in user-facing output as “Cargo” and in function names as `cargo_update`):

- `git` (default: `true`) makes the `--git` flag configurable. `cargo install-update` disables this by default because of cost, we enable it by default. (and before this PR without a way to disable it again)
- `quiet` (default: `false`) suppresses all output when there are no errors. I wish there was a way to say “no output unless something is updated” but AFAIK there’s only this.

Testing done: I ran it on my system first with `git=false` set (to see the “git” part vanish from the output), then with `quiet=true` to see the output vanish.

## Standards checklist

### DIFF
--- a/src/steps/generic.rs
+++ b/src/steps/generic.rs
@@ -71,7 +71,7 @@ pub fn run_cargo_update(ctx: &ExecutionContext) -> Result<()> {
     };
 
     let mut command = ctx.execute(cargo_update);
-    command.arg("install-update");
+    command.args(["install-update", "--all"]);
     if ctx.config().cargo_update_git() {
         command.arg("--git");
     }


GideonBear:
I'm not happy with adding config options for every flag on every step (see https://github.com/topgrade-rs/topgrade/issues/1472#issuecomment-3753486910), so this might get removed later (with a general flags system replacing it). But until then, this is a valid configurability.

GideonBear (approved):
Thanks!

PR #1712, opened 2026-01-28
feat(system,home_manager): integrate nh into nix-based updaters

dtomvan:
## What does this PR do

Fixes a bug where both `nixos-rebuild` and `nh` get used at the same time on NixOS. Both commands effectively do the same, so this duplicates work which is unwanted.

This deprecates the `nix_helper` step.

New config option: `misc.nix_handler`. Decides which handler to run during the `system` (on NixOS) and `home_manager` steps.

Example for the `system` step on NixOS:

| nix\_handler value  | nh available | effective updater |
| ------------------- | ------------ | ----------------- |
| autodetect          | yes          | nh                |
|                     | no           | nixos\-rebuild    |
| nh                  | yes          | nh                |
|                     | no           | ERROR             |
| vanilla             | yes          | nixos\-rebuild    |
|                     | no           | nixos\-rebuild    |

The same is true with s/nixos-rebuild/home-manager/g for the `home_manager` step.

Fixes #1708

## Standards checklist

- [x] The PR title is descriptive
- [x] I have read `CONTRIBUTING.md`
- [x] *Optional:* I have tested the code myself (**dry-ran**)
- [x] If this PR introduces new user-facing messages they are translated

cc @jackaed @gideonbear

PR #1719, opened 2026-02-01
refactor(system,home-manager): make matches more explicit

GideonBear:
## What does this PR do
Follow-up on https://github.com/topgrade-rs/topgrade/pull/1712#discussion_r2749661970
Ping @dtomvan, please review

Personally, I don't like `_ =>`. There are two separate cases which can hit it, writing out those cases makes it more self-documenting. I like knowing that my mental model is checked by the compiler.

I also removed `.is_ok()` and the `.unwrap()`, opting for an explicit `unreachable!` case instead, which (again) utilizes the compiler, and makes the only unverified operation trivially verifiable manually.

Comments were added because the cases were getting long, and especially with `(_, _, _)`, a bit hard to read without context.

Do you agree with these changes?

I thought about
```rust
enum RequireOneOfTwoResult {
    Both(PathBuf, PathBuf),
    Left(PathBuf),
    Right(PathBuf),
}
require_one_of_two() -> Result<RequireOneOfTwoResult>
```
Or:
```rust
enum Preference {
    None,
    Left,
    Right,
}
require_one_of_two(preference: Preference) -> Result<PathBuf>
```

To completely remove the unreachable case, but that's very overkill (for now)

## Standards checklist

### DIFF
--- a/src/steps/os/unix.rs
+++ b/src/steps/os/unix.rs
@@ -945,8 +945,7 @@ pub fn run_home_manager(ctx: &ExecutionContext) -> Result<()> {
 
             cmd.status_checked()
         }
-        // Unreachable because of require_one
-        (Err(_), _, Err(_)) => unreachable!(),
+        (Err(_), _, Err(_)) => unreachable!("require_one([\"home-manager\", \"nh\"])?; was called, so either tool must be available"),
     }
 }
 


src/steps/os/unix.rs
```diff
@@ -939,6 +945,8 @@ pub fn run_home_manager(ctx: &ExecutionContext) -> Result<()> {
 
             cmd.status_checked()
         }
+        // Unreachable because of require_one
+        (Err(_), _, Err(_)) => unreachable!(),
```
dtomvan: You can specify why some codepath is unreachable inside the macro: `unreachable!("require_one([\"home-manager\", \"nh\"])?; was called, so either tool must be available")`

Issue #1612, opened 2025-12-13
`impl Clone for Executor`


GideonBear:
Possibly related to #1227
Ping @AThePeanut4 

`impl Clone for Executor`

To simplify e.g.:
* A `-y` that needs to be added for every command
* Use sudo for every command
* Use sudo for every command in some cases, and sudo for none of the commands in others

This pattern could be used:
```rs
base_cmd = ctx.execute(program).add_args_and_stuff();
base_cmd.clone().arg("update").status_checked()?
```

This requires `Executor` to become a pre-`Command` type (`(PathBuf, Vec<String>)`) that is only converted to a `Command` when executing, because `Command` is not `Clone`.

An example:
https://github.com/topgrade-rs/topgrade/blob/639e1241ec90f4e4a5cae1218248d95ef4200330/src/steps/os/archlinux.rs#L273-L307
Would become:
```rs
let mut base_cmd = if version >= version_no_sudo {
    ctx.execute(&self.executable)
else {
    let sudo = ctx.require_sudo()?;
    sudo.execute(ctx, &self.executable)?
};

if ctx.config().yes(Step::System) {
    base_cmd.arg("--noconfirm");
}
// Assuming the order of arguments doesn't matter
base_cmd.args(ctx.config().aura_pacman_arguments().split_whitespace());

base_cmd.clone().arg("-Au").status_checked()?;
base_cmd.clone().arg("-Syu").status_checked()?;
```
Removing three layers of duplication (`sudo`, `--noconfirm`, `ctx.config().aura_pacman_arguments()`).

GideonBear:
With #1735, what is now the `DryCommand` could just become the pre-`Command` type (because it is already convertible to a `Command`).

GideonBear:
It's theoretically possible, after this is implemented, to make `require` return a `Command` (`ctx.require`, which will return `ctx.execute(require_path(cmd))`).
This would need some research on where this would make the code cleaner/less clean, this might turn out to be a bad idea (if the path is used in a lot of places, and/or if `.clone` everywhere is more cumbersome than `ctx.execute`)
Ping @uwuclxdy

GideonBear:
Additional related changes that could be made, inspired by / building on #2226:
* Possibly add `env_remove_if` and `env_remove_if_some` and `current_dir_if`, etc.
[...]

PR #1735, opened 2026-02-04
feat(executor): add `Executor::always()` and ban `Command::new`

### LINKED ISSUE #1227: `Command::new` should be banned if possible, and removed from everywhere
Especially with the possible addition of #1217, the `Executor` is actually important besides just facilitating dry-running. Let's add `Executor::execute_always` to run regardless of dry-running, then ban `Command::new`, then for each invocation of it, see if we want to use `Executor::execute_always` or `Executor::execute`.
Ref #1208 (https://github.com/topgrade-rs/topgrade/pull/1208#discussion_r2209543814)

rbutoi:
## What does this PR do

Adds a `Executor::always()` method that converts `Dry` to `Wet`, allowing read-only commands to run even in dry-run mode. Bans `Command::new` using clippy and migrates all `Command::new` calls to `ctx.execute().always()` for version checks, config queries, and environment detection (or avoids running a subprocess altogether). 

Closes #1227

## Standards checklist

## For new steps

  the underlying command

### DIFF
--- a/src/execution_context.rs
+++ b/src/execution_context.rs
@@ -80,6 +80,7 @@ impl<'a> ExecutionContext<'a> {
     }
 
     /// Create an instance of `Executor` that should run `program`.
+    #[allow(clippy::disallowed_methods)]
     pub fn execute<S: AsRef<OsStr>>(&self, program: S) -> Executor {
         match self.run_type {
             RunType::Dry => Executor::Dry(DryCommand::new(program)),
--- a/src/executor.rs
+++ b/src/executor.rs
@@ -243,6 +243,7 @@ impl DryCommand {
     }
 
     /// Convert this dry command into a real Command that will execute.
+    #[allow(clippy::disallowed_methods)]
     fn into_command(self) -> Command {
         let mut cmd = Command::new(&self.program);
         cmd.args(&self.args);
--- a/src/steps/tmux.rs
+++ b/src/steps/tmux.rs
@@ -64,6 +64,7 @@ impl Tmux {
         }
     }
 
+    #[allow(clippy::disallowed_methods)]
     fn build(&self) -> Command {
         let mut command = Command::new(&self.tmux);
         if let Some(args) = self.args.as_ref() {
--- a/src/terminal.rs
+++ b/src/terminal.rs
@@ -30,6 +30,7 @@ pub fn shell() -> &'static str {
     which("pwsh").map(|_| "pwsh").unwrap_or("powershell")
 }
 
+#[allow(clippy::disallowed_methods)]
 pub fn run_shell() -> eyre::Result<()> {
     Command::new(shell()).env("IN_TOPGRADE", "1").status_checked()
 }
--- a/clippy.toml
+++ b/clippy.toml
@@ -1,4 +1,5 @@
 disallowed-methods = [
+    { path = "std::process::Command::new", reason = "Use `ctx.execute()` or `ctx.execute().always()`" },
     { path = "std::process::Command::output", reason = "Use `output_checked[_with][_utf8]`" },
     { path = "std::process::Command::spawn", reason = "Use `spawn_checked`" },
     { path = "std::process::Command::status", reason = "Use `status_checked`" },
--- a/src/config.rs
+++ b/src/config.rs
@@ -736,6 +736,7 @@ impl ConfigFile {
         let command = which(&editor[0])?;
         let args: Vec<&String> = editor.iter().skip(1).collect();
 
+        #[allow(clippy::disallowed_methods)]
         Command::new(command)
             .args(args)
             .arg(config_path)
--- a/src/self_update.rs
+++ b/src/self_update.rs
@@ -59,6 +59,7 @@ pub fn self_update(ctx: &ExecutionContext) -> Result<()> {
         {
             if result.updated() {
                 print_info(t!("Respawning..."));
+                #[allow(clippy::disallowed_methods)]
                 let mut command = Command::new(current_exe?);
                 command.args(env::args().skip(1)).env("TOPGRADE_NO_SELF_UPGRADE", "");
 


rbutoi:
lmk if I should split this up!

src/executor.rs
```diff
@@ -77,6 +91,19 @@ impl Executor {
...
     }
 
+    #[allow(dead_code)]
+    /// See `std::process::Command::stdin`
+    pub fn stdin<T: Into<Stdio>>(&mut self, cfg: T) -> &mut Executor {
+        match self {
+            Executor::Wet(c) | Executor::Damp(c) => {
+                c.stdin(cfg);
+            }
+            Executor::Dry(_) => (),
```
GideonBear: Why no `DryCommand.stdin`, seems like this gets lost now? All other commands store the args when `Dry`.

GideonBear:
> There are also a lot of normal executions that should run in dry-runs, like quering the `--version`. These should also be converted.

Looks like you didn't do (all of) these (I saw `fisher --version` randomly), that's fine, one of us can do that in a follow-up PR.

src/steps/generic.rs
```diff
@@ -32,9 +31,9 @@ use crate::{
 
 #[cfg(target_os = "linux")]
 pub fn is_wsl() -> Result<bool> {
-    let output = Command::new("uname").arg("-r").output_checked_utf8()?.stdout;
-    debug!("Uname output: {}", output);
-    Ok(output.contains("microsoft"))
+    let output = std::fs::read_to_string("/proc/version").unwrap_or_default();
+    debug!("Proc version output: {}", output);
+    Ok(output.to_lowercase().contains("microsoft"))
```
GideonBear, line 36: Tested on wsl?

src/steps/node.rs
```diff
@@ -42,14 +41,15 @@ impl Display for NPMVariant {
 }
 
 #[allow(clippy::upper_case_acronyms)]
-struct NPM {
+struct NPM<'a> {
     command: PathBuf,
     variant: NPMVariant,
+    ctx: &'a ExecutionContext<'a>,
```
GideonBear: Can the `ctx` be passed in through its methods instead? What was the rationale for adding it as struct member?

src/steps/node.rs
```diff
@@ -115,13 +120,14 @@ impl NPM {
     }
 }
 
-struct Yarn {
+struct Yarn<'a> {
     command: PathBuf,
+    ctx: &'a ExecutionContext<'a>,
```
GideonBear: Same here

src/steps/node.rs
```diff
@@ -171,19 +184,20 @@ impl Yarn {
     }
 }
 
-struct Deno {
+struct Deno<'a> {
     command: PathBuf,
+    ctx: &'a ExecutionContext<'a>,
```
GideonBear: Same here

src/steps/tmux.rs
```diff
@@ -64,6 +64,7 @@ impl Tmux {
         }
     }
 
+    #[allow(clippy::disallowed_methods)]
```
GideonBear, line 67: Why?

src/execution_context.rs
```diff
@@ -56,7 +56,7 @@ pub struct ExecutionContext<'a> {
     under_ssh: bool,
     #[cfg(target_os = "linux")]
     distribution: &'a Result<Distribution>,
-    powershell: LazyLock<Option<Powershell>>,
+    powershell: OnceLock<Option<Powershell>>,
```
GideonBear, line 59: Is this related?

src/terminal.rs
```diff
@@ -30,8 +30,8 @@ pub fn shell() -> &'static str {
     which("pwsh").map(|_| "pwsh").unwrap_or("powershell")
 }
 
-pub fn run_shell() -> eyre::Result<()> {
-    Command::new(shell()).env("IN_TOPGRADE", "1").status_checked()
+pub fn run_shell(ctx: &ExecutionContext) -> eyre::Result<()> {
+    ctx.execute(shell()).env("IN_TOPGRADE", "1").status_checked()
```
GideonBear: Missing `.always()`?
Anyway I think this should be `Command::new` since it's not a "command" that's being ran, it's only used to drop into shell when a step fails.

src/utils.rs
```diff
@@ -163,10 +163,8 @@ pub fn hostname() -> Result<String> {
 
 #[cfg(windows)]
 pub fn hostname() -> Result<String> {
-    Command::new("hostname")
-        .output_checked_utf8()
+    std::env::var("COMPUTERNAME")
```
GideonBear, line 166: Tested on Windows?

src/utils.rs
```diff
@@ -248,8 +246,8 @@ pub mod merge_strategies {
 ///
 /// We do this check through `python -V`, a shim will just give `Python` with
 /// no version number.
-pub fn check_is_python_2_or_shim(python: PathBuf) -> Result<PathBuf> {
-    let output = Command::new(&python).arg("-V").output_checked_utf8()?;
+pub fn check_is_python_2_or_shim(ctx: &ExecutionContext, python: PathBuf) -> Result<PathBuf> {
+    let output = ctx.execute(&python).arg("-V").output_checked_utf8()?;
```
GideonBear: Missing `.always()`?

rbutoi: Fixed

src/terminal.rs
rbutoi: Fixed

src/executor.rs
rbutoi: Oops, not sure what I was thinking here. Thanks, fixed and added a test.

src/execution_context.rs
rbutoi: Yes- `Powershell::new` needs a back-reference to the `ExecutionContext` now, so `Powershell::new(self)` must wait until access time.

src/steps/tmux.rs
rbutoi: This function uses `Command::new`, so it needs to be made an exception

src/steps/node.rs
rbutoi: done

src/steps/node.rs
rbutoi: done

src/utils.rs
rbutoi: I used [a one-off test and CI workflow on a branch](https://github.com/rbutoi/topgrade/commit/41f441fb8293e9e19fe125fab3cdb9f3104775bb) to validate this: https://github.com/rbutoi/topgrade/actions/runs/21843324330

Let me know if I should bring those changes into this PR.

src/steps/generic.rs
rbutoi: https://github.com/rbutoi/topgrade/actions/runs/21843324330

rbutoi:
Thank *you* for the thorough review! I've responded to all the review comments in a set of follow-up commits. These will be squashed into the original 3 commits if there are no more comments.

> > There are also a lot of normal executions that should run in dry-runs, like quering the `--version`. These should also be converted.
> 
> Looks like you didn't do (all of) these (I saw `fisher --version` randomly), that's fine, one of us can do that in a follow-up PR.

I went through all references to `ExecutionContext::execute` and was properly thorough about all version checks / list calls this time.

src/steps/node.rs
rbutoi: Done - the rationale was basically that `ctx` is required for every method so it may as well be a member. But it could definitely go either way

src/utils.rs
GideonBear: Not necessary, thanks!

src/steps/generic.rs
```diff
@@ -1460,7 +1478,7 @@ pub fn run_poetry(ctx: &ExecutionContext) -> Result<()> {
 
         let check_official_install_script =
             "import sys; from os import path; print('Y') if path.isfile(path.join(sys.prefix, 'poetry_env')) else print('N')";
-        let mut command = Command::new(&interp);
+        let mut command = ctx.execute(&interp);
```
GideonBear: `.always()`? This runs a diagnostic script.

src/main.rs
GideonBear, line 1: How did you manage to do that 😂
Please revert

src/terminal.rs
```diff
@@ -203,7 +205,7 @@ impl Terminal {
     }
 
     #[allow(unused_variables)]
-    fn should_retry(&mut self, step_name: &str) -> eyre::Result<ShouldRetry> {
+    fn should_retry(&mut self, ctx: &ExecutionContext, step_name: &str) -> eyre::Result<ShouldRetry> {
```
GideonBear: Useless now and uncaught because of `allow(unused_variables)`

src/terminal.rs
```diff
@@ -30,6 +31,7 @@ pub fn shell() -> &'static str {
     which("pwsh").map(|_| "pwsh").unwrap_or("powershell")
 }
 
+#[allow(clippy::disallowed_methods)]
```
GideonBear, line 33: Unused

src/steps/tmux.rs
GideonBear: Oh this is for running Topgrade in tmux, not upgrading tmux... I was confused by it being `src/steps/tmux.rs` and not `src/tmux.rs`, that should probably be changed.

GideonBear:
`uv` panics in dry-run:
```
Message:  internal error: entered unreachable code: the whole function returns when we run `uv --version` under dry-run
Location: src/steps/generic.rs:1592
```
~~Probably need to make the call that gets that output `.always()` and change the unreachable message to something like `used .always(), cannot return dry output`~~ that's wrong, this is the actual `self update` and the check is secondary
Let's just `return Err(DryRun);` instead of `unreachable!()`.

GideonBear:
I don't actually know if `Damp` is the way to go here. Take this output for example:
```
── 11:46:59 - Cargo ────────────────────────────────────────────────────────────
Dry running: /home/gideon/.cargo/bin/cargo-install-update install-update --all --git
Dry running: /home/gideon/.cargo/bin/cargo-cache -a
Executing: /home/gideon/.local/bin/python -V
Executing: /home/gideon/.local/bin/python3 -V
Executing: /home/gideon/.local/bin/python -m pip
Executing: /home/gideon/.local/bin/python -c 'import sysconfig; from os import path; print('\''Y'\'') if path.isfile(path.join(sysconfig.get_path('\''stdlib'\''), '\''EXTERNALLY-MANAGED'\'')) else print('\''N'\'')'
Executing: /home/gideon/.local/bin/python -m pip config get global.break-system-packages
Executing: /usr/bin/gh extensions list

── 11:47:03 - GitHub CLI Extensions ────────────────────────────────────────────
Dry running: /usr/bin/gh extension upgrade --all
```
In between `Cargo` and `GitHub CLI extensions` it's executing pip commands. Maybe we should just make it `Wet` so we don't see it? I understand your reasoning for `Damp`, but people can also just use `-v` if they really need to see every `--version` call. I think the main reason to use `--dry-run` is to see what *effective* commands will be ran. IMO `command update` > `command --version; command update` > `command --version`. But let me know what you think!

GideonBear:
(I just realized that the way you implement this with a `DryCommand` that converts to a `Command` is very in line with what I had planned in #1612, making that a lot easier!)

GideonBear:
> I went through all references to `ExecutionContext::execute` and was properly thorough about all version checks / list calls this time.

Thanks! I found only one you might've missed:
https://github.com/topgrade-rs/topgrade/blob/015374022c341b61492e520f5129d1405111e9d6/src/steps/os/archlinux.rs#L261

rbutoi:
> I don't actually know if `Damp` is the way to go here. Take this output for example:
> 
> ```
> ── 11:46:59 - Cargo ────────────────────────────────────────────────────────────
> Dry running: /home/gideon/.cargo/bin/cargo-install-update install-update --all --git
> Dry running: /home/gideon/.cargo/bin/cargo-cache -a
> Executing: /home/gideon/.local/bin/python -V
> Executing: /home/gideon/.local/bin/python3 -V
> Executing: /home/gideon/.local/bin/python -m pip
> Executing: /home/gideon/.local/bin/python -c 'import sysconfig; from os import path; print('\''Y'\'') if path.isfile(path.join(sysconfig.get_path('\''stdlib'\''), '\''EXTERNALLY-MANAGED'\'')) else print('\''N'\'')'
> Executing: /home/gideon/.local/bin/python -m pip config get global.break-system-packages
> Executing: /usr/bin/gh extensions list
> 
> ── 11:47:03 - GitHub CLI Extensions ────────────────────────────────────────────
> Dry running: /usr/bin/gh extension upgrade --all
> ```
> 
> In between `Cargo` and `GitHub CLI extensions` it's executing pip commands. Maybe we should just make it `Wet` so we don't see it? I understand your reasoning for `Damp`, but people can also just use `-v` if they really need to see every `--version` call. I think the main reason to use `--dry-run` is to see what _effective_ commands will be ran. IMO `command update` > `command --version; command update` > `command --version`. But let me know what you think!

Yup totally, I see what you mean. If commands are read-only, they're not worth mentioning in a dry run either. Changed to `Wet`.

GideonBear:
One more thing, there's a mode change on `execution_context.sh` in the second commit that you revert in the third.

rbutoi:
Not sure how these chmods keep sneaking in there. Thanks again

GideonBear:
Thanks for this, saves me a lot of work!

PR #1744, opened 2026-02-07
fix(vim): AstroNvim - wait for AstroUpdateCompleted and finish.

slowsage:
## What does this PR do                                                                                                                                                                                                                                         
### Summary of Changes:                                                                                                                                                                                                                                           
- ~~Paq: async quitall~~                                                                                                                                                                                                                                     
- AstroNvim: async quitall + finish (it uses Lazy + Mason, we end early)
- ~~Lazy: remove " | qa" to enable Mason updates. Use async.~~
- ~~All plugin managers: run MasonUpdate and MasonToolsUpdate~~

~~When trying to fix AstroNvim (it used to quit early), I added autocmd (See [AstroNvim/astrocore#62](https://github.com/AstroNvim/astrocore/pull/62)). I refactored upgrade.vim to run plugin managers, then MasonUpdate (tools registry) and then MasonToolsUpdate.~~

~~It uses async when possible including Lazy.~~

~~This is a bigger change than I need personally. It needs testing by other plugin manager users (tagged other contributors to this file below).~~

~~If not ok, I can modify PR to just fix AstroNvim but felt this structure is more extensible and would keep nvim upgrades on par with vscode extension updates.~~

Reduced scope based on feedback. 
## Standards checklist

-  [x] The PR title is descriptive
-  [x] I have read `CONTRIBUTING.md`
- :x: *Optional:* I have tested the code myself - I have only tested AstroNvim. I have tagged other contributors to test other plugin managers.

## For new steps

  the underlying command

### DIFF
--- a/src/steps/upgrade.vim
+++ b/src/steps/upgrade.vim
@@ -4,9 +4,11 @@
 "   skip_prompts = true,
 " },
 if exists(":AstroUpdate")
-   echo "AstroUpdate"
-   AstroUpdate
-   quitall
+    echo "AstroUpdate"
+    autocmd User AstroUpdateCompleted quitall
+    AstroUpdate
+    " Astro includes Lazy etc. So we end early.
+    finish
 endif
 
 if exists(":MasonUpdate")


vzze:
Works fine for lazy + mason (custom config not astrovim), though I think that topgrade should only invoke the top level package manager of a Neovim configuration and not sub-package managers like mason.

A lot of people that use their own custom configs (myself included) already use post-update hooks to deal with mason. Example:

```lua
    build = function()
        pcall(vim.api.nvim_command, [[MasonUpdate]])
    end,
```

In my opinion it's a more sensible choice to exclude the post update autocommands and end call for `s:MasonAndQuit()` because it's not topgrade's fault but the configuration's fault for not updating a package properly.

slowsage:
Did not realize that (post-update hooks). Makes a lot of sense. Reduced scope to just fix AstroNvim.

PR #1760, opened 2026-02-10
feat: add notify_end tristate option

### LINKED ISSUE #1692: Only send notification on items failing: warnings/errors.
## Checklist
- [X] I have searched the issue tracker for relevant or duplicate issues.

Topgrade should have a option to only send a notification if there is a issue in the steps. This is useful because if you have setup a script to automatically do topgrade every so often you only get notified when there are errors, instead of everytime.

## More information

No, extra information.
- [ ] I am able and willing to implement this feature myself

rbutoi:
## What does this PR do

Adds a `notify_end` option that deprecates the `skip_notify` option (but continues respecting it for back compat with a warning). I thought this was a better option than outright removing the old option or adding another bool option (`notify_on_success`).

Closes #1692.

## Standards checklist

## For new steps

  the underlying command

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -370,7 +370,10 @@ pub struct Misc {
 
     notify_each_step: Option<bool>,
 
+    /// Deprecated: use `notify_end = "never"` instead
     skip_notify: Option<bool>,
+
+    notify_end: Option<NotifyEnd>,
 
     bashit_branch: Option<String>,
 
@@ -393,6 +396,20 @@ pub enum TmuxSessionMode {
     #[default]
     AttachIfNotInSession,
     AttachAlways,
+}
+
+/// Controls when the end-of-run desktop notification is sent.
+#[derive(Clone, Copy, Debug, Deserialize, ValueEnum, Default)]
+#[clap(rename_all = "snake_case")]
+#[serde(rename_all = "snake_case")]
+pub enum NotifyEnd {
+    /// Always send a notification (default)
+    #[default]
+    Always,
+    /// Never send a notification
+    Never,
+    /// Only send a notification if there were failures
+    OnFailure,
 }
 
 pub struct TmuxConfig {
@@ -816,9 +833,13 @@ pub struct CommandLineArgs {
     #[arg(short = 'k', long = "keep")]
     keep_at_end: bool,
 
-    /// Skip sending a notification at the end of a run
-    #[arg(long = "skip-notify")]
+    /// Skip sending a notification at the end of a run (deprecated: use --notify-end never)
+    #[arg(long = "skip-notify", hide = true)]
     skip_notify: bool,
+
+    /// When to send a notification at the end of a run
+    #[arg(long = "notify-end", value_enum, default_value_t)]
+    notify_end: NotifyEnd,
 
     /// Say yes to package manager's prompt
     #[arg(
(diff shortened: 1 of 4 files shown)


src/config.rs
```diff
@@ -399,6 +402,20 @@ pub enum TmuxSessionMode {
...
+#[derive(Clone, Copy, Debug, Deserialize, ValueEnum, Default)]
+#[clap(rename_all = "snake_case")]
+#[serde(rename_all = "snake_case")]
+pub enum NotifyEnd {
+    /// Always send a notification
+    Always,
+    /// Never send a notification
+    Never,
+    /// Only send a notification if there were failures (default)
+    #[default]
```
GideonBear: I think I agree but this is a breaking change. Please keep the default the same and open a new issue (I'll milestone it for v17) and argumentate the new default there.

src/config.rs
```diff
@@ -1256,13 +1277,34 @@ impl Config {
...
+
+        let notify_end = self
+            .config_file
+            .misc
+            .as_ref()
+            .and_then(|misc| misc.notify_end)
+            .unwrap_or(self.opt.notify_end);
+
+        if skip_notify {
+            if !matches!(notify_end, NotifyEnd::Always) {
```
GideonBear: Shouldn't this be `NotifyEnd::Never`?

src/config.rs
```diff
@@ -1256,13 +1277,34 @@ impl Config {
...
+                print_warning(format!(
+                    "`skip_notify` conflicts with `notify_end = {:?}`, ignoring `notify_end`. \
+                     Please migrate to `notify_end = \"never\"` and remove `skip_notify`.",
+                    notify_end
+                ));
```
GideonBear: I don't understand exactly when this warning is raised (see also previous comment). But I don't want hard deprecation, e.g. anyone using *only the old settings* should not get any warnings, nor should anyone using *only the new settings*; conflicting settings can warn, but `no_retry`/`ask_retry` doesn't do that either.

v17 (or v18) will remove the deprecated options, I don't want to burden users with partial migrations before then.

src/config.rs
GideonBear, line 1: mode change

rbutoi: 🤦

src/config.rs
rbutoi: Yes, my bad

src/config.rs
rbutoi: Agreed with you, I just made it confusing. The warning shouldn't come up by default, and should only if people explicitly enable conflicting old/new-style options.

GideonBear: But this does warn for me right now (with `skip_notify = true` and `notify_end` unset)

GideonBear: Because the default `notify_end` is `always`, which will always conflict with `skip_notify = true`

rbutoi:
> Good except for the warning

I realized there's nothing that can be done because there's no way to differentiate between "unset" and "unset, defaulted". So no warning, just a TODO. Made https://github.com/topgrade-rs/topgrade/issues/1767 to track it.

Also fwiw I realized that was making my files executable: https://github.com/emacs-mirror/emacs/blob/badef58d0ad97401540ea961f34841c2911683ab/lisp/progmodes/executable.el#L275, and the files beginning with `#![allow(dead_code)]` (since `#!` is a shebang). 🤦

GideonBear:
Thanks! If you still want to change the default please open a new issue.

PR #1799, opened 2026-02-26
feat(step): add `protonplus`

uwuclxdy:
### Adds support for `protonplus`, as it now offers [cli support](https://github.com/Vysp3r/ProtonPlus/pull/813).

image

### It also detects failures and prints out a hint to update it, as that's the most likely cause of the issue:
(that's what the issue is in this case, but the cause may not be obvious from ProtonPlus error message alone).

image

## Standards checklist

## For new steps

The `--yes` option has no effect to this step as the underlying command is `protoplus update all`, which always updates without confirmation; as designed, it should always be safe to do this.

### DIFF
--- a/src/steps/os/linux.rs
+++ b/src/steps/os/linux.rs
@@ -1175,12 +1175,12 @@ pub fn run_cinnamon_spices_updater(ctx: &ExecutionContext) -> Result<()> {
 pub fn run_protonplus_update(ctx: &ExecutionContext) -> Result<()> {
     let protonplus = require("protonplus")?;
 
-    let check = ctx.execute(&protonplus).args(["invalidarg67"]).output_checked();
-    if let Err(e) = &check {
-        if let Some(TopgradeError::ProcessFailedWithOutput(_, _, stderr)) = e.downcast_ref() {
-            if stderr.contains("This application can not open files") {
-                return Err(SkipStep("Updates unsupported for ProtonPlus versions under v0.5.18".to_string()).into());
-            }
+    if let Err(e) = ctx.execute(&protonplus).args(["invalidarg67"]).output_checked() {
+        if matches!(
+            e.downcast_ref(),
+            Some(TopgradeError::ProcessFailedWithOutput(_, _, stderr)) if stderr.contains("This application can not open files")
+        ) {
+            return Err(SkipStep("Updates unsupported for ProtonPlus versions under v0.5.18".to_string()).into());
         }
     }
 


src/steps/os/linux.rs
```diff
@@ -1172,6 +1172,30 @@ pub fn run_cinnamon_spices_updater(ctx: &ExecutionContext) -> Result<()> {
...
+    match cmd.status_checked() {
+        Ok(()) => Ok(()),
+        Err(e) => {
+            if let Some(TopgradeError::ProcessFailed(_, _)) = e.downcast_ref() {
+                println!();
+                println!(
+                    "{}",
+                    t!("Make sure ProtonPlus version is at least v0.5.17 and try again.")
+                );
+                return Err(e);
```
GideonBear: Unnecessary
```suggestion
```

src/steps/os/linux.rs
```diff
@@ -1172,6 +1172,30 @@ pub fn run_cinnamon_spices_updater(ctx: &ExecutionContext) -> Result<()> {
...
+    cmd.args(["update", "all"]);
+
+    match cmd.status_checked() {
+        Ok(()) => Ok(()),
+        Err(e) => {
+            if let Some(TopgradeError::ProcessFailed(_, _)) = e.downcast_ref() {
+                println!();
+                println!(
+                    "{}",
+                    t!("Make sure ProtonPlus version is at least v0.5.17 and try again.")
```
GideonBear: This is not ideal since it might take a while for packaged versions to update. Usually we parse `--version` but that also doesn't really work since `version` was also added in this release. Can you instead capture `stderr` and check for the error message (`This application cannot open files`), if it's found skip the step silently?

uwuclxdy: oh yeah skipping silently until its updated instead of always failing and potentially forcing users to add it to skipped steps makes way more sense! thanks

src/steps/os/linux.rs
```diff
@@ -1172,6 +1172,27 @@ pub fn run_cinnamon_spices_updater(ctx: &ExecutionContext) -> Result<()> {
...
 
+pub fn run_protonplus_update(ctx: &ExecutionContext) -> Result<()> {
+    let protonplus = require("protonplus")?;
+
+    print_separator("ProtonPlus");
+
+    let mut cmd = ctx.execute(protonplus);
+    cmd.args(["update", "all"]);
+
+    match cmd.output_checked_with(|o| if o.status.success() { Ok(()) } else { Err(()) }) {
```
GideonBear: Was there a reason for this?
`output_checked` is literally defined as:
```rs
fn output_checked(&mut self) -> eyre::Result<Output> {
    self.output_checked_with(|output: &Output| if output.status.success() { Ok(()) } else { Err(()) })
}
```
```suggestion
    match cmd.output_checked() {
```

src/steps/os/linux.rs
```diff
@@ -1172,6 +1172,27 @@ pub fn run_cinnamon_spices_updater(ctx: &ExecutionContext) -> Result<()> {
...
+        Ok(_) => Ok(()),
+        Err(e) => {
+            if let Some(TopgradeError::ProcessFailedWithOutput(_, _, stderr)) = e.downcast_ref() {
+                if stderr.contains("This application can not open files") {
+                    return Err(SkipStep(String::new()).into());
```
GideonBear: Add an error message here (similar to the one you used previously)

GideonBear (changes requested):
Can you confirm you've tested with both new and old versions on this commit?

src/steps/os/linux.rs
uwuclxdy: stupid mistake sorry

uwuclxdy:
> Can you confirm you've tested with both new and old versions on this commit?

while it does technically work, i'd still want to make it not print the separator if skipped. (the first is outdated bin)
also, there's no command output.

uwuclxdy:
got it working like it should! tested old bin, new and missing

uwuclxdy:
basically just run with an invalid argument (invalid for all versions, can we pls keep it like that) and the program will immediately error but the versions without cli support (eg. update support) will have the string we're looking for in the output :D

GideonBear (approved):
Awesome, thanks!

GideonBear:
> can we pls keep it like that

As long as you're the one in the git blame ;)

uwuclxdy:
error message is wrong, should say `.17` not `.18`

Issue #1471, opened 2025-11-11
Add a "sudo loop"


atoerien:
## I want to suggest some general feature

By default, sudo caches credentials for 5 minutes. This means that if a step that uses sudo runs for longer than 5 minutes, the credentials will have expired when the next step that needs sudo runs, causing the user to be prompted for a password again.

The solution is to run `sudo -v` in a background thread every ~4 minutes - this is what yay and paru do. There are [security risks](https://www.reddit.com/r/archlinux/comments/dnxjbx/comment/f5harho/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button) with this though, so it should be an option which is *not* enabled by default.

This would supersede #784.

PR #1803, opened 2026-02-27
feat: add sudo loop

uwuclxdy:
## What does this PR do

Implements sudo loop in a background thread to avoid password re-prompts. Optionally enabled with `--sudoloop` and `--sudoloop-interval [SECONDS]` or via config.toml:

```toml
sudo_loop = true
sudo_loop_interval = 240
```

## Standards checklist

Fixes https://github.com/topgrade-rs/topgrade/issues/1471

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -19,6 +19,8 @@
 # pre_sudo = false
 
 # Periodically runs `sudo -v` or `please -w` to avoid password re-prompts during updates (default: false)
+# WARNING: This is a potential security risk; if you walk away from the computer while topgrade is running,
+# another person can come by, CTRL+C, and gain access to a sudo session.
 # sudo_loop = true
 # sudo_loop_interval = 240 # 4 minutes (240s); default if sudo_loop enabled
 


src/execution_context.rs
```diff
@@ -42,6 +42,16 @@ impl RunType {
             RunType::Damp => false,
         }
     }
+
```
GideonBear, line 45: What is this for?

src/main.rs
```diff
@@ -333,6 +336,30 @@ fn run() -> Result<()> {
     }
 }
 
+fn spawn_sudo_loop(ctx: &execution_context::ExecutionContext, config: &Config) -> Option<std::sync::mpsc::Sender<()>> {
+    let sudo = ctx.sudo().as_ref().filter(|_| config.sudo_loop())?.clone();
```
GideonBear: Unnecessarily using `.filter` (I think)
```suggestion
    if !config.sudo_loop() { return None; }
    let sudo = ctx.sudo()?.as_ref().clone();
```

config.example.toml
```diff
@@ -18,6 +18,10 @@
 # (default: false)
 # pre_sudo = false
 
+# Periodically runs `sudo -v` to avoid password re-prompts during updates (default: false)
```
GideonBear:
```suggestion
# Periodically runs `sudo -v` or `please -w` to avoid password re-prompts during updates (default: false)
```

GideonBear (changes requested):
Nits, looks great!

GideonBear:
I tested it with:
```toml
[commands]
"one" = "sudo echo start"
"two" = "sleep 960"
"three" = "sudo echo done"
```
```bash
cargo run -- --only custom_commands --custom-commands one --custom-commands two --custom-commands three
```
* Without sudo-loop enabled: `three` asked for sudo password again
* With `sudo_loop = true`: `three` didn't ask for sudo password :tada:

So it works

src/execution_context.rs
uwuclxdy: I moved `execute()` logic to `RunType` because `ExecutionContext<'a>` borrows `Config` and therefore can't be moved into a background thread (where `sudo -v` runs). Meanwhile, `RunType` is just a `Copy` enum, which allows this.

config.example.toml
```diff
@@ -18,6 +18,10 @@
 # (default: false)
 # pre_sudo = false
 
+# Periodically runs `sudo -v` or `please -w` to avoid password re-prompts during updates (default: false)
```
GideonBear, line 21: Something like:
```suggestion
# Periodically runs `sudo -v` or `please -w` to avoid password re-prompts during updates (default: false)
# WARNING: This poses serious security risks. Please be aware of these risks when you enable this.
```

GideonBear (approved):

Issue #1020, opened 2025-01-21
Add support for upgrading plasmoids


CavaleriOmar:
I currently use apdatifier (https://github.com/exequtic/apdatifier) to upgrade my system (EndeavourOS), which allows to upgrade arch packages, flatpaks and KDE plasmoids packages in one click...

Know possible to add such feature also with TopGrade? (Or if it is already possible, how to configure it)?

phnx47:
I also use Apdatifier and Topgrade because Topgrade does not upgrade [Plasma Widgets(Plasmoids)](https://store.kde.org/browse?cat=705).

This is how it is implemented in Apdatifier: https://github.com/exequtic/apdatifier/blob/main/package/contents/tools/sh/widgets. I'm happy to help, but I don't have much experience with Rust.

GideonBear:
> I also use Apdatifier and Topgrade because Topgrade does not upgrade [Plasma Widgets(Plasmoids)](https://store.kde.org/browse?cat=705).
> 
> This is how it is implemented in Apdatifier: [exequtic/apdatifier@`main`/package/contents/tools/sh/widgets](https://github.com/exequtic/apdatifier/blob/main/package/contents/tools/sh/widgets?rgh-link-date=2025-06-15T16%3A07%3A41.000Z). I'm happy to help, but I don't have much experience with Rust.

If it's implemented in shell in there, it's certainly possible to be implemented in Topgrade.
@SteveLauC I think there are multiple paths forward here:
1. Port the Plasmoid updating code to Rust (we can do an almost 1-to-1 ai-assisted translation), and watch that repo for any updates so that they can be ported too.
2. Invoke `apdatifier` to update widgets when `apdatifier` is installed. (This can also be a temporary solution until 1. is implemented)
3. We could in theory also bundle the shell code with Topgrade and run that, but implementing it in Rust would be a lot better long-term.

What is your preference?

uwuclxdy:
i'm working on [`plasmoid-updater`](https://github.com/uwuclxdy/plasmoid-updater), should be integrated into topgrade soon™

Trollwut:
Do we have news on that? Would really like if topgrade would take care of plasmoids as well :)

uwuclxdy:
Yea well, from my testing it works fine... I told @GideonBear to wait with the code audit before I make some changes, but I guess it's fine as it is now... 

[...]

PR #1804, opened 2026-02-28
feat(step): update KDE Plasmoids using libplasmoid-updater

uwuclxdy:
## What does this PR do

Adds integration with `libplasmoid-updater` to update KDE Plasma components (plasmoids) user and system-wide.
Fixes https://github.com/topgrade-rs/topgrade/issues/1020

It can also be used as a standalone program for those that wish to test (or just use) the library itself: https://github.com/uwuclxdy/plasmoid-updater

## Standards checklist

## For new steps

- [x] *Optional:* Topgrade skips this step where needed (**needs testing**)
  the underlying command

My best try at this ig. Needs testing (edge cases, error handling). As far as library is concerned, the code is a messy and worse I'd like it to be, but it works! It successfully finds and updates components that either Discover or Apdatify sometimes doesn't find.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,6 +74,9 @@ assets = [
   ["deployment/deb/_topgrade", "usr/share/zsh/vendor-completions/", "644"],
 ]
 
+[target.'cfg(target_os = "linux")'.dependencies]
+libplasmoid-updater = { version = "0.1.4", features = ["cli"] }
+
 [target.'cfg(unix)'.dependencies]
 nix = { version = "0.31", features = ["hostname", "signal", "user"] }
 rust-ini = "0.21"
--- a/src/steps/os/linux.rs
+++ b/src/steps/os/linux.rs
@@ -2,13 +2,14 @@ use std::path::{Path, PathBuf};
 
 use color_eyre::eyre::Result;
 use ini::Ini;
+use libplasmoid_updater::Config as PlasmoidConfig;
 use rust_i18n::t;
 use tracing::{debug, warn};
 
 use crate::HOME_DIR;
 use crate::command::CommandExt;
 use crate::config::NixHandler;
-use crate::error::{SkipStep, TopgradeError};
+use crate::error::{SkipStep, StepFailed, TopgradeError};
 use crate::execution_context::ExecutionContext;
 use crate::step::Step;
 use crate::steps::generic::is_wsl;
@@ -1216,6 +1217,44 @@ pub fn run_protonplus_update(ctx: &ExecutionContext) -> Result<()> {
     print_separator("ProtonPlus");
 
     ctx.execute(&protonplus).args(["update", "all"]).status_checked()
+}
+
+pub fn run_plasmoid_updater(ctx: &ExecutionContext, system: bool, step_name: &str) -> Result<()> {
+    let with_yes = ctx
+        .config()
+        .yes(if system { Step::PlasmoidsSystem } else { Step::Plasmoids });
+    let excluded = ctx.config().plasmoids_get_excluded(system);
+    let config = PlasmoidConfig::new()
+        .with_system(system)
+        .with_excluded_packages(excluded)
+        .with_yes(with_yes);
+
+    print_separator(step_name);
+
+    if ctx.run_type().dry() {
+        println!("Dry running plasmoid-updater");
+        return Ok(());
+    }
+
+    match libplasmoid_updater::update(&config) {
+        Err(e) => {
+            println!(
+                "{}",
+                t!("plasmoid-updater encountered an unexpected error during updating:")
+            );
+            println!("{e:#?}");
+            Err(StepFailed.into())
+        }
+        Ok(result) => {
+            if result.has_failures() {
+                println!("{}", t!("plasmoid-updater failed to update some components:"));
+                result.print_error_table();
+                Err(StepFailed.into())
+            } else {
+                Ok(())
+            }
+        }
+    }
 }
 
 #[cfg(test)]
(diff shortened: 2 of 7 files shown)


src/steps/os/linux.rs
```diff
@@ -1172,6 +1172,68 @@ pub fn run_cinnamon_spices_updater(ctx: &ExecutionContext) -> Result<()> {
...
+            println!("{e:#?}");
+            return Err(StepFailed.into());
+        }
+        Ok(result) => {
+            if result.available_updates.is_empty() {
+                // libplasmoid-updater already prints "no updates available" by this point
+                return Ok(());
+            }
+        }
+    };
```
uwuclxdy: could probably remove this because `update()` checks for updates and prints the same "no updates available."

we're basically doing the same thing twice if updates are found.

GideonBear:
FYI the MSRV check in CI will fail because of the current MSRV of 1.84 which will be bumped to 1.88 in v17 (which I will release soon, I hope). You don't have to change the edition and you can ignore the MSRV check CI for now.
`plasmoid-updater` does compile with Rust 1.88, I checked

uwuclxdy:
@GideonBear yeah, i fixed the ordering problem tho. they were ordered but not in the right place in proportion to the entire enum

uwuclxdy:
**before:**
https://asciinema.org/a/5yUWkvgGfgczGTs5

**after:**
https://asciinema.org/a/vJeRALgInwaWbQR2

Cargo.toml
```diff
@@ -80,6 +80,9 @@ assets = [
   ["deployment/deb/_topgrade", "usr/share/zsh/vendor-completions/", "644"],
 ]
 
+[target.'cfg(target_os = "linux")'.dependencies]
+libplasmoid-updater = { version = "0.1.3", features = ["cli"] }
```
GideonBear: Why the CLI feature?

src/steps/os/linux.rs
```diff
@@ -1172,6 +1172,46 @@ pub fn run_cinnamon_spices_updater(ctx: &ExecutionContext) -> Result<()> {
     ctx.execute(cinnamon_spice_updater).arg("--update-all").status_checked()
 }
 
+#[cfg(target_os = "linux")]
```
GideonBear: Unnecessary since the whole module is under `#[cfg(target_os = "linux")]` in `mod.rs`

src/steps/os/linux.rs
```diff
@@ -1172,6 +1172,46 @@ pub fn run_cinnamon_spices_updater(ctx: &ExecutionContext) -> Result<()> {
...
+pub fn run_plasmoid_updater(ctx: &ExecutionContext, system: bool, step_name: &str) -> Result<()> {
+    use crate::error::StepFailed;
+    use libplasmoid_updater::{update, Config as PlasmoidConfig};
```
GideonBear: Can these just be global imports?

src/steps/os/linux.rs
```diff
@@ -1172,6 +1172,46 @@ pub fn run_cinnamon_spices_updater(ctx: &ExecutionContext) -> Result<()> {
...
+
+    let with_yes = ctx.config().yes(Step::Plasmoids) || ctx.config().yes(Step::PlasmoidsSystem);
```
GideonBear: Doesn't take into account `system`?

src/steps/os/linux.rs
```diff
@@ -1172,6 +1172,46 @@ pub fn run_cinnamon_spices_updater(ctx: &ExecutionContext) -> Result<()> {
...
+    if ctx.run_type().dry() {
+        println!("Dry running plasmoid-updater");
+        return Ok(());
+    }
+
+    match update(&config) {
+        Err(e) => {
+            println!(
+                "{}",
+                t!("plasmoid-updater encountered an unexpected error during updating:")
```
GideonBear, line 1243: These are not in the locale file yet. Why doesn't CI complain??

Cargo.toml
uwuclxdy: printing to console from library crate is gated with cli feature (if someone else wished to implement a frontend for it, like a gui or something)

GideonBear: Oh in that way, it's not "this is a cli" but "this can be used for a cli". Got it!

src/steps/os/linux.rs
uwuclxdy: doesn't build without this on other targets (check CI run from previous commit(s))

src/steps/os/linux.rs
uwuclxdy: wdym? `ctx.config().yes(Step::Plasmoids) || ctx.config().yes(Step::PlasmoidsSystem)`

uwuclxdy: ah, i see the issue, will fix

src/steps/os/linux.rs
uwuclxdy: nvm you're right

src/steps/os/linux.rs
```diff
@@ -2,12 +2,13 @@ use std::path::{Path, PathBuf};
 
 use color_eyre::eyre::Result;
 use ini::Ini;
+use libplasmoid_updater::{update, Config as PlasmoidConfig};
```
GideonBear: eh actually can you just use `libplasmoid_updater::update` directly? Global `update` is not good

GideonBear (changes requested):

src/steps/os/linux.rs
uwuclxdy: makes sense

Issue #1244, opened 2025-07-30
Step request: Gear Lever (AppImage manager)


zero-ninetynine:
I use Gear Lever ([Github](https://github.com/mijorus/gearlever)) to manage my AppImages. It's primarily used to easily add appimages to a central location and automatically create .desktop files, as well as other management operations. However, it also has a feature where either the AppImage's developer or the user can specify an update url (either a static url or a release from github, gitlab and other sites) for each AppImage and Gear Lever can update them. I think it'd be useful to be able to update AppImages along with everything else using topgrade.

The tool is flatpak only, so as far as I can tell it should work on any linux distribution that supports flatpaks. It also has a cli which can be used to update the managed AppImages from the command line, which topgrade could use. It's invoked as follows to upgrade all AppImages: `flatpak run it.mijorus.gearlever --update --all`. The command shows a yes/no confirmation prompt (oddly enough it does so even if there's nothing to update) which can be skipped using a `--yes` option. An AppImage name can also be provided instead of the `--all` option to update a single AppImage. It does not appear to have a `--dry-run` or similar option.

I've been using topgrade for a while and it is incredibly convenient, so thanks for all the work!

GideonBear:
Sounds simple enough, the only challenge is checking if it's installed, a simple `require` won't do. I don't use flatpak, is there a command to check whether it is installed?

zero-ninetynine:
Thanks for the quick response! Poking around, flatpak has an `info` command that errors if the package id provided is not installed and provides info if it is. For example:
- Installed package:
```
$ flatpak info it.mijorus.gearlever

Gear Lever - Manage AppImages

          ID: it.mijorus.gearlever
         Ref: app/it.mijorus.gearlever/x86_64/stable
        Arch: x86_64
      Branch: stable
     Version: 3.3.4
     License: GPL-3.0-or-later
      Origin: flathub
  Collection: org.flathub.Stable
Installation: system
   Installed: 21,2 MB
     Runtime: org.gnome.Platform/x86_64/47
         Sdk: org.gnome.Sdk/x86_64/47

      Commit: a88ca3469207cd1ee8dea91aefaee3ce2f6af9ff60d12879afc10d555c9b44c3
      Parent: d327a95165ec779c57ff142bc9decff41c381c07565126cb36c54c56ad82c505
     Subject: Merge pull request #117 from flathub/mijorus-patch-11 (d32e19d511d4)
        Date: 2025-07-04 09:24:18 +0000
[...]

PR #1805, opened 2026-02-28
NOSQUASH: new step (Gear Lever) and `require_flatpak` helper

uwuclxdy:
## What does this PR do

Fixes https://github.com/topgrade-rs/topgrade/issues/1244 and adds Gear Lever + `require_flatpak()` function as discussed there.

image
image

**i've canceled it mid run on purpose to show what command is ran (automatically, it uses flatpak as fallback if `require()` doesn't find packaged install).**

as an example of it completing successfully:
image

## Standards checklist

## For new steps

  the underlying command

### DIFF
--- a/src/steps/os/linux.rs
+++ b/src/steps/os/linux.rs
@@ -15,7 +15,7 @@ use crate::steps::os::archlinux;
 use crate::steps::unix::{can_nh_switch, nh_switch, NhSwitchArgs};
 use crate::sudo::SudoExecuteOpts;
 use crate::terminal::{print_separator, prompt_yesno};
-use crate::utils::{require, require_one, which, PathExt};
+use crate::utils::{require, require_flatpak, require_one, which, PathExt};
 use crate::HOME_DIR;
 
 static OS_RELEASE_PATH: &str = "/etc/os-release";
@@ -1162,6 +1162,23 @@ pub fn run_auto_cpufreq(ctx: &ExecutionContext) -> Result<()> {
 
     let sudo = ctx.require_sudo()?;
     sudo.execute(ctx, &auto_cpu_freq)?.arg("--update").status_checked()
+}
+
+pub fn run_gearlever(ctx: &ExecutionContext) -> Result<()> {
+    let (mut cmd, native) = match require("gearlever") {
+        Ok(p) => (ctx.execute(p), true),
+        Err(_) => (require_flatpak(ctx, "it.mijorus.gearlever")?, false),
+    };
+
+    print_separator(if native { "Gear Lever" } else { "Gear Lever (Flatpak)" });
+
+    cmd.args(["--update", "--all"]);
+
+    if ctx.config().yes(Step::Gearlever) {
+        cmd.arg("--yes");
+    }
+
+    cmd.status_checked()
 }
 
 pub fn run_cinnamon_spices_updater(ctx: &ExecutionContext) -> Result<()> {
--- a/src/step.rs
+++ b/src/step.rs
@@ -62,6 +62,7 @@ pub enum Step {
     Flutter,
     Fossil,
     Gcloud,
+    Gearlever,
     Gem,
     Ghcup,
     GitRepos,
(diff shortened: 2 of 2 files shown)


GideonBear:
CI failure is because `require_flatpak` is only used on Linux. You can add `#[allow(unused)]` to fix it, see suggestion.

src/utils.rs
```diff
@@ -103,6 +103,25 @@ pub fn require<T: AsRef<OsStr> + Debug>(binary_name: T) -> Result<PathBuf> {
...
+        .always()
+        .args(["info", name])
+        .output_checked_with(|output| {
+            debug!("flatpak info exited with status: {:?}", output.status);
+            Ok(())
+        });
+
+    match result {
+        Ok(output) if output.status.success() => {
+            debug!("Flatpak {:?} is installed", name);
```
```suggestion
            debug!("Flatpak {name:?} is installed");
```

src/utils.rs
```diff
@@ -103,6 +103,25 @@ pub fn require<T: AsRef<OsStr> + Debug>(binary_name: T) -> Result<PathBuf> {
...
+        .execute("flatpak")
+        .always()
+        .args(["info", name])
+        .output_checked_with(|output| {
+            debug!("flatpak info exited with status: {:?}", output.status);
+            Ok(())
+        });
+
+    match result {
+        Ok(output) if output.status.success() => {
```
GideonBear: Since we used `output_checked`, any non-success exit codes will result in `Err` (IIRC)
```suggestion
        Ok(output) => {
```

src/utils.rs
```diff
@@ -103,6 +103,25 @@ pub fn require<T: AsRef<OsStr> + Debug>(binary_name: T) -> Result<PathBuf> {
     }
 }
 
+pub fn require_flatpak(ctx: &ExecutionContext, name: &str) -> Result<()> {
```
```suggestion
#[allow(unused)]
pub fn require_flatpak(ctx: &ExecutionContext, name: &str) -> Result<()> {
```

src/utils.rs
```diff
@@ -103,6 +103,25 @@ pub fn require<T: AsRef<OsStr> + Debug>(binary_name: T) -> Result<PathBuf> {
...
+            debug!("flatpak info exited with status: {:?}", output.status);
+            Ok(())
+        });
+
+    match result {
+        Ok(output) if output.status.success() => {
+            debug!("Flatpak {:?} is installed", name);
+            Ok(())
+        }
+        _ => Err(SkipStep(t!("Flatpak {name:?} is not installed", name = name).to_string()).into()),
```
GideonBear: Not added to the locale file, see also `check_locale` CI failure.

src/utils.rs
```diff
@@ -103,6 +103,25 @@ pub fn require<T: AsRef<OsStr> + Debug>(binary_name: T) -> Result<PathBuf> {
     }
 }
 
+pub fn require_flatpak(ctx: &ExecutionContext, name: &str) -> Result<()> {
+    let result = ctx
+        .execute("flatpak")
```
GideonBear: `require("flatpak")?` first, it might not be installed

src/steps/os/linux.rs
```diff
@@ -1164,6 +1164,31 @@ pub fn run_auto_cpufreq(ctx: &ExecutionContext) -> Result<()> {
...
+pub fn run_gearlever(ctx: &ExecutionContext) -> Result<()> {
+    let native = require("gearlever").is_ok();
+    if !native {
+        require_flatpak(ctx, "it.mijorus.gearlever")?;
+    }
+
+    print_separator(if native { "Gear Lever" } else { "Gear Lever (Flatpak)" });
+
+    let mut cmd = if native {
+        ctx.execute("gearlever")
```
GideonBear: Avoid this, use the result of `require` instead.
Same for the flatpak call below

src/steps/os/linux.rs
```diff
@@ -1164,6 +1164,31 @@ pub fn run_auto_cpufreq(ctx: &ExecutionContext) -> Result<()> {
     sudo.execute(ctx, &auto_cpu_freq)?.arg("--update").status_checked()
 }
 
+pub fn run_gearlever(ctx: &ExecutionContext) -> Result<()> {
+    let native = require("gearlever").is_ok();
+    if !native {
+        require_flatpak(ctx, "it.mijorus.gearlever")?;
+    }
```
GideonBear: Given that you still need the path how about (with AI inspiration):
```rust
let (cmd, native) = match require("gearlever") {
    Ok(p) => (ctx.execute(p), true),
    Err(_) => (require_flatpak(ctx, "it.mijorus.gearlever")?, false),
};
```
And then make `require_flatpak` build and return the command `flatpak run it.mijorus.gearlever`?

GideonBear:
I want to merge the commits separately (that's why I added `NOSQUASH` to the title). If you can, you can edit the commits with rebases. If you don't know how, please add the suggestions in different commits (e.g. the ones on the require_flatpak function separately from the ones on the gearlever step) and then I can rebase them myself.

src/utils.rs
uwuclxdy: sorry, forgot these strings aren't localized

keunes:
Thanks both for this addition!

PR #1841, opened 2026-03-10
fix: skip Gear Lever update prompt when no updates exist

YuF-9468:
## Summary
- probe Gear Lever with `--list-updates` before launching the interactive update flow
- skip the step when Gear Lever reports that nothing needs updating
- add focused tests for the output parser

## Testing
- Not run locally in this environment (`cargo` is unavailable on the runner)

### DIFF
--- a/src/steps/os/linux.rs
+++ b/src/steps/os/linux.rs
@@ -1179,7 +1179,7 @@ pub fn run_auto_cpufreq(ctx: &ExecutionContext) -> Result<()> {
 fn gearlever_has_pending_updates(output: &str) -> bool {
     let trimmed = output.trim();
 
-    !trimmed.is_empty() && !trimmed.eq_ignore_ascii_case("No updates available")
+    !trimmed.is_empty() && trimmed != "No updates available" && trimmed != "No updates available."
 }
 
 pub fn run_gearlever(ctx: &ExecutionContext) -> Result<()> {
@@ -1202,7 +1202,11 @@ pub fn run_gearlever(ctx: &ExecutionContext) -> Result<()> {
     list_updates.arg("--list-updates");
 
     let list_updates = list_updates.output_checked_utf8()?;
-    let update_candidates = format!("{}\n{}", list_updates.stdout, list_updates.stderr);
+    let update_candidates = if list_updates.stdout.trim().is_empty() {
+        list_updates.stderr
+    } else {
+        list_updates.stdout
+    };
 
     if !gearlever_has_pending_updates(&update_candidates) {
         println!("No updates available.");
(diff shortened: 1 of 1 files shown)


uwuclxdy:
Can confirm the issue
image

Really bad design from Gear Lever, I think Topgrade should try to work around that like suggested in this PR. I tested it myself and it works:

### before
image

### after
image

1. I think that if the step runs (after `requires`) then it should be shown in Summary as OK - even if no updates.
2. Topgrade should print "No updates available."

I am also to blame for missing this when adding Gear Lever step, because I always use `assume_yes = true`, sorry.

src/steps/os/linux.rs
```diff
@@ -1176,14 +1176,46 @@ pub fn run_auto_cpufreq(ctx: &ExecutionContext) -> Result<()> {
     sudo.execute(ctx, &auto_cpu_freq)?.arg("--update").status_checked()
 }
 
+fn gearlever_has_pending_updates(output: &str) -> bool {
+    let trimmed = output.trim();
+
+    !trimmed.is_empty() && !trimmed.eq_ignore_ascii_case("No updates available")
```
GideonBear: Is it necessary to ignore case?

src/steps/os/linux.rs
```diff
@@ -1176,14 +1176,46 @@ pub fn run_auto_cpufreq(ctx: &ExecutionContext) -> Result<()> {
...
+    } else {
+        let flatpak = require("flatpak")?;
+        let mut cmd = ctx.execute(&flatpak).always();
+        cmd.args(["run", "it.mijorus.gearlever"]);
+        cmd
+    };
+    list_updates.arg("--list-updates");
+
+    let list_updates = list_updates.output_checked_utf8()?;
+    let update_candidates = format!("{}\n{}", list_updates.stdout, list_updates.stderr);
```
GideonBear: Please check which it is and check only one

src/steps/os/linux.rs
```diff
@@ -1176,14 +1176,46 @@ pub fn run_auto_cpufreq(ctx: &ExecutionContext) -> Result<()> {
...
+
+    if !gearlever_has_pending_updates(&update_candidates) {
+        return Err(SkipStep("No Gear Lever updates available".to_string()).into());
```
GideonBear: `SkipStep` is incorrect here. We didn't skip the step, we just didn't find any updates. Please write stdout and stderr of the `--list-updates` output and `return Ok(())`.

src/steps/os/linux.rs
```diff
@@ -1284,6 +1316,22 @@ mod tests {
         test_template(include_str!("os_release/oracle"), Distribution::CentOS);
     }
 
+    #[test]
```
GideonBear: Insert below and not between `os_release` tests

src/steps/os/linux.rs
YuF-9468: Good point — I removed the case-insensitive match and now only treat exact "No updates available" (with/without trailing period) as the no-update sentinel.

src/steps/os/linux.rs
YuF-9468: Adjusted to check only one stream: stdout by default, and stderr only when stdout is empty.

src/steps/os/linux.rs
YuF-9468: Now prints "No updates available." and returns Ok; also switched to checking only one stream for the sentinel as discussed.

src/steps/os/linux.rs
YuF-9468: Moved the Gear Lever tests below the os_release test block as requested.

GideonBear:
Looks like you rebased `main` onto `fix-1838` instead of the other way around?

GideonBear:
Nvm don't bother; I will extract the changes in a different PR since I want to make some changes anyway.

YuF-9468:
Thanks for the confirmation and the helpful notes — much appreciated!

PR #1846, opened 2026-03-11
fix(protonplus): fix display name capitalization

uwuclxdy:
Fixes capitalisation in summary for ProtonPlus. Maybe some other steps need this fix as well, lmk if I should check and fix them too.


GideonBear (approved):

GideonBear:
Nah I want to formalize the key/display name anyway. Another thing for v18 :S

PR #1865, opened 2026-03-17
feat(step): add Microsoft Office auto update on macOS

vrioux:
## What does this PR do
Adds a new MicrosoftOffice step that updates Microsoft Office apps on macOS using the msupdate CLI from Microsoft AutoUpdate (MAU) on macOS. 

## Standards checklist

### AI involvement
I consulted an AI to help me understand the code base, and to suggest the simplest approach for this. I also used AI to help test the solution.

## For new steps

- [X] *Optional:* The `--dry-run` option works with this step
- [X] *Optional:* The `--yes` option works with this step if it is supported by the underlying command

### DIFF
--- a/src/steps/os/macos.rs
+++ b/src/steps/os/macos.rs
@@ -46,8 +46,8 @@ pub fn run_microsoft_office(ctx: &ExecutionContext) -> Result<()> {
     debug!("msupdate --list output: {:?}", output);
 
     if output.stdout.contains("No updates available") {
-        io::stdout().write_all(&output.stdout)?;
-        io::stderr().write_all(&output.stderr)?;
+        io::stdout().write_all(output.stdout.as_bytes())?;
+        io::stderr().write_all(output.stderr.as_bytes())?;
         return Ok(());
     }
 


src/steps/os/macos.rs
```diff
@@ -34,6 +35,48 @@ pub fn run_mas(ctx: &ExecutionContext) -> Result<()> {
...
+        return Ok(());
+    }
+
+    // Install updates, waiting up to 600 seconds for completion
+    ctx.execute(&msupdate)
+        .args(["--install", "--wait", "600"])
+        .status_checked()
+}
+
+fn find_msupdate() -> Result<PathBuf> {
```
GideonBear, line 60: Isn't it just on path? If it is, just use `require("msupdate")`

src/steps/os/macos.rs
```diff
@@ -34,6 +35,48 @@ pub fn run_mas(ctx: &ExecutionContext) -> Result<()> {
...
+    let output = ctx
+        .execute(&msupdate)
+        .always()
+        .arg("--list")
+        .output_checked_utf8()?;
+
+    debug!("msupdate --list output: {:?}", output);
+
+    if output.stdout.contains("No updates available") {
+        println!("{}", t!("No new software available."));
```
GideonBear:
```suggestion
        io::stdout().write_all(&output.stdout)?;
        io::stderr().write_all(&output.stderr)?;
```

src/steps/os/macos.rs
vrioux: It's not on path, at least on my system.

GideonBear:
(tip if you might not know: you can just run `cargo fmt` to do all the formatting automatically)

GideonBear:
(just need to `use std::io`)

src/steps/os/macos.rs
vrioux: Oops, I think we also need to use std::io::{self, Write};

PR #1875, opened 2026-03-23
refactor: remove --no-self-update in favor of --disable self_update

### LINKED ISSUE #1828: Remove `--no-self-update` and `no_self_update` in favor of `--disable self_update`
This flag/config option was added in #388, but after #585, should not be necessary anymore.

mvanhorn:
## Summary

Removes the `--no-self-update` CLI flag and `no_self_update` config file option. Since #585 added `--disable self_update`, the dedicated flag is redundant.

## Changes

- `src/config.rs`: removed `no_self_update` from `CommandLineArgs`, `Misc` config struct, and the `Config::no_self_update()` method
- `src/step.rs`: removed the `!ctx.config().no_self_update()` condition from the `SelfUpdate` step handler (the step is already excluded when disabled via `--disable`)
- `src/steps/toolbx.rs`: replaced `"--no-self-update"` with `"--disable", "self_update"` in the subprocess invocation

The `TOPGRADE_NO_SELF_UPGRADE` env var check is preserved as it serves a different purpose (preventing recursive self-updates).

## Testing

`cargo build` passes.

Closes #1828

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -89,7 +89,6 @@
 # Run specific steps - same options as the command line flag
 # only = ["system", "emacs"]
 
-
 # Extra tracing filter directives
 # These are prepended to the `--log-filter` argument
 # See: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html#directives


GideonBear:
Please format, and also remove the option from `config.example.toml`

mvanhorn:
Fixed in e2ef64e: ran `cargo fmt` and removed the `no_self_update` block from `config.example.toml`.

PR #1879, opened 2026-03-24
feat(claude code): add plugin updates

uwuclxdy:
## What does this PR do
Adds updating claude code plugins in addition to self-update. Maybe we should add opt-out config option, since they started testing auto plugin updates at startup, however that feature is opt-in and increases startup time (so doing it with topgrade may still be better after official release).

image

---

Also, this is the cleanest way to do it, because:
```
$ claude plugin update --all
error: unknown option '--all'
```

## Standards checklist

### AI involvement
Inline hints (Copilot) and Claude Code for refactor before commit.

### DIFF
--- a/src/steps/generic.rs
+++ b/src/steps/generic.rs
@@ -2023,6 +2023,7 @@ pub fn run_claude_code(ctx: &ExecutionContext) -> Result<()> {
     print_separator("Claude Code");
 
     ctx.execute(&claude).arg("update").status_checked()?;
+
     ctx.execute(&claude)
         .args(["plugin", "marketplace", "update"])
         .status_checked()?;


GideonBear:
> however that feature is opt-in and increases startup time (so doing it with topgrade may still be better after official release).

Yeah definitely, that is the Topgrade way ;)

GideonBear:
Is it just me or is `claude plugin` undocumented?
https://code.claude.com/docs/en/cli-reference

uwuclxdy:
Oh yea, it's definitely not there. I guess Claude cooks features faster than they can remember to update the docs haha

src/steps/generic.rs
```diff
@@ -2015,11 +2015,29 @@ pub fn run_typst(ctx: &ExecutionContext) -> Result<()> {
 }
 
 pub fn run_claude_code(ctx: &ExecutionContext) -> Result<()> {
+    static PLUGIN_RE: LazyLock<regex::Regex> =
+        LazyLock::new(|| regex::Regex::new(r"[[:alnum:]_.-]+@[[:alnum:]_.-]+").unwrap());
```
GideonBear: Nit: this is a PCRE-specific feature of regex, I would rather replace it with a generic version which is arguably more readable even
```suggestion
        LazyLock::new(|| regex::Regex::new(r"[a-zA-Z0-9_.-]+@[a-zA-Z0-9_.-]+").unwrap());
```

GideonBear:
> Oh yea, it's definitely not there. I guess Claude cooks features faster than they can remember to update the docs haha

(you make this sound like a positive thing... :P)

I wanted to check to make sure there's absolutely no way to update all the plugins at once. It seems weird to me if that's not a thing. Can you check all available flags for `claude plugin update`, and maybe `claude plugin update *` or something?

src/steps/generic.rs
```diff
@@ -2015,11 +2015,29 @@ pub fn run_typst(ctx: &ExecutionContext) -> Result<()> {
...
 
     print_separator("Claude Code");
 
-    ctx.execute(claude).arg("update").status_checked()
+    ctx.execute(&claude).arg("update").status_checked()?;
+    ctx.execute(&claude)
+        .args(["plugin", "marketplace", "update"])
+        .status_checked()?;
+
+    let output = ctx.execute(&claude).args(["plugins", "list"]).output()?;
```
```suggestion
    let output = ctx.execute(&claude).args(["plugins", "list"]).output_utf8_checked()?;
```
And then remove `let output: Utf8Output = output.try_into()?;`

src/steps/generic.rs
```diff
@@ -2015,11 +2015,29 @@ pub fn run_typst(ctx: &ExecutionContext) -> Result<()> {
...
 pub fn run_claude_code(ctx: &ExecutionContext) -> Result<()> {
+    static PLUGIN_RE: LazyLock<regex::Regex> =
+        LazyLock::new(|| regex::Regex::new(r"[[:alnum:]_.-]+@[[:alnum:]_.-]+").unwrap());
+
     let claude = require("claude")?;
 
     print_separator("Claude Code");
 
-    ctx.execute(claude).arg("update").status_checked()
+    ctx.execute(&claude).arg("update").status_checked()?;
```
GideonBear, line 2025:
```suggestion
    ctx.execute(&claude).arg("update").status_checked()?;

```
Less separation between self-update and marketplace update than marketplace update and plugins update

uwuclxdy:
> (you make this sound like a positive thing... :P)

My bad on the wording, I know that's not the way to do it lol.

> I wanted to check to make sure there's absolutely no way to update all the plugins at once. It seems weird to me if that's not a thing.

Yea, as I said, this feature exists but is currently opt-in via env flag (testing stage). It also slows down program startup.

> Can you check all available flags for `claude plugin update`, and maybe `claude plugin update *` or something?

I did this before trying to add it to topgrade 😅

GideonBear:
> Yea, as I said, this feature exists but is currently opt-in via env flag (testing stage). It also slows down program startup.

> I did this before trying to add it to topgrade 😅

Alr thanks, sorry!

uwuclxdy:
tested it again, still works

PR #1898, opened 2026-04-01
feat(step): add tpack

alonsodomin:
## What does this PR do

`tmux` OG plugin manager (`tpm`) [is dead](https://github.com/tmux-plugins/tpm/issues/318) and, as discussed in the linked issue, a couple of alternatives have showed up. This PR adds support for [`tpack`](https://github.com/tmuxpack/tpack), a modern drop-in replacement.

## Standards checklist

### AI involvement

I did not use AI at all.

## For new steps

  the underlying command

### DIFF
--- a/src/step.rs
+++ b/src/step.rs
@@ -653,9 +653,10 @@ impl Step {
             }
             Tldr => runner.execute(*self, "TLDR", || generic::run_tldr(ctx))?,
             Tlmgr => runner.execute(*self, "tlmgr", || generic::run_tlmgr_update(ctx))?,
-            Tmux => {
+            Tmux =>
+            {
                 #[cfg(unix)]
-                runner.execute(*self, "tmux", || tmux::run_tpm(ctx))?;
+                runner.execute(*self, "tmux", || tmux::run_tpm(ctx))?
             }
             Toolbx =>
             {


GideonBear:
Why does `tpm` use `tpm all` and `tpack` `tpack update all`?

src/step.rs
```diff
@@ -652,10 +652,11 @@ impl Step {
             }
             Tldr => runner.execute(*self, "TLDR", || generic::run_tldr(ctx))?,
             Tlmgr => runner.execute(*self, "tlmgr", || generic::run_tlmgr_update(ctx))?,
-            Tmux =>
-            {
+            Tmux => {
```
GideonBear: Please add a separate step called `Tpack`

alonsodomin:
Well,  `tpack` is a `tpm` replacement in the sense that as a user you don't need to change your `tmux.conf` file but they are different programs with different CLIs. In the case of `tpack` the syntax to trigger an update is `tpack update [plugin_name]` for a single plugin or `tpack update all` to update all of them.

src/step.rs
GideonBear: (it can still be located in `steps/tmux.rs`)

GideonBear:
> Well, `tpack` is a `tpm` replacement in the sense that as a user you don't need to change your `tmux.conf` file but they are different programs with different CLIs. In the case of `tpack` the syntax to trigger an update is `tpack update [plugin_name]` for a single plugin or `tpack update all` to update all of them.

Ah okay, that's fine but not what I would call a "drop-in replacement".

alonsodomin:
> Ah okay, that's fine but not what I would call a "drop-in replacement".

Well, the author of `tpack` made it such that it can be installed as a system package or git-cloned into tmux's plugins repo the same way `tpm` is (was?) - in that later case the repo provides with a shim that makes tpack behave like `tpm` with same cli.

So it _can_ be installed as a drop-in replacement, this PR is just to support the case it's been installed as a system package.

GideonBear:
> > Ah okay, that's fine but not what I would call a "drop-in replacement".
> 
> Well, the author of `tpack` made it such that it can be installed as a system package or git-cloned into tmux's plugins repo the same way `tpm` is (was?) - in that later case the repo provides with a shim that makes tpack behave like `tpm` with same cli.
> 
> So it _can_ be installed as a drop-in replacement, this PR is just to support the case it's been installed as a system package.

Oooh okay, that makes sense. Thanks for the clarification!

GideonBear (approved):
Thanks!

PR #1911, opened 2026-04-05
feat(gup): add exclude config

### LINKED ISSUE #604: Exclude specified bin(s) from 'gup update'

When [`gup update` is run](https://github.com/topgrade-rs/topgrade/blob/9ea6628b5ca8a38aebb74ffd5bfee4c0a7d5b550/src/steps/go.rs#L22-L28), it would be useful if I could [configure some binaries for its exclude flag](https://github.com/nao1215/gup#exclude-binaries-during-gup-update).

This would be useful because every so often there are some binaries that I have installed with `go install` to use a specific non-latest version.
I do not want them to be automatically brought up to date with the latest version, and want to stay pinned on the specific version that I have installed at the moment.

## More information

In the config.toml file, maybe something like a `[go]` section with an `exclude` parameter:

```toml
[go]
# Go binaries to exclude from updates
exclude = [
  "kind",
  "pkgsite"
]
```

The [other option for updating Go binaries, `go-global-update`](https://github.com/Gelio/go-global-update), has different logic for specifying which binaries to update.
It seems that if no arguments are specified, everything in `$GOBIN` is updated.
If some binary names are specified in the CLI arguments, only those specified binaries are updated.

tmchow:
## What does this PR do

Adds a `[go]` config section with a `gup_exclude` option that passes `--exclude` flags to `gup update`. This lets users pin specific Go-installed binaries to non-latest versions without skipping the entire gup step.

Example config:
```toml
[go]
gup_exclude = ["kind", "pkgsite"]
```

This produces: `gup update --exclude kind --exclude pkgsite`

The approach follows the config subsection pattern endorsed by @SteveLauC in [#604](https://github.com/topgrade-rs/topgrade/issues/604#issuecomment-1805451374).

Closes #604

## Standards checklist

- [x] I have tested the code myself, with the relevant tools installed.

### AI involvement

used AI tools (Claude Code) for code suggestions while implementing this. The config pattern, approach, and testing were done by me 👦🏻

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -322,6 +322,10 @@
 # (default: false)
 # interactive = false
 
+[go]
+# Exclude specified binaries from `gup update`
+# gup_exclude = ["kind", "pkgsite"]
+
 
 [npm]
 # Use sudo if the NPM directory isn't owned by the current user
--- a/src/config.rs
+++ b/src/config.rs
@@ -214,6 +214,13 @@ pub struct Brew {
     fetch_head: Option<bool>,
 }
 
+#[derive(Deserialize, Default, Debug, Merge)]
+#[serde(deny_unknown_fields)]
+pub struct Go {
+    #[merge(strategy = crate::utils::merge_strategies::vec_prepend_opt)]
+    gup_exclude: Option<Vec<String>>,
+}
+
 #[derive(Debug, Deserialize, Clone, Copy, Default)]
 #[serde(rename_all = "snake_case")]
 pub enum ArchPackageManager {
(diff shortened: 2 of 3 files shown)


tmchow:
Quick note: the config approach here uses a flat `[go]` section with `gup_exclude`, which follows the existing pattern of other config sections like `[brew]` and `[npm]`. The original issue discussion from 2023 floated a nested `[go.gup]` subsection idea, but that was from a previous maintainer (i think). Happy to restructure if you'd prefer the nested approach or a different field name.

tmchow:
@GideonBear thanks for quick approval and merge!

PR #1913, opened 2026-04-05
feat: show step id in execution separator

### LINKED ISSUE #1467: Show the step id during execution (e.g. in seperator)

Original suggestion by @AThePeanut4

Topgrade should show the step id (used for `--only` and `--disable`) during execution somewhere, e.g. in the seperator and/or the summary.

mvanhorn:
## What does this PR do

Shows the step ID (used for `--only` and `--disable` flags) alongside the display name in the execution separator. The output changes from:

```
── Sparkle ──────────────────
```

to:

```
── Sparkle (sparkle) ────────
```

This makes it easy to find the right step name when you want to skip or isolate a step. The step ID is the snake_case name that `--only` and `--disable` accept.

Implementation adds `strum::Display` to the Step enum and passes the current step ID from the runner to the separator via a static in `terminal.rs`.

Closes #1467

## Standards checklist

### AI involvement

Used Claude for Rust syntax help with strum derives and mutex patterns. The implementation approach (storing step ID in terminal static, reading it in print_separator) was designed by me.

### DIFF
--- a/src/runner.rs
+++ b/src/runner.rs
@@ -9,7 +9,7 @@ use crate::ctrlc;
 use crate::error::{DryRun, MissingSudo, SkipStep};
 use crate::execution_context::ExecutionContext;
 use crate::step::Step;
-use crate::terminal::{ShouldRetry, print_error, print_warning, should_retry};
+use crate::terminal::{ShouldRetry, print_error, print_warning, set_current_step_id, should_retry};
 
 pub enum StepResult {
     Success,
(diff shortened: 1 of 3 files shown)


GideonBear:
Hi, thanks! The current implementation is not acceptable though, a static mutex is unnecessary.

However looking at it now, I think the best way to do this is currently locked behind some v18 stuff, where I kind of want to abolish an explicit `print_separator` call entirely, and handle it within the runner (which already knows which step it is, thus removing the need for communicating that).

I will close this and wait with this until we're implementing that, sorry, that issue probably shouldn't've been labeled as "needs PR" but I only realize that now. If I end up using parts of your PR I will credit you as co-author. Thank you for the PR!

mvanhorn:
Makes sense - handling it in the runner for v18 sounds like the cleaner approach. Thanks for the context and for considering co-author credit. Happy to help if anything's useful when you get to it.

PR #1952, opened 2026-04-13
feat(skills): pass --global

luizkowalski:
edit: defaults to `--global` for skills step, as topgrade shouldn't depend on the current dir

## Standards checklist

### AI involvement
Used Claude to scan the codebase, find the established patterns for configurations and where to edit/add changes

### DIFF
--- a/.github/workflows/lint_pr.yml
+++ b/.github/workflows/lint_pr.yml
@@ -16,7 +16,7 @@ concurrency:
 
 jobs:
   main:
-    name: Validate PR title
+    name: Validate PR title # zizmor: ignore[concurrency-limits] see above
     runs-on: ubuntu-latest
     permissions:
       pull-requests: read # amannn/action-semantic-pull-request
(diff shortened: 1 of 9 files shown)


GideonBear:
Relevant upstream changes: https://github.com/vercel-labs/skills/pull/913

Weird that they just removed a subcommand in a minor version...

I don't think `--project` makes sense at all, topgrade shouldn't depend on its cwd.
Unless you can do something similar to the claude code step and iterate over a list of project-scoped skills, I think we should just add `--global`. Does that make sense?

GideonBear:
(also about the current code, you can use an `enum` to prevent typos in the field)

luizkowalski:
> I  don't think --project makes sense at all, topgrade shouldn't depend on its cwd.

yup, looking back now, I agree. it works (at least work on my machine™) but it doesn't really make any sense given the idea of topgrade (if that makes sense)

> you can use an enum to prevent typos in the field

TIL. I'm forcing myself to learn Rust so I'm not familiar with some things yet 😄 I will update the PR later

luizkowalski:
updated. I don't think the enum idea applies anymore as there's no configuration with the new approach

GideonBear (approved):
Thank you!

PR #1001, opened 2024-12-10
feat: add Ollama support

SteveLauC:
## What does this PR do

Add an `Ollama` step that would pull all the installed LLMs, closes #988.

## Standards checklist

- [x] The PR title is descriptive
- [x] I have read `CONTRIBUTING.md`
- [x] *Optional:* I have tested the code myself
- [x] If this PR introduces new user-facing messages they are translated
 
## For new steps

- [x] *Optional:* Topgrade skips this step where needed
- [x] *Optional:* The `--dry-run` option works with this step
- [ ] *Optional:* The `--yes` option works with this step if it is supported by 
  the underlying command
  This step does not support the `--yes` option

If you developed a feature or a bug fix for someone else and you do not have the
means to test it, please tag this person here.

PR #1968, opened 2026-04-19
feat(step): support Ollama model pulling

### LINKED ISSUE #988: Support Ollama LLM model pulling

I want topgrade to support `ollama`, a tool for managing LLMs locally, see https://github.com/ollama/ollama.
Topgrade should invoke model pulling to update all locally installed models.

Windows, macOS, Linux

if `ollama` is found in path it should be invoked,  this also has to work when `ollama` was installed through package managers like homebrew.

I am currently running through a custom command with `ollama list | awk 'NR>1 && !/reviewer/ {system(\"ollama pull \"$1)}'`
This allows to pull all new model files for currently installed models.
I am not sure if this command works OS independent.

No dry run option available

  option to skip this step?
No user confirmation needed

## More information
To test this a model should be downloaded on the local machine first, e.g. via `ollama run qwen2.5-coder:0.5b`.
When topgrade invokes `ollama` it should show an output similar to
```
pulling manifest
pulling 170370233dd5... 100% ▕████████████████████████████████████████████████████▏ 4.1 GB
pulling 72d6f08a42f6... 100% ▕████████████████████████████████████████████████████▏ 624 MB
pulling 43070e2d4e53... 100% ▕████████████████████████████████████████████████████▏  11 KB
pulling c43332387573... 100% ▕████████████████████████████████████████████████████▏   67 B
pulling ed11eda7790d... 100% ▕████████████████████████████████████████████████████▏   30 B
pulling 7c658f9561e5... 100% ▕████████████████████████████████████████████████████▏  564 B
verifying sha256 digest
writing manifest
success
```

uwuclxdy:
## What does this PR do

Closes https://github.com/topgrade-rs/topgrade/pull/1001
Closes https://github.com/topgrade-rs/topgrade/issues/988 and implements Ollama model pulling as discussed:
- Finds installed models by scanning local manifest directory (`$OLLAMA_MODELS` or `~/.ollama/models`)
- Skips locally created models (they can be detected by looking for a  `from` field in their manifest)
- Pull each remote model by `name:tag`
- If Ollama server is not running, temporarily starts one for the duration of the step and shuts it down after

## Standards checklist

    - with ollama already running: https://asciinema.org/a/pG34eHJpgawmnvtt
    - without ollama running: https://asciinema.org/a/5MNZhplNukItwdNp
    - tested all run types (see recordings)
- [x] ~If this PR introduces new user-facing messages they are translated~

### AI involvement

Claude: partial contribution and review.

## For new steps

  the underlying command

### DIFF
--- a/src/steps/generic.rs
+++ b/src/steps/generic.rs
@@ -14,11 +14,12 @@ use std::{env, path::Path};
 use std::{fs, io::Write};
 use tempfile::tempfile_in;
 use tracing::{debug, error, warn};
+use walkdir::WalkDir;
 
 use crate::HOME_DIR;
 use crate::command::{CommandExt, Utf8Output};
 use crate::execution_context::ExecutionContext;
-use crate::executor::ExecutorOutput;
+use crate::executor::{ExecutorChild, ExecutorOutput};
 use crate::output_changed_message;
 use crate::step::Step;
 use crate::sudo::SudoExecuteOpts;
@@ -2096,4 +2097,135 @@ pub fn run_skills(ctx: &ExecutionContext) -> Result<()> {
     command.args(["skills", "update", "--global"]);
 
     command.status_checked()
+}
+
+fn ollama_serve(ctx: &ExecutionContext, ollama: &Path) -> Result<ExecutorChild> {
+    ctx.execute(ollama)
+        .arg("serve")
+        .stdin(std::process::Stdio::null())
+        .spawn()
+        .context("Failed to start Ollama server")
+}
+
+fn ollama_manifests_path() -> PathBuf {
+    std::env::var_os("OLLAMA_MODELS")
+        .map(PathBuf::from)
+        .unwrap_or_else(|| HOME_DIR.join(".ollama/models"))
+        .join("manifests/registry.ollama.ai")
+}
+
+fn ollama_list_models() -> Result<Vec<String>> {
+    let manifests = ollama_manifests_path();
+    if !manifests.exists() {
+        return Ok(Vec::new());
+    }
+
+    let mut models = Vec::new();
+    for entry in WalkDir::new(&manifests).min_depth(3).max_depth(3) {
+        let entry = entry?;
+        if !entry.file_type().is_file() {
+            continue;
+        }
+
+        let path = entry.path();
+        let tag = path
+            .file_name()
+            .ok_or_else(|| eyre!("invalid manifest path: {}", path.display()))?
+            .to_string_lossy();
+        let name = path
+            .parent()
+            .and_then(|p| p.file_name())
+            .ok_or_else(|| eyre!("invalid manifest path: {}", path.display()))?
+            .to_string_lossy();
+        let namespace = path
+            .parent()
+            .and_then(|p| p.parent())
+            .and_then(|p| p.file_name())
+            .ok_or_else(|| eyre!("invalid manifest path: {}", path.display()))?
+            .to_string_lossy();
+
+        models.push(if namespace == "library" {
+            format!("{name}:{tag}")
+        } else {
+            format!("{namespace}/{name}:{tag}")
+        });
+    }
+
+    Ok(models)
+}
+
+// Locally created model manifests have a `from` field on their model layers.
+fn is_local_ollama_model(model: &str) -> bool {
+    let last_slash = model.rfind('/');
+    let (name, tag) = match model.rfind(':') {
+        Some(index) if last_slash.is_none_or(|slash| index > slash) => (&model[..index], &model[index + 1..]),
+        _ => (model, "latest"),
+    };
+
+    let name_path: PathBuf = if name.contains('/') {
+        name.split('/').collect()
+    } else {
+        PathBuf::from("library").join(name)
+    };
+
+    let manifest_path = ollama_manifests_path().join(name_path).join(tag);
+
+    fs::read_to_string(manifest_path)
+        .ok()
+        .and_then(|content| serde_json::from_str::<serde_json::Value>(&content).ok())
+        .is_some_and(|manifest| {
+            manifest["layers"].as_array().is_some_and(|layers| {
+                layers.iter().any(|layer| {
+                    layer["mediaType"].as_str() == Some("application/vnd.ollama.image.model")
+                        && layer.get("from").is_some()
+                })
+            })
+        })
+}
+
+pub fn run_ollama_pull(ctx: &ExecutionContext) -> Result<()> {
+    let ollama = require("ollama")?;
+
+    let models = ollama_list_models()?;
+    let remote_models: Vec<_> = models.iter().filter(|m| !is_local_ollama_model(m)).collect();
+    if models.is_empty() || remote_models.is_empty() {
+        return Ok(());
+    }
+
+    print_separator("Ollama");
+
+    // Start ollama if needed
+    let mut server: Option<ExecutorChild> = None;
+    if let ExecutorOutput::Wet(out) = ctx.execute(&ollama).always().args(["list"]).output()?
+        && String::from_utf8_lossy(&out.stderr).contains("could not connect")
+        && !ctx.run_type().dry()
+    {
+        debug!("Ollama server not running, starting temporary server");
+        server = Some(ollama_serve(ctx, &ollama)?);
+        // Poll for server readiness (2-second budget: 10 × 200ms).
+        let _ = (0..10).any(|_| {
+            std::thread::sleep(std::time::Duration::from_millis(200));
+            ctx.execute(&ollama).args(["list"]).status_checked().is_ok()
+        });
+    }
+
+    let mut failed = 0;
+    for model in &remote_models {
+        if ctx.execute(&ollama).args(["pull", model]).status_checked().is_err() {
+            failed += 1;
+            warn!("Failed to pull Ollama model '{model}'");
+            print_warning(format!("Failed to pull Ollama model '{model}'."));
+        }
+    }
+
+    if let Some(ExecutorChild::Wet(mut child)) = server {
+        let _ = child.kill();
+        let _ = child.wait();
+    }
+
+    if failed == remote_models.len() {
+        bail!("All Ollama model pulls failed");
+    }
+
+    Ok(())
 }
--- a/src/steps/generic.rs
+++ b/src/steps/generic.rs
@@ -2121,7 +2121,7 @@ struct OllamaModel {
 
 fn ollama_list_models() -> Result<impl Iterator<Item = OllamaModel>> {
     let manifests = ollama_manifests_path();
-    
+
     let iter = WalkDir::new(&manifests)
         .min_depth(3)
         .max_depth(3)
--- a/src/steps/generic.rs
+++ b/src/steps/generic.rs
@@ -2114,7 +2114,12 @@ fn ollama_manifests_path() -> PathBuf {
         .join("manifests/registry.ollama.ai")
 }
 
-fn ollama_list_models() -> Result<Vec<String>> {
+struct OllamaModel {
+    name: String,
+    manifest_path: PathBuf,
+}
+
+fn ollama_list_models() -> Result<Vec<OllamaModel>> {
     let manifests = ollama_manifests_path();
     let mut models = Vec::new();
     if !manifests.exists() {
@@ -2144,10 +2149,13 @@ fn ollama_list_models() -> Result<Vec<String>> {
             .ok_or_else(|| eyre!("invalid manifest path: {}", path.display()))?
             .to_string_lossy();
 
-        models.push(if namespace == "library" {
-            format!("{name}:{tag}")
-        } else {
-            format!("{namespace}/{name}:{tag}")
+        models.push(OllamaModel {
+            name: if namespace == "library" {
+                format!("{name}:{tag}")
+            } else {
+                format!("{namespace}/{name}:{tag}")
+            },
+            manifest_path: path.to_path_buf(),
         });
     }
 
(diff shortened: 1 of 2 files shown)


src/steps/generic.rs
```diff
@@ -2096,3 +2097,134 @@ pub fn run_skills(ctx: &ExecutionContext) -> Result<()> {
...
+}
+
+fn ollama_list_models() -> Result<Vec<String>> {
+    let manifests = ollama_manifests_path();
+    if !manifests.exists() {
+        return Ok(Vec::new());
+    }
+
+    let mut models = Vec::new();
+    for entry in WalkDir::new(&manifests).min_depth(3).max_depth(3) {
```
GideonBear: Nit, maybe not actually better so try first; but instead of walking like this, wouldn't a nested loop be somewhat cleaner, since you're using all the parents?

src/steps/generic.rs
```diff
@@ -2096,3 +2097,134 @@ pub fn run_skills(ctx: &ExecutionContext) -> Result<()> {
...
+            format!("{namespace}/{name}:{tag}")
+        });
+    }
+
+    Ok(models)
+}
+
+// Locally created model manifests have a `from` field on their model layers.
+fn is_local_ollama_model(model: &str) -> bool {
+    let last_slash = model.rfind('/');
```
GideonBear: Comment with example(s) maybe? I have no context for this

src/steps/generic.rs
```diff
@@ -2096,3 +2097,134 @@ pub fn run_skills(ctx: &ExecutionContext) -> Result<()> {
...
+        name.split('/').collect()
+    } else {
+        PathBuf::from("library").join(name)
+    };
+
+    let manifest_path = ollama_manifests_path().join(name_path).join(tag);
+
+    fs::read_to_string(manifest_path)
+        .ok()
+        .and_then(|content| serde_json::from_str::<serde_json::Value>(&content).ok())
```
GideonBear: Possible to deser a struct instead?

src/steps/generic.rs
```diff
@@ -2096,3 +2097,134 @@ pub fn run_skills(ctx: &ExecutionContext) -> Result<()> {
...
+        .is_some_and(|manifest| {
```
GideonBear: I don't fully understand these chains, but can we `fs::read_to_string(manifest_path)?` for example? It's fine to wrap the result in `Ok()` if necessary.

src/steps/generic.rs
```diff
@@ -2096,3 +2097,134 @@ pub fn run_skills(ctx: &ExecutionContext) -> Result<()> {
...
+            print_warning(format!("Failed to pull Ollama model '{model}'."));
+        }
+    }
+
+    if let Some(ExecutorChild::Wet(mut child)) = server {
+        let _ = child.kill();
+        let _ = child.wait();
+    }
+
+    if failed == remote_models.len() {
```
GideonBear: If we're not going the ignore failures route anymore, why are even a few failures ignored?

src/steps/generic.rs
```diff
@@ -2096,3 +2097,134 @@ pub fn run_skills(ctx: &ExecutionContext) -> Result<()> {
...
+    let mut failed = 0;
+    for model in &remote_models {
+        if ctx.execute(&ollama).args(["pull", model]).status_checked().is_err() {
+            failed += 1;
+            warn!("Failed to pull Ollama model '{model}'");
+            print_warning(format!("Failed to pull Ollama model '{model}'."));
+        }
+    }
+
+    if let Some(ExecutorChild::Wet(mut child)) = server {
```
GideonBear: Maybe add comments to be a little more explicit that we need to not have any returning code paths in between these two blocks

src/steps/generic.rs
```diff
@@ -2096,3 +2097,134 @@ pub fn run_skills(ctx: &ExecutionContext) -> Result<()> {
...
+        && String::from_utf8_lossy(&out.stderr).contains("could not connect")
+        && !ctx.run_type().dry()
+    {
+        debug!("Ollama server not running, starting temporary server");
+        server = Some(ollama_serve(ctx, &ollama)?);
+        // Poll for server readiness (2-second budget: 10 × 200ms).
+        let _ = (0..10).any(|_| {
+            std::thread::sleep(std::time::Duration::from_millis(200));
+            ctx.execute(&ollama).args(["list"]).status_checked().is_ok()
+        });
```
```suggestion
        for _ in 0..10 {
            std::thread::sleep(std::time::Duration::from_millis(200));
            if ctx.execute(&ollama).args(["list"]).status_checked().is_ok() { break; }
        }
```
Style; `any(|_|` seems unnecessary and confusing.

GideonBear (changes requested):

uwuclxdy:
thanks for review! i fixed those issues now

src/steps/generic.rs
```diff
@@ -2097,3 +2098,148 @@ pub fn run_skills(ctx: &ExecutionContext) -> Result<()> {
...
+        }
+    }
+
+    // Accumulate failures rather than propagating — ensures server shutdown below always runs.
+    for model in &remote_models {
+        if ctx.execute(&ollama).args(["pull", model]).status_checked().is_err() {
+            warn!("Failed to pull Ollama model '{model}'");
+            print_warning(format!("Failed to pull Ollama model '{model}'."));
+        }
+    }
```
GideonBear: Would still like to fail the step on any error IMO. Any reason not to; *should* there be any case where we can't handle it?

src/steps/generic.rs
```diff
@@ -2097,3 +2098,148 @@ pub fn run_skills(ctx: &ExecutionContext) -> Result<()> {
...
+            .and_then(|p| p.parent())
+            .and_then(|p| p.file_name())
+            .ok_or_else(|| eyre!("invalid manifest path: {}", path.display()))?
+            .to_string_lossy();
+
+        models.push(if namespace == "library" {
+            format!("{name}:{tag}")
+        } else {
+            format!("{namespace}/{name}:{tag}")
+        });
```
GideonBear: Aren't we parsing this in `is_local_ollama_model`? Can we just store it in a struct or something instead? Then add add a method for making it into a string to pass into `ollama pull`.

src/steps/generic.rs
```diff
@@ -2097,3 +2098,148 @@ pub fn run_skills(ctx: &ExecutionContext) -> Result<()> {
...
+        .any(|l| l.media_type == "application/vnd.ollama.image.model" && l.from.is_some())
+}
+
+pub fn run_ollama_pull(ctx: &ExecutionContext) -> Result<()> {
+    let ollama = require("ollama")?;
+
+    let models = ollama_list_models()?;
+    let remote_models: Vec<_> = models.iter().filter(|m| !is_local_ollama_model(m)).collect();
+    if remote_models.is_empty() {
+        return Ok(());
```
GideonBear: If exiting before separator, `return Err(SkipStep)`

GideonBear (changes requested):

src/steps/generic.rs
```diff
@@ -2097,3 +2098,139 @@ pub fn run_skills(ctx: &ExecutionContext) -> Result<()> {
...
+}
+
+fn ollama_list_models() -> Result<Vec<OllamaModel>> {
+    let manifests = ollama_manifests_path();
+    let mut models = Vec::new();
+    if !manifests.exists() {
+        return Ok(models);
+    }
+
+    for entry in WalkDir::new(&manifests).min_depth(3).max_depth(3) {
```
GideonBear: Nit: you can probably ` -> Result<impl Iterator<Item = OllamaModel>>` and `.filter_map`

src/steps/generic.rs
```diff
@@ -2097,3 +2098,139 @@ pub fn run_skills(ctx: &ExecutionContext) -> Result<()> {
...
+    manifest
+        .layers
+        .iter()
+        .any(|l| l.media_type == "application/vnd.ollama.image.model" && l.from.is_some())
+}
+
+pub fn run_ollama_pull(ctx: &ExecutionContext) -> Result<()> {
+    let ollama = require("ollama")?;
+
+    let remote_models: Vec<OllamaModel> = ollama_list_models()?
```
GideonBear: Here as well, I don't think there's any need to collect

src/steps/generic.rs
```diff
@@ -2097,3 +2098,139 @@ pub fn run_skills(ctx: &ExecutionContext) -> Result<()> {
...
+
+    let mut pull_result = Ok(());
+    for model in &remote_models {
+        if let Err(e) = ctx.execute(&ollama).args(["pull", &model.name]).status_checked() {
+            pull_result = Err(e);
+            break;
+        }
+    }
+
+    // Do not add early returns between here and the server shutdown
```
GideonBear: Wrong place?

GideonBear (changes requested):

src/steps/generic.rs
uwuclxdy: kinda, mb on the wording

GideonBear (approved):
Awesome, thank you! ❤️

GideonBear (approved):

PonyLucky:
I'm wondering if I need to enable it in the config file. Because I have the right version yet if doesn't show the `ollama` section. Maybe it's because there is nothing to update... Can someone tell me if that's normal ?

image


Config:
image

uwuclxdy:
Hmm weird.. Could you run `topgrade --only ollama -v` and show the output after "Sudo" step @PonyLucky ?

Also, does `cat $OLLAMA_MODELS` print anything? And where are your model manifests saved (could you share `tree` command of the folder, for me it's `~/.ollama/models/manifests/`). Thanks

PonyLucky:
@uwuclxdy  Yes, here it is:
image

I think the `ollama` section is not showing even if installed if there is nothing to do. Or maybe because the models I have don't have any 'version' (first and only release, see <https://ollama.com/library/gemma4/tags>).

uwuclxdy:
Thanks @PonyLucky , this looks like topgrade not finding any updatable models with current logic in your folder. Could you please tell me the path to your manifests folder and a manifest file of any model that you've pulled?

For example `/home/uwuclxdy/.ollama/models/manifests/registry.ollama.ai/library/nomic-embed-text/latest`:
```json
{"schemaVersion":2,"mediaType":"application/vnd.docker.distribution.manifest.v2+json","config":{"mediaType":"application/vnd.docker.container.image.v1+json","digest":"sha256:31df23ea7daa448f9ccdbbcecce6c14689c8552222b80defd3830707c0139d4f","size":420},"layers":[{"mediaType":"application/vnd.ollama.image.model","digest":"sha256:970aa74c0a90ef7482477cf803618e776e173c007bf957f635f1015bfcfef0e6","size":274290656},{"mediaType":"application/vnd.ollama.image.license","digest":"sha256:c71d239df91726fc519c6eb72d318ec65820627232b2f796219e87dcf35d0ab4","size":11357},{"mediaType":"application/vnd.ollama.image.params","digest":"sha256:ce4a164fc04605703b485251fe9f1a181688ba0eb6badb80cc6335c0de17ca0d","size":17}]}
```

This one is considered updatable, identified by absence of "from" field.

PonyLucky:
@uwuclxdy I took a bit of time to find the path `/usr/share/ollama/.ollama/models/manifests/registry.ollama.ai/library`.

`gemma4/26b`:
```json
{"schemaVersion":2,"mediaType":"application/vnd.docker.distribution.manifest.v2+json","config":{"mediaType":"application/vnd.docker.container.image.v1+json","digest":"sha256:90294508afbbc106b71a808a80e13bdf88c20ecbefcd6cf3380eafa484cef3e4","size":474},"layers":[{"mediaType":"application/vnd.ollama.image.model","digest":"sha256:7121486771cbfe218851513210c40b35dbdee93ab1ef43fe36283c883980f0df","size":17987569344},{"mediaType":"application/vnd.ollama.image.license","digest":"sha256:7339fa418c9ad3e8e12e74ad0fd26a9cc4be8703f9c110728a992b193be85cb2","size":11355},{"mediaType":"application/vnd.ollama.image.params","digest":"sha256:56380ca2ab89f1f68c283f4d50863c0bcab52ae3f1b9a88e4ab5617b176f71a3","size":42}]}
```

`gpt-oss/20b`:
```json
{"schemaVersion":2,"mediaType":"application/vnd.docker.distribution.manifest.v2+json","config":{"mediaType":"application/vnd.docker.container.image.v1+json","digest":"sha256:776beb3adb235076157cfea408b8ea2a2d25eae99d7f5da997f607f6b69fa0fa","size":489},"layers":[{"mediaType":"application/vnd.ollama.image.model","digest":"sha256:e7b273f9636059a689e3ddcab3716e4f65abe0143ac978e46673ad0e52d09efb","size":13793422144},{"mediaType":"application/vnd.ollama.image.template","digest":"sha256:fa6710a93d78da62641e192361344be7a8c0a1c3737f139cf89f20ce1626b99c","size":7240},{"mediaType":"application/vnd.ollama.image.license","digest":"sha256:f60356777647e927149cbd4c0ec1314a90caba9400ad205ddc4ce47ed001c2d6","size":11353},{"mediaType":"application/vnd.ollama.image.params","digest":"sha256:d8ba2f9a17b3bbdeb5690efaa409b3fcb0b56296a777c7a69c78aa33bbddf182","size":18}]}
```

uwuclxdy:
The models should be recognized as updatable **if** topgrade would see this directory (which I'm pretty confident it doesn't since it's a non-standard path). However, topgrade also checks `OLLAMA_MODELS` env variable:

```
export OLLAMA_MODELS="/usr/share/ollama/.ollama/models/"
topgrade --only ollama -v
```

ping @PonyLucky

PonyLucky:
@uwuclxdy On the non standard path, I beg to differ.

For instance, if I do on duckduckgo "ollama models path" I find this link in second place:
<https://medium.com/@rosgluk/move-ollama-models-to-different-location-755eaec1df96>

In here:
> By default the model files are stored:
>
>    Windows: `C:\Users\%username%\.ollama\models`
>    Linux: `/usr/share/ollama/.ollama/models`
>    macOS: `~/.ollama/models`

I just installed ollama from the command on the website and it put itself there. Maybe an old/new default.

I think it can be safe to add this folder in fallback to the actual one (`$HOME/.ollama/models/` if not found.

What do you think ?

uwuclxdy:
@PonyLucky looks like you're right. I don't remember finding this at time of implementing it. I'll open a PR tomorrow (say if you want to do it instead) to check these paths before failing, thanks!

BTW, did you try setting the env variable and did it work with it? Could be a temporary fix until next release / merge.

PonyLucky:
@uwuclxdy I can try creating a PR. It shouldn't be hard to implement.

Thank you for the opportunity.

GideonBear:
Looks like the difference is in a user- or systemwide installation of ollama. Adding another `.unwrap_or_else` with the system-wide path in `ollama_manifests_path` should be enough. PR appreciated from any of you!

uwuclxdy:
@PonyLucky quick question, do you run ollama with sudo?

PonyLucky:
No, not with sudo.

I do it with the current logged in user ($USER).

uwuclxdy:
> No, not with sudo.

Awesome, this should work then!

PR #1978, opened 2026-04-21
fix(mise): handle all self-update errors

salim-b:
## What does this PR do

mise errors with a different error message when the `self_update` feature was disabled at build time (as is the case for e.g. the [conda-forge package](https://anaconda.org/conda-forge/mise)) than when it has only been soft-disabled (e.g. by setting an env var). In both cases, the error message [ends with the words `cannot update`](https://github.com/search?q=repo%3Ajdx%2Fmise%20%2Fcannot%20update%2F&type=code), so let's just check for that.
## Standards checklist

### AI involvement

None.


GideonBear (approved):
Thanks!

PR #1985, opened 2026-04-25
feat: passwordless reboot and shutdown using `system_shutdown`

### LINKED ISSUE #861: Add support for systemd's systemctl reboot for passwordless reboots
Topgrade should allow for reboot without requiring a sudo password by using command "systemctl reboot" if the system is running systemd.

## More information
If the linux system uses systemd you do not need to reboot using sudo, which requires typing in a password. Instead use the command "systemctl reboot" which will just reboot the system without further user action

https://www.freedesktop.org/software/systemd/man/latest/sd_booted.html
"Internally, this function checks whether the directory /run/systemd/system/ exists. A simple check like this can also be implemented trivially in shell or any other language."

Seems like you can check for this folder in rust and then execute the alternate command "systemctl reboot" instead in the below function

`pub fn reboot() -> Result<()> {
    print!("Rebooting...");
    Command::new("sudo").arg("reboot").status_checked()
}`

I do not know how to test this from a side branch, as I am using bazzite, an immutable desktop. I'm willing to give it a try though if there are instructions somewhere.

### LINKED ISSUE #1810: feat request: shutdown option
## Checklist

I'd like there to be a Shutdown option after topgrade completes (there are currently: quit, shell, reboot). Don't know what key binding would be ideal for it tho. Maybe `P` or `O` as in `poweroff`? idk

## More information

Possibly related with #861

uwuclxdy:
## What does this PR do

Closes #861 
Closes #1810 

Replaces custom reboot implementation with `system_shutdown` crate (for Linux, macOS and Windows) and adds a new poweroff option.

- (P)oweroff option in the keep_at_end menu.
- Passwordless reboot on Linux/macOS.
- Windows: replaced `shutdown.exe /R /T 0` with `system_shutdown::reboot()` - the crate says Windows reboots are stable.
- BSD and others: no change; existing sudo fallback for reboot and shutdown.

Testers are welcome to try and give feedback!

## Standards checklist

	- not yet

### AI involvement

Claude only for review.

### DIFF
--- a/src/steps/os/unix.rs
+++ b/src/steps/os/unix.rs
@@ -1097,7 +1097,7 @@ pub fn shutdown(ctx: &ExecutionContext) -> Result<()> {
         return Ok(());
     }
     match ctx.sudo() {
-        Some(sudo) => sudo.execute(ctx, "shutdown")?.args(["-h", "now"]).status_checked(),
-        None => ctx.execute("shutdown").args(["-h", "now"]).status_checked(),
+        Some(sudo) => sudo.execute(ctx, "shutdown")?.args(["now"]).status_checked(),
+        None => ctx.execute("shutdown").args(["now"]).status_checked(),
     }
 }
--- a/src/steps/os/unix.rs
+++ b/src/steps/os/unix.rs
@@ -1081,8 +1081,23 @@ pub fn run_atuin(ctx: &ExecutionContext) -> Result<()> {
 }
 
 pub fn reboot(ctx: &ExecutionContext) -> Result<()> {
+    #[cfg(any(target_os = "linux", target_os = "macos"))]
+    if let Ok(()) = system_shutdown::reboot() {
+        return Ok(());
+    }
     match ctx.sudo() {
         Some(sudo) => sudo.execute(ctx, "reboot")?.status_checked(),
         None => ctx.execute("reboot").status_checked(),
+    }
+}
+
+pub fn shutdown(ctx: &ExecutionContext) -> Result<()> {
+    #[cfg(any(target_os = "linux", target_os = "macos"))]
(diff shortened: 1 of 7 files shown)


src/steps/os/unix.rs
```diff
@@ -1081,8 +1081,23 @@ pub fn run_atuin(ctx: &ExecutionContext) -> Result<()> {
...
 }
+
+pub fn shutdown(ctx: &ExecutionContext) -> Result<()> {
+    #[cfg(any(target_os = "linux", target_os = "macos"))]
+    if let Ok(()) = system_shutdown::shutdown() {
+        return Ok(());
+    }
+    match ctx.sudo() {
+        Some(sudo) => sudo.execute(ctx, "shutdown")?.args(["-h", "now"]).status_checked(),
+        None => ctx.execute("shutdown").args(["-h", "now"]).status_checked(),
```
GideonBear: `-h` can probably be removed:
>        -P, --poweroff
>           Power the machine off (the default).
>       -h
>           The same as --poweroff, but does not override the action to take if it is "halt". E.g.  shutdown --reboot -h means "poweroff", but shutdown --halt -h means "halt".

GideonBear (changes requested):

GideonBear:
When you're testing poweroff and restart, can you temporarily block the `Err` codepath to make sure the crate is actually doing its job?

GideonBear (approved):

uwuclxdy:
Tested with:

```rs
pub fn shutdown(ctx: &ExecutionContext) -> Result<()> {
    #[cfg(any(target_os = "linux", target_os = "macos"))]
    if let Ok(()) = system_shutdown::shutdown() {
        return Ok(());
    }
	return Ok(());
}
```

and it worked, thanks for merge :D

PR #1991, opened 2026-04-28
ci: auto-fix cargo fmt and clippy on PRs

uwuclxdy:
## What does this PR do

Adds an auto-fix job to `ci.yml` that runs on every pull request. It runs `cargo fmt` and `cargo clippy --fix`, commits any changes back to the PR branch, which triggers a fresh CI run. Also adds rustfmt and clippy components to rust-toolchain.toml so they're available without an explicit rustup step (which is also removed in this PR).

There is one annoying thing about this tho - it needs a PAT (`secrets.TOKEN`) because GitHub blocks pushes made with `GITHUB_TOKEN` from re-triggering workflows (to prevent infinite loops). However, a PAT push is treated as a regular user push and does re-trigger CI. `secrets.TOKEN` should be a fine-grained PAT scoped to `contents: write` for this repo only.

I initially thought this could be done with a pre-commit hook that every contributor runs on their machine, but there were too many problems with that (requires setup, can be bypassed...). Doing it in CI guarantees the auto-fixes happen on every PR.

## Standards checklist

### AI involvement

Claude (Code) for the majority of this, verified it myself before opening the PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,6 +24,33 @@ defaults:
     shell: bash
 
 jobs:
+  auto-fix:
+    name: Auto-fix (cargo fmt & clippy --fix)
+    runs-on: ubuntu-latest
+    if: github.event_name == 'pull_request'
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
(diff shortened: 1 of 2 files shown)


GideonBear:
Because of security concerns, I don't want a PAT with full write access in GHA.

I'm working on a way to do this via pre-commit.ci, but if that doesn't end up working the way I want it to I will drop this completely, it's not very important.

Thanks for trying though!

PR #1993, opened 2026-04-29
fix(powershell): handle unavailable Update-Module

### LINKED ISSUE #1443: PowerShell module update step fails on Windows 11 25H2
## Erroneous Behavior
When running the PowerShell step in Topgrade on Windows 11 25H2, the module update fails in two stages:

1. **Initial failure** with PowerShellGet 1.0.0.1:
   ```
   Updating modules...
   Cannot convert null to type "System.DateTime".
   At C:\Program Files\WindowsPowerShell\Modules\PowerShellGet\1.0.0.1\PSModule.psm1:7016 char:17
   +                 $InstalledDate = $null
   +                 ~~~~~~~~~~~~~~~~~~~~~~
   + CategoryInfo          : MetadataError: (:) [], ArgumentTransformationMetadataException
   + FullyQualifiedErrorId : RuntimeException
   ```

2. After manually updating PowerShellGet, Topgrade then fails to update PackageManagement:
   ```
   WARNING: The version '1.4.8.1' of module 'PackageManagement' is currently in use.
   Retry the operation after closing the applications.
   ```

## Expected Behavior
Topgrade should update PowerShellGet and PackageManagement cleanly without requiring manual intervention or manual unloading of modules.

## Steps to reproduce
1. On Windows 11 25H2, run:
   ```powershell
   topgrade -y -c -k --only powershell
   ```
2. Observe the PowerShell module update step.

## Possible Cause (Optional)
- PowerShellGet 1.0.0.1 contains a bug where `$null` cannot be assigned to a `[DateTime]`.
- PackageManagement is already loaded in the Topgrade PowerShell session, preventing it from being updated in place.

- [x] Yes  
Running `Install-Module -Name PowerShellGet` directly in the same session also fails until manually elevated and updated.

- [ ] Yes  
- [x] No  

```toml
# no special configuration relevant
```

## Additional Details
- **Operating System/Version**: Windows 11 25H2  
- **Installation**: `winget install topgrade`  
- **Topgrade version**: `topgrade 16.1.2`  

<details>
<pre>
(paste verbose output here if desired)
</pre>
</details>

niStee:
## What does this PR do

Closes #1443.

This changes the PowerShell module update step to avoid passing `-Force` to `Update-Module` when `--yes` is used. Instead, it passes `-Confirm:$false` and conditionally adds `AcceptLicense` when the installed `Update-Module` supports it.

It also checks whether `PowerShellGet` can provide `Update-Module` before running the update. If `Update-Module` is unavailable or cannot be loaded, Topgrade now prints a warning and marks the step as skipped instead of failing with a command-not-found error or reporting `OK` after doing no update.

This helps Windows PowerShell 5.1 setups where `PowerShellGet`/`PackageManagement` are broken, blocked by cloud-backed module folders, or otherwise cannot be autoloaded.

## Standards checklist

### AI involvement

I used AI assistance while investigating the PowerShell behavior, drafting the code change, and running local checks. I reviewed the changes myself before submitting.

### DIFF
--- a/src/steps/generic.rs
+++ b/src/steps/generic.rs
@@ -1210,8 +1210,7 @@ pub fn run_powershell(ctx: &ExecutionContext) -> Result<()> {
     // For Windows PowerShell, use sudo (defaults to AllUsers scope).
     let use_sudo = !powershell.is_pwsh();
 
-    let check_use_sudo = use_sudo && !ctx.run_type().dry();
-    if !powershell.has_command(ctx, "Update-Module", check_use_sudo)? {
+    if !powershell.has_command(ctx, "Update-Module", use_sudo)? {
         let message = t!(
             "PowerShellGet Update-Module is unavailable or could not be loaded. Skipping PowerShell module updates."
         )
--- a/src/steps/generic.rs
+++ b/src/steps/generic.rs
@@ -1210,7 +1210,8 @@ pub fn run_powershell(ctx: &ExecutionContext) -> Result<()> {
     // For Windows PowerShell, use sudo (defaults to AllUsers scope).
     let use_sudo = !powershell.is_pwsh();
 
-    if !ctx.run_type().dry() && !powershell_update_module_available(ctx, use_sudo)? {
+    let check_use_sudo = use_sudo && !ctx.run_type().dry();
+    if !powershell.has_command(ctx, "Update-Module", check_use_sudo)? {
         let message = t!(
             "PowerShellGet Update-Module is unavailable or could not be loaded. Skipping PowerShell module updates."
         )
@@ -1226,34 +1227,23 @@ pub fn run_powershell(ctx: &ExecutionContext) -> Result<()> {
     powershell.build_command(ctx, &cmd, use_sudo)?.status_checked()
 }
 
-fn powershell_update_module_available(ctx: &ExecutionContext, use_sudo: bool) -> Result<bool> {
-    let output = ctx
-        .require_powershell()?
-        .build_command(
-            ctx,
-            "if (Get-Command Update-Module -ErrorAction SilentlyContinue) { Write-Output 'true' }",
-            use_sudo,
-        )?
-        .output_checked_utf8()?;
-
-    Ok(output.stdout.trim().eq_ignore_ascii_case("true"))
-}
-
 fn powershell_update_modules_command(verbose: bool, assume_yes: bool) -> String {
-    let mut cmd = "$topgradeUpdateModuleParams = @{};".to_string();
-    cmd.push_str(" $topgradeUpdateModuleCommand = Get-Command Update-Module -ErrorAction Stop;");
+    let mut cmd = "$params = @{};".to_string();
+    cmd.push_str(" $updateModule = Get-Command Update-Module -ErrorAction Stop;");
 
     if verbose {
-        cmd.push_str(" $topgradeUpdateModuleParams['Verbose'] = $true;");
+        cmd.push_str(" $params['Verbose'] = $true;");
     }
     if assume_yes {
         // Avoid -Force here: PowerShellGet uses it to reinstall already-current modules,
         // which can lock PackageManagement in the running Windows PowerShell session.
-        cmd.push_str(" $topgradeUpdateModuleParams['Confirm'] = $false;");
-        cmd.push_str(" if ($topgradeUpdateModuleCommand.Parameters.ContainsKey('AcceptLicense')) { $topgradeUpdateModuleParams['AcceptLicense'] = $true; }");
+        cmd.push_str(" $params['Confirm'] = $false;");
+        cmd.push_str(
+            " if ($updateModule.Parameters.ContainsKey('AcceptLicense')) { $params['AcceptLicense'] = $true; }",
+        );
     }
 
-    cmd.push_str(" & $topgradeUpdateModuleCommand @topgradeUpdateModuleParams");
+    cmd.push_str(" & $updateModule @params");
     cmd
 }
 
@@ -1265,18 +1255,21 @@ mod powershell_tests {
     fn assume_yes_suppresses_confirmation_without_force() {
         let cmd = powershell_update_modules_command(false, true);
 
-        assert!(cmd.contains("$topgradeUpdateModuleParams['Confirm'] = $false"));
-        assert!(cmd.contains("AcceptLicense"));
+        assert!(cmd.contains("$params['Confirm'] = $false"));
+        assert!(cmd.contains("$updateModule.Parameters.ContainsKey('AcceptLicense')"));
+        assert!(cmd.contains("$params['AcceptLicense'] = $true"));
         assert!(cmd.contains("Get-Command Update-Module -ErrorAction Stop"));
+        assert!(cmd.contains("& $updateModule @params"));
         assert!(!cmd.contains("-Force"));
         assert!(!cmd.contains("exit 0"));
+        assert!(!cmd.contains("topgradeUpdateModule"));
     }
 
     #[test]
     fn verbose_sets_verbose_parameter() {
         let cmd = powershell_update_modules_command(true, false);
 
-        assert!(cmd.contains("$topgradeUpdateModuleParams['Verbose'] = $true"));
+        assert!(cmd.contains("$params['Verbose'] = $true"));
         assert!(!cmd.contains("AcceptLicense"));
     }
 }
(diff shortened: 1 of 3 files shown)


niStee:
Tested with:

- `cargo fmt --check`
- `cargo test`
- `cargo clippy --all-targets -- -D warnings`
- Manual PowerShell checks with PowerShell 7, Windows PowerShell 5.1, gsudo, and native Windows sudo.

GideonBear (approved):
Looks fine to me, thanks!

GideonBear:
@atoerien any chance you can review this?

src/steps/generic.rs
```diff
@@ -1206,25 +1206,78 @@ pub fn run_powershell(ctx: &ExecutionContext) -> Result<()> {
...
+        .require_powershell()?
+        .build_command(
+            ctx,
+            "if (Get-Command Update-Module -ErrorAction SilentlyContinue) { Write-Output 'true' }",
+            use_sudo,
+        )?
+        .output_checked_utf8()?;
+
+    Ok(output.stdout.trim().eq_ignore_ascii_case("true"))
+}
```
atoerien: Could we maybe make this generic, and move it to `powershell.rs`, to go along with `Powershell::has_module`? The check could then just be something like `powershell.has_command("UpdateModule")`.

src/steps/generic.rs
```diff
@@ -1206,25 +1206,78 @@ pub fn run_powershell(ctx: &ExecutionContext) -> Result<()> {
...
+    // For PowerShell Core, run without sudo (defaults to CurrentUser scope).
+    // For Windows PowerShell, use sudo (defaults to AllUsers scope).
+    let use_sudo = !powershell.is_pwsh();
 
-    if ctx.config().verbose() {
-        cmd.push_str(" -Verbose");
-    }
-    if ctx.config().yes(Step::Powershell) {
-        cmd.push_str(" -Force");
+    if !ctx.run_type().dry() && !powershell_update_module_available(ctx, use_sudo)? {
```
atoerien: Can we omit the `!ctx.run_type().dry()` here? As I see it, a dry-run should only skip the commands that do the actual updating, i.e. those that have side-effects. Having the warning also show up in a dry run would be useful as well.

We probably shouldn't be using sudo in a dry run though. But do we need to use it? I think the command check should still work, but I haven't tested it.

src/steps/generic.rs
```diff
@@ -1206,25 +1206,78 @@ pub fn run_powershell(ctx: &ExecutionContext) -> Result<()> {
...
+    if assume_yes {
+        // Avoid -Force here: PowerShellGet uses it to reinstall already-current modules,
+        // which can lock PackageManagement in the running Windows PowerShell session.
+        cmd.push_str(" $topgradeUpdateModuleParams['Confirm'] = $false;");
+        cmd.push_str(" if ($topgradeUpdateModuleCommand.Parameters.ContainsKey('AcceptLicense')) { $topgradeUpdateModuleParams['AcceptLicense'] = $true; }");
+    }
+
+    cmd.push_str(" & $topgradeUpdateModuleCommand @topgradeUpdateModuleParams");
+    cmd
+}
```
atoerien, line 1247: I don't think we need to prefix all the variables with `topgradeUpdateModule` here, just adds unnecessary verbosity

atoerien (commented):
Looks good, tested working on macOS, but I haven't been able to test on Windows yet.

Just have a couple relatively minor comments.

niStee:
Latest push addresses the PowerShell review feedback:
- moved the Update-Module command probe into Powershell::has_command
- dry-run now performs the read-only availability check without sudo
- shortened the generated PowerShell variables
- added/updated focused tests for the command builder and PowerShell probe string

Validated on Linux:
- cargo fmt --check
- cargo check
- cargo test powershell
- cargo clippy --all-targets -- -D warnings
- git diff --check

~~Windows runtime testing is still pending. **Please hold merging until I test PowerShell 7 / pwsh and Windows PowerShell 5.1 locally.**~~

src/steps/generic.rs
```diff
@@ -1206,25 +1206,71 @@ pub fn run_powershell(ctx: &ExecutionContext) -> Result<()> {
 
     print_separator(t!("Powershell Modules Update"));
 
-    let mut cmd = "Update-Module".to_string();
+    // For PowerShell Core, run without sudo (defaults to CurrentUser scope).
+    // For Windows PowerShell, use sudo (defaults to AllUsers scope).
+    let use_sudo = !powershell.is_pwsh();
+
+    let check_use_sudo = use_sudo && !ctx.run_type().dry();
```
atoerien: I meant rather just don't ever use sudo for the check - it's not great to run one thing normally but then run something different when in a dry run.

On Windows, which is the only place you can have PowerShell 5.1 and thus the once place we'd use sudo, I'm pretty sure that running powershell under sudo won't change its `PSModulePath` and so won't change what commands are available.

But I could be wrong, I haven't tested it. If it does change what commands are available, then we should just run the check using sudo under a dry run as well. Running different things for dry and wet runs would definitely be a bigger problem than just using sudo in a dry run.

src/steps/powershell.rs
```diff
@@ -157,3 +169,40 @@ impl Powershell {
...
+fn has_command_script(command_name: &str) -> String {
+    format!(
+        "if (Get-Command -Name {} -ErrorAction SilentlyContinue) {{ Write-Output 'true' }}",
+        powershell_single_quote(command_name)
+    )
+}
+
+fn powershell_single_quote(value: &str) -> String {
+    format!("'{}'", value.replace('\'', "''"))
+}
```
atoerien: Hmm, I hadn't thought about quoting.

It would be better I think if we pass the command name as an argument to the command, this works in pwsh (haven't tested on Windows Powershell):
`pwsh -Command '& { param($command); Write-Host $command }' 'Update-Module'`

You would have to add an args parameter to `build_command_internal` to make that work, but I don't think that is a problem.

niStee:
Latest push addresses the remaining PowerShell review feedback:

- `Powershell::has_command` now passes the command name as a `param($command)` argument instead of interpolating/quoting it into the script.
- The availability probe now uses the same `use_sudo` setting in dry and wet runs, so dry-run and real runs check the same execution path.
- The probe is still read-only and executes during dry-run via `.always()`, while the actual `Update-Module` command remains dry-run gated by the normal executor behavior.

Validated locally:

- `cargo fmt --check`
- `cargo check`
- `cargo test`
- `cargo clippy --all-targets -- -D warnings`
- `cargo build`
- `git diff --check`

Windows/PowerShell runtime probes:

- PowerShell 7 / `pwsh`: `Update-Module` probe returns `true`
- Windows PowerShell 5.1 on my current broken module setup: probe exits 0 with no output, so Topgrade will skip cleanly
- Windows PowerShell 5.1 with `PSModulePath` removed: probe exits 0 with no output, so Topgrade will skip cleanly
- Windows PowerShell 5.1 through `gsudo`: probe exits 0 with no output
- Windows PowerShell 5.1 through native Windows sudo: probe exits 0 with no output
- Elevated PowerShell 7 through `gsudo`: probe returns `true`

CI is green on the latest commit (`1567e86`). Ready for review/merge from my side.

src/steps/powershell.rs
```diff
@@ -157,3 +196,25 @@ impl Powershell {
             .unwrap_or(false)
     }
 }
+
+fn has_command_script() -> &'static str {
```
atoerien: There's no need for this constant string to be a function, it can just be used directly in `has_command`

src/steps/powershell.rs
```diff
@@ -157,3 +196,25 @@ impl Powershell {
...
+            has_command_script(),
+            "& { param($command); if (Get-Command -Name $command -ErrorAction SilentlyContinue) { Write-Output 'true' } }"
+        );
+    }
+
+    #[test]
+    fn has_command_script_does_not_embed_specific_command_name() {
+        assert!(!has_command_script().contains("Update-Module"));
+    }
+}
```
atoerien: Neither of these tests do anything useful, they can just be removed

atoerien (changes requested):
Just two final nits, otherwise all good - tested everything seems to be working on Windows with `pwsh` and `powershell` (both with a working `Update-Module`)

niStee:
@atoerien thx for the feedback.

GideonBear (approved):

GideonBear:
Thank you @niStee and @atoerien!

PR #2020, opened 2026-05-15
fix(terminal): only accept actual keypresses in menus

### LINKED ISSUE #2016: Topgrade should not treat (pasted) text as keybinds in the end and retry menu
## Checklist

## Erroneous Behavior

I pasted some text by mistake in the end menu, which contained the letter "r", which then lead to an unwanted reboot.

## Expected Behavior

Topgrade should not treat any text as commands and instead only execute the corresponding action on actual keypress.

## Steps to reproduce

1. run topgrade with `-k` option
2. paste in some text that contains one of "

## Possible Cause (Optional)

Topgrade treats text the same as keypresses.

- [ ] Yes
- [ ] No

- [ ] Yes
- [x] No

remote host?

- [ ] Yes
- [ ] No

```toml

```

## Additional Details

- Operation System/Version
  

Arch Linux (btw)

- Installation
  

<details>
<pre>

</pre>
</details>

uwuclxdy:
## What does this PR do

Fixes https://github.com/topgrade-rs/topgrade/issues/2016
Uses crossterm crate to get the keypress events and discards any other ones (like paste). Enter is used for default selection in `prompt_yesno()`.

## Standards checklist

- [ ] The PR title is descriptive
- [x] ~If this PR introduces new user-facing messages they are translated~

### AI involvement

Consulted with Claude about it and modified some code snippets that it gave me. Also for review.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -254,9 +254,9 @@ checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
 
 [[package]]
 name = "aws-lc-rs"
-version = "1.17.0"
+version = "1.16.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5ec2f1fc3ec205783a5da9a7e6c1509cc69dedf09a1949e412c1e18469326d00"
+checksum = "0ec6fb3fe69024a75fa7e1bfb48aa6cf59706a101658ea01bfd33b2b248a038f"
 dependencies = [
  "aws-lc-sys",
  "zeroize",
@@ -264,9 +264,9 @@ dependencies = [
 
 [[package]]
 name = "aws-lc-sys"
-version = "0.41.0"
+version = "0.40.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1a2f9779ce85b93ab6170dd940ad0169b5766ff848247aff13bb788b832fe3f4"
+checksum = "f50037ee5e1e41e7b8f9d161680a725bd1626cb6f8c7e901f91f942850852fe7"
 dependencies = [
  "cc",
  "cmake",
@@ -450,9 +450,9 @@ dependencies = [
 
 [[package]]
 name = "clap_complete"
-version = "4.6.5"
+version = "4.6.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e0a7a9bfdb35811f9e59832f0f05975114d2251b415fb534108e6f34060fd772"
+checksum = "e3e962dae2b1e5007fe9e3db363ddc43a8bf25546d279f7a8a4401204690e80c"
 dependencies = [
  "clap",
 ]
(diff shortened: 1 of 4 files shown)


GideonBear:
The commit message (PR title) will end up in the changelog;
* If there's no user-facing changes, make it a `chore` or `refactor`
* If there are user-facing changes caused by this dependency switch, explain the user-facing changes instead (and leave the implementation in the PR body)

uwuclxdy:
Thank you!! Let me try

src/terminal.rs
```diff
@@ -258,7 +278,18 @@ impl Terminal {
...
+        let _raw_mode_guard = RawTerminalMode::enter()?;
+        loop {
+            let Event::Key(key) = read()? else { continue };
+            if key.kind != KeyEventKind::Press {
+                continue;
+            }
+            break Ok(match key.code {
+                KeyCode::Char(c) => Key::Char(c),
+                KeyCode::Enter => Key::Enter,
+                _ => Key::Unknown,
```
GideonBear: We do not need to translate to `console::Key`; just change the callers to use `crossterm::KeyCode`

src/terminal.rs
```diff
@@ -44,6 +46,24 @@ struct Terminal {
...
+        let guard = Self;
+        crossterm::execute!(io::stdout(), EnableBracketedPaste)?;
+        Ok(guard)
+    }
+}
+
+impl Drop for RawTerminalMode {
+    fn drop(&mut self) {
+        crossterm::execute!(io::stdout(), DisableBracketedPaste).ok();
+        disable_raw_mode().ok();
```
GideonBear:
```suggestion
        disable_raw_mode()?;
```
Specify why the failure is ignored if that's really necessary.

Cargo.toml
```diff
@@ -23,6 +23,7 @@ clap_complete = "4.5"
 clap_mangen = "0.3"
 walkdir = "2.5"
 console = "0.16"
+crossterm = { version = "0.29", default-features = false, features = ["bracketed-paste", "events", "use-dev-tty", "windows"] }
```
GideonBear, line 26: Do we need `use-dev-tty`?

Cargo.lock
```diff
@@ -254,19 +254,19 @@ checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
 
 [[package]]
 name = "aws-lc-rs"
-version = "1.16.3"
+version = "1.17.0"
```
GideonBear: Looks like you also ran a `cargo update`?

GideonBear (changes requested):

Cargo.lock
uwuclxdy: yes mb

Cargo.toml
uwuclxdy: Yes, that just makes crossterm read from `/dev/tty` instead of `stdin` on Unix systems, which previous implementation also did (`console::Term::stdout().read_key()`)

src/terminal.rs
uwuclxdy: yea but you cant do that in `Drop`
```
cannot use the `?` operator in a method that returns `()`
```

GideonBear: Maybe just `.unwrap()` / `.expect()` then?

src/terminal.rs
uwuclxdy: okies

src/terminal.rs
uwuclxdy: bro really dislikes `.ok()` :skull: 

You're abosuletly right—will remember to keep `.ok()` out of this in the future :ok_hand:

GideonBear: fail fast & fail hard :)

GideonBear (approved):

GideonBear:
Thanks!

uwuclxdy:
yayy, thank you, no more reboot by pasting text :D

PR #2044, opened 2026-05-21
feat(vim): add opt-in vim.pack pruning

### LINKED ISSUE #2042: Add opt-in pruning for `vim.pack`

Topgrade recently added support for Neovim's built-in `vim.pack`. Neovim 0.13 adds deletion support for inactive packages via `:packdel` / `vim.pack.del()`.

request is for Topgrade to add opt-in pruning. something like

```toml
[misc]
vim_pack_prune = true
```

Example shape:

```vim
if has("nvim") && exists(":packdel") && g:topgrade_vim_pack_prune
  packdel ++all
endif

if has("nvim")
  lua if vim.pack and next(vim.pack.get(nil, { info = false })) ~= nil then vim.pack.update(nil, { force = true }) end
endif
```

## Why opt-in

Inactive `vim.pack` packages are not always stale. Some user configs could add plugins conditionally based on machine, env vars, local files, or other runtime state. In those cases, automatic pruning could delete a package that is intentionally absent in that Topgrade run.

Probably fits under https://github.com/topgrade-rs/topgrade/issues/515

TyceHerrman:
## What does this PR do

- Adds `[vim] vim_pack_prune`, defaulting to `false`.
- Adds `Config::vim_pack_prune()`.
- When enabled, `src/steps/vim.rs` forwards `TOPGRADE_VIM_PACK_PRUNE=true` to
  the Vim upgrade script.
- In `src/steps/upgrade.vim`, Neovim checks both the opt-in env var and
  `:packdel` availability before pruning.
- Uses `packdel ++all` before `vim.pack.update` to remove inactive `vim.pack`
  packages.
- Updates `config.example.toml` under `[vim]`.
- Does not wire this into global `cleanup = true`.
- The option is opt-in because inactive `vim.pack` packages may be intentionally
  kept by users.

Related issue:

- Closes #2042

## Standards checklist

- [x] PR title is descriptive
- [x] If this PR introduces new user-facing messages, they are translated

Verification already run:

- `cargo test test_default_config`
- `cargo test vim_pack_prune`
- `cargo test steps::vim`
- `cargo test`
- `cargo fmt --check`
- `git diff --check`

### AI involvement

- Codex implemented the code changes, ran the verification commands listed
  above, committed, and pushed the branch.
- Codex initially opened PR #2043, which GitHub Actions closed because the
  repository requires the PR template and does not allow AI agents to interact
  with the repository.
  description.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -392,6 +392,9 @@ pub struct Composer {
 pub struct Vim {
     #[merge(strategy = merge::option::overwrite_none)]
     force_plug_update: Option<bool>,
+
+    #[merge(strategy = merge::option::overwrite_none)]
+    vim_pack_prune: Option<bool>,
 }
 
 #[derive(Deserialize, Default, Debug, Merge)]
@@ -1575,6 +1578,15 @@ impl Config {
             .unwrap_or(false)
     }
 
+    /// Whether to prune inactive `vim.pack` packages in Neovim
+    pub fn vim_pack_prune(&self) -> bool {
+        self.config_file
+            .vim
+            .as_ref()
+            .and_then(|c| c.vim_pack_prune)
+            .unwrap_or(false)
+    }
+
     /// Binaries to exclude from `gup update`
     pub fn gup_exclude(&self) -> &[String] {
         self.config_file
@@ -2382,6 +2394,23 @@ x = "cmd_x"
             config_file: toml::from_str(toml_str).expect("toml parse error"),
             allowed_steps: Vec::new(),
         }
+    }
+
+    #[test]
+    fn test_vim_pack_prune_defaults_to_false() {
+        assert!(!config().vim_pack_prune());
+    }
+
+    #[test]
+    fn test_vim_pack_prune_can_be_enabled() {
+        let config = config_from_toml(
+            r#"
+        [vim]
+        vim_pack_prune = true
+        "#,
+        );
+
+        assert!(config.vim_pack_prune());
     }
 
     #[test]
--- a/src/config.rs
+++ b/src/config.rs
@@ -2397,23 +2397,6 @@ x = "cmd_x"
     }
 
     #[test]
-    fn test_vim_pack_prune_defaults_to_false() {
-        assert!(!config().vim_pack_prune());
-    }
-
-    #[test]
-    fn test_vim_pack_prune_can_be_enabled() {
-        let config = config_from_toml(
-            r#"
-        [vim]
-        vim_pack_prune = true
-        "#,
-        );
-
-        assert!(config.vim_pack_prune());
-    }
-
-    #[test]
     fn test_steps_default_order_without_first_or_last() {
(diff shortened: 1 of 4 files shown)


src/steps/vim.rs
```diff
@@ -157,3 +160,37 @@ pub fn run_voom(ctx: &ExecutionContext) -> Result<()> {
 
     ctx.execute(voom).arg("update").status_checked()
 }
+
+#[cfg(test)]
+mod tests {
+    use super::UPGRADE_VIM;
+
+    #[test]
+    fn vim_pack_update_can_prune_before_updating() {
```
GideonBear: Useless tests

src/config.rs
```diff
@@ -2384,6 +2396,23 @@ x = "cmd_x"
         }
     }
 
+    #[test]
+    fn test_vim_pack_prune_defaults_to_false() {
```
GideonBear: Useless tests

GideonBear:
Can you show this working with the config on and off?

GideonBear (changes requested):

TyceHerrman:
I verified this with two temp Neovim setups, one with pruning off and one with pruning on.

`/private/tmp/topgrade-prune-off`
`/private/tmp/topgrade-prune-on`

Seeded each temp setup with two local dummy `vim.pack` packages, both pointing at this checkout via `file:///Users/<>/GitHub/forks/topgrade`.

- `topgrade-active`: package installed on disk and included in the later `vim.pack.add(...)` call
- `topgrade-inactive`: package installed on disk during setup, then omitted from the later `vim.pack.add(...)` call so Neovim treats it as inactive
- Then I sourced this PR's `src/steps/upgrade.vim` in each temp setup.

With no `TOPGRADE_VIM_PACK_PRUNE` env var (prune off)

```text
vim.pack:   0% Updating (0/2)
vim.pack:  50% Updating (1/2) - topgrade-active
vim.pack: 100% Updating (2/2) - topgrade-inactive
vim.pack: 100% Updating (2/2)
```

`topgrade-inactive` and `topgrade-active` still present


With `TOPGRADE_VIM_PACK_PRUNE=true` (prune on):

```text
vim.pack: Removed plugin: topgrade-inactive
vim.pack:   0% Updating (0/1)
vim.pack: 100% Updating (1/1) - topgrade-active
vim.pack: 100% Updating (1/1)
```

only `topgrade-active` remains.

also when running topgrade locally with my temporary topgrade post-command 

```
local ok, err = pcall(vim.cmd, "packdel ++all")
if not ok then
  if tostring(err):find("E5809: Nothing to remove", 1, true) then
    print("No inactive vim.pack packages to prune")
  else
    error(err)
  end
end
```

it removed actually plugin I was no longer using

src/steps/vim.rs
GideonBear: Still stands; string tests don't do anything useful imo.

GideonBear (changes requested):

GideonBear (approved):

Issue #632, opened 2023-12-29
pnpm upgrade failed


alec-c4:
## Erroneous Behavior
<img width="757" alt="SCR-20231229-myej" src="https://github.com/topgrade-rs/topgrade/assets/25312/c29cad10-d1eb-42f7-8c4d-200181ae879a">

## Expected Behavior
successful upgrade

## Steps to reproduce
just run topgrade from terminal

## Problem persists without calling from topgrade
<img width="693" alt="SCR-20231229-mysj" src="https://github.com/topgrade-rs/topgrade/assets/25312/1bc856ab-5f33-4d80-b4bb-6bd0589c80bb">

- [x] Yes
- [ ] No

## Did you run topgrade through `Remote Execution`

- [ ] Yes
- [x] No

If yes, does the issue still occur when you run topgrade directlly in your
remote host

- [ ] Yes
- [ ] No

## Configuration file (Optional)

```toml

```

## Additional Details
- Operation System/Version
macOS Sonoma 14.2.1 (23C71)

- Installation

```bash
brew install topgrade
```

- Topgrade version (`topgrade -V`)

Topgrade 13.0.0

## Verbose Output (`topgrade -v`)

<details>
<pre>
DEBUG Configuration at /Users/alec/.config/topgrade.toml
DEBUG Version: 13.0.0
DEBUG OS: x86_64-apple-darwin
DEBUG Args { inner: ["topgrade", "-v"] }
DEBUG Binary path: Ok("/usr/local/bin/topgrade")
DEBUG self-update Feature Enabled: false
[...]

PR #2057, opened 2026-05-28
feat(pnpm): use self-update for standalone installs

jdmallen:
## What does this PR do
Addresses part of #632.

A pnpm binary installed via the official `curl | sh` installer registers `@pnpm/exe` in the global manifest as a `file:` link into a mktemp dir. Once that temp dir is cleaned up (e.g. on reboot), `pnpm update -g` can no longer resolve the dangling link and fails with `ERR_PNPM_LINKED_PKG_DIR_NOT_FOUND`, even though pnpm itself works fine.

Detect this case (the only global package is `@pnpm/exe`) and run `pnpm self-update` instead. Corepack-managed installs do not register `@pnpm/exe` globally, so they keep using `pnpm update -g`.

### Output from local test:

**Before:**

```sh
2026-05-28T13:48:58-0400  dev-laptop-t14g5  ~  1
$ topgrade

── 13:51:24 - Sudo ─────────────────────────────────────────────────────────────
[sudo] password for jesse:
Sorry, try again.
[sudo] password for jesse:

# ...

── 13:55:33 - Vim ──────────────────────────────────────────────────────────────
Plugins upgraded

── 13:55:33 - Performant Node Package Manager ──────────────────────────────────
/home/jesse/.local/share/pnpm/global/v11/c6f5-19e6fba36be:
[ERR_PNPM_LINKED_PKG_DIR_NOT_FOUND] Could not install from "/tmp/tmp.xmRqYYgq7B" as it does not exist.

This error happened while installing a direct dependency of /home/jesse/.local/share/pnpm/global/v11/c6f5-19e6fba36be
pnpm failed:
   0: Command failed: `/home/jesse/.local/share/pnpm/bin/pnpm update -g`
   1: `/home/jesse/.local/share/pnpm/bin/pnpm` failed: exit status: 1

Location:
   src/steps/node.rs:96

── 13:55:51 - GitHub CLI Extensions ────────────────────────────────────────────
[copilot]: already up to date
✓ Successfully checked extension upgrades

# ...

── 13:55:55 - Summary ──────────────────────────────────────────────────────────
System update: OK
Brew: OK
Brew Cask: OK
config-update: OK
snap: OK
Firmware: OK
Flatpak: OK
asdf: OK
oh-my-zsh: OK
Gnome Shell Extensions: OK
atuin: OK
rustup: OK
.NET: OK
Visual Studio Code extensions: OK
pip3: OK
vim: OK
pnpm: FAILED
GitHub CLI Extensions: OK
Claude Code: OK
Powershell Modules Update: OK
```

**After:**

```sh
2026-05-28T14:29:34-0400  dev-laptop-t14g5  ~/git/3rd-party/topgrade (feat/pnpm-standalone-self-update ±)  0
$ ./target/debug/topgrade --only pnpm -y

── 14:30:58 - Sudo ─────────────────────────────────────────────────────────────
[sudo] password for jesse:

── 14:31:01 - Performant Node Package Manager ──────────────────────────────────
Checking for updates...
The currently active pnpm v11.4.0 is already "latest" and doesn't need an update

── 14:31:02 - Summary ──────────────────────────────────────────────────────────
pnpm: OK
```

## Standards checklist

  - _Only adds some debug messages, not user-facing._

### AI involvement
I used Claude Opus 4.8 to identify the original issue and write the first draft of the Rust changes. I reviewed them, made minor changes, and tested locally. I wrote this PR.

### DIFF
--- a/src/steps/node.rs
+++ b/src/steps/node.rs
@@ -99,20 +99,28 @@ impl NPM {
         Ok(())
     }
 
-    /// Detects whether pnpm was installed as a standalone binary, i.e. the only
-    /// globally installed "package" is `@pnpm/exe` itself.
+    /// Detects whether pnpm's global registry is in the fragile/broken state
+    /// produced by the standalone `curl | sh` installer.
     ///
-    /// The official `curl | sh` installer extracts the standalone binary into a
-    /// `mktemp` directory and then runs `pnpm add -g @pnpm/exe@<version>` from
-    /// there, so the global manifest registers `@pnpm/exe` as a `file:` link
-    /// into that temp dir. Once the temp dir is cleaned up (e.g. on reboot),
-    /// `pnpm update -g` can no longer resolve the dangling link and fails with
-    /// `ERR_PNPM_LINKED_PKG_DIR_NOT_FOUND`. For such installs `pnpm self-update`
-    /// is the correct tool. Corepack-managed installs do not register
-    /// `@pnpm/exe` as a global package, so they are unaffected by this check.
+    /// That installer extracts the standalone binary into a `mktemp` directory
+    /// and then runs `pnpm add -g @pnpm/exe@<version>` from there, so the global
+    /// manifest registers `@pnpm/exe` as a `file:` link into that temp dir.
+    /// While the temp dir still exists `pnpm update -g` happens to work, but once
+    /// it is cleaned up (e.g. on reboot) the dangling link makes `pnpm update -g`
+    /// fail outright with `ERR_PNPM_LINKED_PKG_DIR_NOT_FOUND`. Running
+    /// `pnpm self-update` re-registers `@pnpm/exe` against a stable,
+    /// registry-resolved version, permanently repairing the manifest so that a
+    /// subsequent `pnpm update -g` can run normally again.
+    ///
+    /// We detect this by the `file:` version specifier that `pnpm list -g --json`
+    /// reports for `@pnpm/exe` (a healthy install reports a plain semver
+    /// version). `pnpm list -g --json` itself keeps working in the broken state,
+    /// so this probe is reliable. Corepack-managed installs do not register
+    /// `@pnpm/exe` as a global package at all, so they are unaffected (and
+    /// `pnpm self-update` does not support them anyway).
     ///
     /// See <https://github.com/topgrade-rs/topgrade/issues/632>.
-    fn is_standalone_pnpm(&self, ctx: &ExecutionContext) -> bool {
+    fn pnpm_global_registry_is_broken(&self, ctx: &ExecutionContext) -> bool {
         if !matches!(self.variant, NPMVariant::Pnpm) {
             return false;
         }
@@ -125,13 +133,15 @@ impl NPM {
         let stdout = match output {
             Ok(output) => output.stdout,
             Err(err) => {
-                debug!("Could not list global pnpm packages, assuming a non-standalone install: {err}");
+                debug!("Could not list global pnpm packages, assuming a healthy install: {err}");
                 return false;
             }
         };
 
         // `pnpm list -g --json` prints an array of project objects, each with a
-        // `dependencies` map keyed by package name.
+        // `dependencies` map keyed by package name. The standalone installer
+        // registers `@pnpm/exe` with a `file:` version specifier pointing into a
+        // temp dir; a healthy registry install reports a plain semver version.
         let parsed: serde_json::Value = match serde_json::from_str(&stdout) {
             Ok(parsed) => parsed,
             Err(err) => {
@@ -140,16 +150,16 @@ impl NPM {
             }
         };
 
-        let global_packages: Vec<&str> = parsed
+        parsed
             .as_array()
             .into_iter()
             .flatten()
             .filter_map(|project| project.get("dependencies"))
             .filter_map(|dependencies| dependencies.as_object())
-            .flat_map(|dependencies| dependencies.keys().map(String::as_str))
-            .collect();
-
-        global_packages == ["@pnpm/exe"]
+            .filter_map(|dependencies| dependencies.get("@pnpm/exe"))
+            .filter_map(|exe| exe.get("version"))
+            .filter_map(|version| version.as_str())
+            .any(|version| version.starts_with("file:"))
     }
 
     /// Update a standalone pnpm binary via `pnpm self-update`.
@@ -434,11 +444,15 @@ pub fn run_pnpm_upgrade(ctx: &ExecutionContext) -> Result<()> {
 
     print_separator(t!("Performant Node Package Manager"));
 
-    // For a standalone pnpm install, `pnpm update -g` is the wrong tool and can
-    // fail outright, so update the binary itself instead. See `is_standalone_pnpm`.
-    if pnpm.is_standalone_pnpm(ctx) {
-        debug!("Detected a standalone pnpm install, using `pnpm self-update`");
-        return pnpm.self_update(ctx);
+    // A standalone (`curl | sh`) pnpm install registers `@pnpm/exe` as a `file:`
+    // link into a temp dir, which eventually breaks `pnpm update -g`. Repair the
+    // registry with `pnpm self-update` first, then fall through to the normal
+    // `pnpm update -g` so every global package (including pnpm itself) is still
+    // updated, which is this step's original purpose. See
+    // `pnpm_global_registry_is_broken`.
+    if pnpm.pnpm_global_registry_is_broken(ctx) {
+        debug!("Detected a broken standalone pnpm registry; running `pnpm self-update` to repair it");
+        pnpm.self_update(ctx)?;
     }
 
     #[cfg(target_os = "linux")]
--- a/src/steps/node.rs
+++ b/src/steps/node.rs
@@ -99,6 +99,65 @@ impl NPM {
         Ok(())
     }
 
+    /// Detects whether pnpm was installed as a standalone binary, i.e. the only
+    /// globally installed "package" is `@pnpm/exe` itself.
+    ///
+    /// The official `curl | sh` installer extracts the standalone binary into a
+    /// `mktemp` directory and then runs `pnpm add -g @pnpm/exe@<version>` from
+    /// there, so the global manifest registers `@pnpm/exe` as a `file:` link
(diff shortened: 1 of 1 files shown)


uwuclxdy:
Just a sidenote; you removed the magic marker at the bottom (or claude left it out xd).

Reopening since this seems valid and I have the same issue.

src/steps/node.rs
```diff
@@ -375,6 +434,13 @@ pub fn run_pnpm_upgrade(ctx: &ExecutionContext) -> Result<()> {
 
     print_separator(t!("Performant Node Package Manager"));
 
+    // For a standalone pnpm install, `pnpm update -g` is the wrong tool and can
+    // fail outright, so update the binary itself instead. See `is_standalone_pnpm`.
+    if pnpm.is_standalone_pnpm(ctx) {
+        debug!("Detected a standalone pnpm install, using `pnpm self-update`");
+        return pnpm.self_update(ctx);
+    }
```
uwuclxdy: correct me if im wrong but there really should be a way to update pnpm packages even when in standalone..

uwuclxdy:
Can you explain how are pnpm packages updated when pnpm is standalone?

jdmallen:
> Just a sidenote; you removed the magic marker at the bottom (or claude left it out xd).
> 
> Reopening since this seems valid and I have the same issue.

Hah, weird. Nah, I wrote the PR message. I try to ensure that I only use Claude as a tool and not a crutch, though my efforts are not always perfect.

> Can you explain how are pnpm packages updated when pnpm is standalone?

I think it's the same. What happens with their `sh|curl` installer is it uses `mktemp` to create a temporary install directory and then registers `@pnpm/exe` as a `file:` link to that transient directory. After a reboot, `pnpm update -g` fails to upgrade `pnpm` itself, since that directory is gone. After a single execution of `pnpm self-upgrade`, `@pnpm/exe` is updated to a stable version number, which allows `pnpm update -g` to run on subsequent attempts.

What this means is that we _could_ just close this PR and say it's strictly upstream and recommend users to just run `pnpm self-update` once to fix the global registry. Alternatively, we can check on behalf of the users, and I can update this PR to actually perform the global package upgrade, which I think I might have accidentally squashed.

uwuclxdy:
Hm okay, I guess it makes a bit more sense now. But still; let's just run self upgrade and then pnpm packages upgrade (you're saying it will work after) - remove `return` on line https://github.com/topgrade-rs/topgrade/pull/2057/changes#diff-630be80f2c58b3a64da62297ac4abb30a765d139c2dbd18db246d61c3773e150R441 and try. Also, I think `is_standalone_pnpm` could be done in a much cleaner way than this.

Also, I managed to resolve my issue by uninstalling pnpm with pacman (arch package manager) and installing it with the official installer after. Then `pnpm setup --force` and topgrade worked w/o issues. So I'm still leaning to the "it's an upstream thing" side of things.

jdmallen:
I cleaned it up, restored the original behavior, and added unit tests. Please let me know.

jdmallen:
Also I have zero problem with just closing this and letting `pnpm` fix their installer on their end.

uwuclxdy:
I'm sorry but I think this is just a misconfiguration thing. I just ran `topgrade --only pnpm` to test it and not only were my global pnpm packages updated, pnpm itself seems to be included in that as well. 
We can always add self update if it turns out it's needed, but IMO this PR is fixing a non existing issue, sorry.

jdmallen:
> I'm sorry but I think this is just a misconfiguration thing. I just ran `topgrade --only pnpm` to test it and not only were my global pnpm packages updated, pnpm itself seems to be included in that as well. We can always add self update if it turns out it's needed, but IMO this PR is fixing a non existing issue, sorry.

Worth noting that that does make sense provided you have successfully updated `pnpm` at least _once_ with `pnpm self-update` since using the `curl|sh` installer. After it's run once, the issue goes away, because the registry no longer uses a `file:` link. As I said, this is really just a bandaid for an issue with `pnpm`'s installer, not an inherent problem with `topgrade`. However, this fix would make `topgrade` more robust`, especially given that it's goal is to a be a one-stop-shop to update all the things.

Still, I'm still OK with just closing this and leaving a detailed note on the issue :)

jdmallen:
Also `pnpm` does update itself with `pnpm update -g` whenever `@pnpm/exe` is in the registry, as it likely is on your system. That was the source of the issue: `@pnpm/exe` is pointing to a transient `file://` link, put there by the `curl|sh` installer. So, yes, a configuration issue, but one that's created by `pnpm`, not by the users. Once `@pnpm/exe` points to a semver number, everything is peachy.

uwuclxdy:
Aight yea this technically is an upstream issue but maybe worth handling it in topgrade like you suggested. The code is still an issue; your "cleanup" just renamed stuff and added (IMO unneeded) comments and tests. Lets just do: pnpm -> (is standalone? -> self-upgrade) -> upgrade -g

jdmallen:
> Aight yea this technically is an upstream issue but maybe worth handling it in topgrade like you suggested. The code is still an issue; your "cleanup" just renamed stuff and added (IMO unneeded) comments and tests. Lets just do: pnpm -> (is standalone? -> self-upgrade) -> upgrade -g

I can certainly delete the tests if you'd like, just thought they'd help.

The flow you suggested is exactly the one that the latest update I pushed takes.

1. [Checks](https://github.com/jdmallen/topgrade/blob/a7c114a6209ff5740e0eb7ef044be9b53afbba49/src/steps/node.rs#L461) to see if the pnpm install is broken with [this](https://github.com/jdmallen/topgrade/blob/a7c114a6209ff5740e0eb7ef044be9b53afbba49/src/steps/node.rs#L123-L142) helper function.
2. If so, [runs](https://github.com/jdmallen/topgrade/blob/a7c114a6209ff5740e0eb7ef044be9b53afbba49/src/steps/node.rs#L463) `pnpm self-update` with [this](https://github.com/jdmallen/topgrade/blob/a7c114a6209ff5740e0eb7ef044be9b53afbba49/src/steps/node.rs#L145-L148) helper function.
3. Then [updates](https://github.com/jdmallen/topgrade/blob/a7c114a6209ff5740e0eb7ef044be9b53afbba49/src/steps/node.rs#L466-L469) as normal.

My last run at this [returned out](https://github.com/jdmallen/topgrade/blob/7388bce26a608da22120165758beb8c132082217/src/steps/node.rs#L441) of the update process if pnpm was broken with just the self-update, not updating the global packages.

One note on the check itself: rather than just asking "is it standalone," I check for the actual broken `file:` link, so a healthy standalone install (where `@pnpm/exe` is already a semver) isn't needlessly self-updated on every run. That's also what the renamed helper is doing, addressing your "could be cleaner" note about the old `is_standalone_pnpm`.

jdmallen:
FWIW, I added a comment to the relevant open issue on the `pnpm` side :smile:: https://github.com/pnpm/pnpm/issues/11473#issuecomment-4619671051

src/steps/node.rs
```diff
@@ -375,6 +453,16 @@ pub fn run_pnpm_upgrade(ctx: &ExecutionContext) -> Result<()> {
...
     print_separator(t!("Performant Node Package Manager"));
 
+    // A standalone (`curl | sh`) pnpm install registers `@pnpm/exe` as a `file:`
+    // link into a temp dir, which eventually breaks `pnpm update -g`. Repair the
+    // registry with `pnpm self-update` first, then fall through to the normal
+    // `pnpm update -g` so every global package (including pnpm itself) is still
+    // updated, which is this step's original purpose.
+    if pnpm.pnpm_global_registry_is_broken(ctx) {
+        debug!("Detected a broken standalone pnpm registry; running `pnpm self-update` to repair it");
+        pnpm.self_update(ctx)?;
```
uwuclxdy, line 463:
```suggestion
        if let Err(err) = pnpm.self_update(ctx) {
            warn!("`pnpm self-update` failed, continuing with `pnpm update -g`: {err}");
        }
```

`?` kills the whole pnpm step if self-update fails. but if the temp dir's still around, the registry reads as broken while update -g would actually still work, so one bad self-update skips an update that would have been fine. just log it and move on, the repair's best-effort.

src/steps/node.rs
```diff
@@ -99,6 +99,54 @@ impl NPM {
...
+    /// subsequent `pnpm update -g` can run normally again.
+    ///
+    /// The below detects this by the `file:` version specifier that
+    /// `pnpm list -g --json` reports for `@pnpm/exe` (a healthy install reports
+    /// a plain semver version). `pnpm list -g --json` itself keeps working in
+    /// the broken state, so this probe is reliable. Corepack-managed installs do
+    /// not register `@pnpm/exe` as a global package at all, so they are
+    /// unaffected (and `pnpm self-update` does not support them anyway).
+    ///
+    /// See <https://github.com/topgrade-rs/topgrade/issues/632>.
```
uwuclxdy, line 122:
```suggestion
    /// Detects pnpm's broken global registry, produced by the standalone
    /// `curl | sh` installer: it registers `@pnpm/exe` as a `file:` link into a
    /// `mktemp` dir, so once that dir is cleaned up (e.g. on reboot) `pnpm update
    /// -g` fails with `ERR_PNPM_LINKED_PKG_DIR_NOT_FOUND`. A one-off `pnpm
    /// self-update` re-points `@pnpm/exe` at a registry semver and repairs it.
    ///
    /// Probed via the `file:` specifier in `pnpm list -g --json` (which keeps
    /// working in the broken state). Corepack installs don't register
    /// `@pnpm/exe` at all, so they're unaffected.
    ///
    /// See <https://github.com/topgrade-rs/topgrade/issues/632>.
```

src/steps/node.rs
```diff
@@ -99,6 +99,54 @@ impl NPM {
...
+    fn pnpm_global_registry_is_broken(&self, ctx: &ExecutionContext) -> bool {
+        if !matches!(self.variant, NPMVariant::Pnpm) {
+            return false;
+        }
+
+        let output = ctx
+            .execute(&self.command)
+            .always()
+            .args(["list", "-g", "--depth=0", "--json"])
```
uwuclxdy, line 131:
```suggestion
            .args(["list", "-g", "--depth", "0", "--json"]
```

uwuclxdy (changes requested):
looked into this again, sorry for the wait. i think this is good to have in topgrade, only a few fixes i'd want before merge.

Issue #2045, opened 2026-05-21
claude_code: `claude plugin list --json` output truncated at 64 KiB causes `EOF while parsing a string`


xJTCx:
## Summary

The `claude_code` step fails with `EOF while parsing a string` when `claude plugin list --json` produces more than **65 536 bytes (64 KiB)** of output. The JSON is well-formed but is truncated mid-string before being handed to `serde_json::from_str`.

On a workstation with a moderate number of installed Claude plugins this is easily exceeded — my output is 78 KB.

## Repro

1. Install enough Claude Code plugins/marketplaces that `claude plugin list --json` exceeds 64 KiB of output:
   ```
   $ claude plugin list --json | wc -c
   78390
   ```
2. Run `topgrade`.
3. `claude update` and `claude plugin marketplace update` succeed, then the step fails:
   ```
   Claude Code failed:
      0: The output of `claude plugin list --json` changed: json output is invalid or does not match expected structure. ...
      1: EOF while parsing a string at line 2116 column 45

   Location:
      src/steps/generic.rs:2129
   ```

## Diagnosis

The error position 2116:45 in my output corresponds to **byte offset exactly 65 536** — i.e., the 64 KiB boundary, which happens to be the default pipe buffer size on macOS / Linux:

```
$ awk 'NR<2116{total+=length+1} NR==2116{print total+45; exit}' claude-plugins.json
65536
```

The raw JSON itself is valid (`jq -e . claude-plugins.json` passes), and `claude plugin list --json` does not stream / chunk its output. The truncation is happening on topgrade's side while collecting the child process's stdout.

The 64 KiB symmetry strongly suggests the helper that collects stdout (`output_checked_utf8()` at the call site in `src/steps/generic.rs` near the `serde_json::from_str(&output.stdout)` line referenced in the error — function `run_claude_code`) is reading only a single pipe-buffer's worth instead of draining to EOF, or is otherwise capping at 64 KiB.

## Environment

- topgrade `17.5.1`
- claude `2.1.147` (also reproduced on 2.1.146)
- macOS Darwin 25.5.0, aarch64
- Installed claude plugin count: enough to push the JSON past 64 KiB; see byte count above.

## Workaround

Disable the built-in `claude_code` step and replace it with a custom command that does the actual upgrade work and skips `plugin list --json`:

```toml
disable = ["claude_code", ...]

[commands]
"Claude Code" = "claude update && claude plugin marketplace update"
```

[...]

PR #2062, opened 2026-06-01
fix(claude-code): don't fail with EOF on long output

uwuclxdy:
## What does this PR do

Maybe fixes a macos specific issue https://github.com/topgrade-rs/topgrade/issues/2045 by streaming stdout into `serde_json::from_reader` over a piped `BufReader`.

## Standards checklist

### AI involvement

this pr was entirely ai generated

### DIFF
--- a/src/executor.rs
+++ b/src/executor.rs
@@ -107,6 +107,14 @@ impl Executor {
         self
     }
 
+    /// See `std::process::Command::stdout`
+    pub fn stdout<T: Into<Stdio>>(&mut self, cfg: T) -> &mut Executor {
+        if let Executor::Wet(c) | Executor::Damp(c) = self {
+            c.stdout(cfg);
+        }
+        self
+    }
+
     #[allow(dead_code)]
     /// See `std::process::Command::remove_env`
     pub fn env_remove<K>(&mut self, key: K) -> &mut Executor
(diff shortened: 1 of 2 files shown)


GideonBear:
@xJTCx Could you check if this fixes #2045 for you?
`cargo install --git https://github.com/uwuclxdy/topgrade --branch claude-eof-fix && ~/.cargo/bin/topgrade -v --only claude_code`

xJTCx:
Thanks for the follow-up, and for the PR.

One observation in case it helps narrow this down: the truncation looks like it may be on the `claude` side rather than in how topgrade reads the output.

- `claude plugin list --json` written to a regular file: complete, valid JSON (~78 KB).
- The same command through a pipe (read continuously, so not a reader-side buffering problem): truncated at exactly 65,536 bytes, mid-string, so the JSON is invalid.

So `claude` appears to cap its own stdout at 64 KiB when stdout is not a TTY. If that holds, reading the pipe with `serde_json::from_reader` would still receive the truncated JSON, since it is still a pipe. Routing the step's output to a temp file and reading that back (file output is complete here) may be more robust.

uwuclxdy:
okay apparently 65536 bytes is exactly the macOS kernel pipe buffer cap. @GideonBear should we fix this with temp file as suggested?

GideonBear (approved):

GideonBear:
@xJTCx Can you test this?

uwuclxdy:
@xJTCx do you still have this issue and are you able to test if this pr fixes it for you?

PR #2063, opened 2026-06-01
fix(brew): don't panic when linuxbrew owner uid has no passwd entry

### LINKED ISSUE #2055: topgrade is failing to update Homebrew on Linux if the user doesn't own it
## Checklist

## Erroneous Behavior

After updating Topgrade on Bazzite to the latest possible version (distributed via Terra repo) some users have started to experience Topgrade crashes if the user doesn't own `/home/linuxbrew/.linuxbrew` folder which seems to be the cause of the crash. It usually happens on multiple users setup since Homebrew installs itself to previously mentioned path and assigning ownership to the user who performed the installation. Branches 16.x.x and 15.x.x, previously used on Bazzite, did not have this problem in a first place.

```
Message:  this user should exist
Location: src/steps/os/unix.rs:314
error: Recipe `update` failed with exit code 101
```

For some users logs may reference a different code snippet: `src/steps/os/unix.rs:64` so it may be a good starting point.

## Expected Behavior

The update should have proceed further even if Homebrew was owned by a different user.

## Steps to reproduce

```bash
$ /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)" # or install the system with homebrew preinstalled
# create a new user and login into it
$ topgrade
```

Running `brew update` directly doesn't cause the problem apparently, at least by looking at reported issues on Bazzite repo.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

remote host?

- [ ] Yes
- [ ] No

## Additional Details

- Operation System/Version
  Bazzite 44

- Installation
  Provided with the system

  topgrade 17.5.1

Personally I can't provide it since I never experienced it myself.

Reported issues on Bazzite repo: https://github.com/ublue-os/bazzite/issues/4698 https://github.com/ublue-os/bazzite/issues/4900 https://github.com/ublue-os/bazzite/issues/4920

uwuclxdy:
## What does this PR do

Fixes #2055; topgrade doesn't panic anymore when `/home/linuxbrew/.linuxbrew` is owned by a UID without a passwd entry (e.g. Bazzite). It runs brew as the current user instead of crashing with "this user should exist". 

## Standards checklist

	- testers appreciated
	- confirmed working: https://github.com/topgrade-rs/topgrade/issues/2055#issuecomment-4595330691

### AI involvement

Only `src/steps/os/unix.rs:59`

### DIFF
--- a/src/steps/os/unix.rs
+++ b/src/steps/os/unix.rs
@@ -56,14 +56,14 @@ fn brew_get_sudo() -> Option<String> {
     #[cfg(target_os = "linux")]
     {
         let sudo_uid = brew_linux_sudo_uid();
-        // if brew is owned by another user, execute "sudo -Hu <uid> brew update"
+        // if brew is owned by another user, execute "sudo -Hu <user> brew update"
         if let Some(user_id) = sudo_uid {
             let uid = nix::unistd::Uid::from_raw(user_id);
-            let user = nix::unistd::User::from_uid(uid)
-                .expect("failed to call getpwuid()")
-                .expect("this user should exist");
-
-            return Some(user.name);
+            match nix::unistd::User::from_uid(uid) {
+                Ok(Some(user)) => return Some(user.name),
+                Ok(None) => warn!("no matching owner of linuxbrew found; running brew as the current user"),
+                Err(err) => warn!("getpwuid() failed for UID {user_id}: {err}; running brew as the current user"),
+            }
         }
     }
     None


GideonBear (approved):
Thanks!
Tbh I have never fully understood this code (only moved it around so it could be used for both formulea and casks); but this looks non-destructive. The warn is appreciated, we could potentially remove it in the future with a better understanding of how (linux)brew works.

PR #2067, opened 2026-06-02
refactor: replace console with crossterm

### LINKED ISSUE #2022: Move from console to crossterm
If there ends up being a reason why crossterm cannot fully replace console, we can keep both.

uwuclxdy:
## What does this PR do

Removes `console` crate in favor of `crossterm`. Previous behaviour was kept the same or improved.
Closes https://github.com/topgrade-rs/topgrade/issues/2022

`style` helper fn was needed because crossterm doesn't do that on it's own (it technically does but not fully like console used to). Also uses `unicode-width` to ensure translations don't break anything. Additionally, replaces some `println!` with `write_fmt` and `write_all` because of consistency and to avoid panics on failed writes.

## Standards checklist

	- tested what i could and found no regressions

### AI involvement

About half of this was ~inspired~ written by Claude :broken_heart:


uwuclxdy:
~Might fix #2160 as well but need to find a reliable way to reproduce it first.~

PR #2098, opened 2026-06-20
fix: improve default step order

GideonBear:


uwuclxdy (approved):
maybe mise env thing can be fixed by just putting it at the top :thinking:

GideonBear:
> maybe mise env thing can be fixed by just putting it at the top 🤔

No, because the problem is that PATH doesn't update, so even if the other step runs after it, it will try to call the old version of the tool.

PR #2105, opened 2026-06-20
NOSQUASH: feat(jetbrains-toolbox): remove download cache if cleanup = true

GideonBear:
@uwuclxdy One set of eyes before I commit a `remove_dir_all`. Any reason why people would *not* want Topgrade to remove the download cache?


uwuclxdy (approved):
i cant think of any reason why someone would want to keep specifically the jetbrains cache and delete the others. this is good imo

PR #2109, opened 2026-06-22
feat(git): add config option to fetch default branch instead of failing when on another branch

### LINKED ISSUE #2107: Fetch default branch instead of failing when on another branch
## Checklist

Topgrade should be able to update a repo's default branch even when a different branch is checked out, instead of failing.

Right now the git step runs `git pull --ff-only` on whatever branch is checked out. If a repo is on a feature branch (with no upstream) or a detached HEAD, the pull fails with *"There is no tracking information for the current branch,"* and the repo is marked as failed — even though the default branch could have been updated.

Same problem as #69, which was closed as a duplicate of #86. But #86 was an unrelated panic, so this was never fixed.

**Proposal:** add an opt-in option `[git] fallback_to_fetch_default` (default `false`). When it's on and the repo is not on its default branch, run this instead of failing:

```
git fetch <remote> <default>:<default> --recurse-submodules
```

The remote and default branch are found per-repo from `origin/HEAD`. This updates the local default branch, so a later `git checkout main` already has the new commits. If the default branch can't fast-forward (it has diverged), the error is shown as usual — nothing is forced or lost.

## More information

To test from a side branch: set `fallback_to_fetch_default = true` under `[git]`, check out a feature branch in a repo, run `topgrade --only git`, and confirm the default branch updates instead of failing.

benner:
## What does this PR do

Closes #2107.

When a Git repo is checked out on a non-default branch, Topgrade's git step runs
`git pull --ff-only` and **fails** — and the local default branch stays behind, so
a later `git switch main` misses the upstream commits.

This adds an opt-in `fallback_to_fetch_default = true/false` to the `[git]` section.
When enabled and a repo is **not** on its default branch (a different branch, or a
detached HEAD), Topgrade fetches the default branch into its local ref
(`git fetch <remote> <default>:<default> --recurse-submodules`) instead of pulling,
so the local default stays current without touching HEAD or the working tree. The remote and default
branch are detected per repo (preferring `origin`, reading `origin/HEAD`, falling
back to `ls-remote`). The refspec is non-forced, so a diverged local default still
fails rather than losing commits. Default behavior is unchanged.

The option is documented in `config.example.toml`, and the branch-decision logic
is covered by unit tests.

## Standards checklist

Test — a repo checked out on `feature/wip` while its `main` is two commits behind
the remote, with `fallback_to_fetch_default = true`:

Dry run (prints "Would fetch" because the repo is off its default branch):

```
―― Git repositories ――
Would fetch /tmp/tg-demo/work
```

Real run (local `main` fast-forwards while HEAD stays on `feature/wip`):

```
―― Git repositories ――

Only updated repositories will be shown...

Changed /tmp/tg-demo/work
6f4577b Fix a bug on main
5a8c237 Add new upstream feature
```

Before: local `main` at `8f84f60`, HEAD on `feature/wip`.
After:  local `main` at `6f4577b`, HEAD still on `feature/wip`, working tree clean.

### AI involvement

I designed the feature and reviewed and tested the change; AI wrote most of the code.

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -284,6 +284,11 @@
 # Whether to perform a `git fetch` instead of `git pull`
 # fetch_only = false
 
+# When a repo is not on its default branch, fetch the default branch (advancing
+# its local ref) instead of failing. The remote and default branch are detected
+# per repo. Has no effect on repos already on their default branch.
+# fallback_to_fetch_default = false
+
 [windows]
 # Manually select Windows updates
 # accept_all_updates = false
--- a/src/config.rs
+++ b/src/config.rs
@@ -90,6 +90,9 @@ pub struct Git {
 
     #[merge(strategy = merge::option::overwrite_none)]
     fetch_only: Option<bool>,
+
+    #[merge(strategy = merge::option::overwrite_none)]
+    fallback_to_fetch_default: Option<bool>,
 }
 
 #[derive(Deserialize, Default, Debug, Merge)]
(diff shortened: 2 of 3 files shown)


GideonBear (approved):
Thanks!

Issue #2111, opened 2026-06-23
Add support for mise on Windows.


teh-hippo:
## Checklist

- [x] I have searched the issue tracker for relevant or duplicate issues.

## I want to suggest some general feature

Topgrade should support use of `mise` on Windows.  `mise` is supported on Unix platforms, but seems to have been missed.  Windows is a supported platform `mise` operates on.

## More information

- [x] I am able and willing to implement this feature myself

PR #2112, opened 2026-06-23
NOSQUASH: feat(mise): support Windows

teh-hippo:
The mise step was gated to Unix. Move it into the shared module and run it everywhere, since mise works on Windows too.

## What does this PR do

This implements #2111.

`mise` is a key piece of my dotfiles setup these days.  I've added `topgrade` recently, but have missed the support on Windows.  The 'fix' is fairly straightforward - and just shifts the module from unix to shared, so its supported on Windows.

## Standards checklist

```
―― mise ――
mise is already up to date
mise All tools are up to date
―― Summary ――
mise: OK
```

### AI involvement
Heavy.  AI did the code move and drove local testing.  I have also evaluated it myself.
The AI-driven testing produced local builds in Linux & Windows, ensured `mise` was called appropriately with a dry-run and real execution just to be sure.

## For new steps

  the underlying command

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -333,9 +333,8 @@
 # (default: false)
 # bump = false
 
-# Number of jobs to run in parallel
-# (default: 4)
-# jobs = 4
+# Number of jobs to run in parallel (mise's default)
+# jobs = 8
 
 # Run interactively
 # (default: false)
--- a/src/steps/generic.rs
+++ b/src/steps/generic.rs
@@ -2355,3 +2355,64 @@ pub fn run_ollama_pull(ctx: &ExecutionContext) -> Result<()> {
 
     pull_result
 }
+
+pub fn run_mise(ctx: &ExecutionContext) -> Result<()> {
+    let mise = require("mise")?;
+
+    print_separator("mise");
+
+    ctx.execute(&mise).args(["plugins", "update"]).status_checked()?;
+
+    let output = ctx
+        .execute(&mise)
+        .args(["self-update"])
+        .output_checked_with(|_| Ok(()))?;
+    let status_code = output
+        .status
+        .code()
+        .ok_or_eyre("Couldn't get status code (terminated by signal)")?;
+    let stderr = std::str::from_utf8(&output.stderr).wrap_err("Expected output to be valid UTF-8")?;
+    if stderr.contains("cannot update") && status_code == 1 {
+        debug!("Mise self-update not available")
+    } else {
+        std::io::stdout().lock().write_all(&output.stdout)?;
+        std::io::stderr().lock().write_all(&output.stderr)?;
+        if status_code != 0 {
+            return Err(StepFailed.into());
+        }
+    }
+
+    let mut cmd = ctx.execute(&mise);
+
+    cmd.arg("upgrade");
+
+    if ctx.config().mise_interactive() {
+        cmd.arg("--interactive");
+    }
+
+    if ctx.config().mise_bump() {
+        cmd.arg("--bump");
+    }
+
+    if ctx.config().mise_silent() {
+        cmd.arg("--silent");
+    }
+
+    if ctx.config().mise_quiet() {
+        cmd.arg("--quiet");
+    }
+
+    if ctx.config().mise_verbose() {
+        cmd.arg("--verbose");
+    }
+
+    if ctx.config().yes(Step::Mise) {
+        cmd.arg("--yes");
+    }
+
+    if ctx.config().mise_jobs() != 4 {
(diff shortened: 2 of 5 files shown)


uwuclxdy (approved):

src/steps/generic.rs
```diff
@@ -2355,3 +2355,65 @@ pub fn run_ollama_pull(ctx: &ExecutionContext) -> Result<()> {
...
+        .output_checked_with(|_| Ok(()))?;
+    let status_code = output
+        .status
+        .code()
+        .ok_or_eyre("Couldn't get status code (terminated by signal)")?;
+    let stderr = std::str::from_utf8(&output.stderr).wrap_err("Expected output to be valid UTF-8")?;
+    if stderr.contains("cannot update") && status_code == 1 {
+        debug!("Mise self-update not available")
+    } else {
+        // Write the output
```
```suggestion
```

nit: probably not needed

src/steps/generic.rs
```diff
@@ -2355,3 +2355,65 @@ pub fn run_ollama_pull(ctx: &ExecutionContext) -> Result<()> {
...
+        std::io::stdout().write_all(&output.stdout)?;
+        std::io::stderr().write_all(&output.stderr)?;
```
```suggestion
        std::io::stdout().lock().write_all(&output.stdout)?;
        std::io::stderr().lock().write_all(&output.stderr)?;
```

unless there's a reason not to do this

src/steps/generic.rs
```diff
@@ -2355,3 +2355,65 @@ pub fn run_ollama_pull(ctx: &ExecutionContext) -> Result<()> {
...
+        cmd.arg("--verbose");
+    }
+
+    if ctx.config().yes(Step::Mise) {
+        cmd.arg("--yes");
+    }
+
+    if ctx.config().mise_jobs() != 4 {
+        cmd.args(["--jobs", &ctx.config().mise_jobs().to_string()]);
+    }
```
uwuclxdy: unrelated; this `!= 4` guard is pre-existing (carried over from #1548) and is worth a separate PR to fix.

the "4" was meant to mirror mise's own `--jobs` default, only forwarding the flag when the user overrides it. however, mise's default is now 8 (https://mise.jdx.dev/configuration/settings.html#jobs). what happens is:

- unset -> `mise_jobs()` returns 4 -> flag not passed -> mise uses 8, yet config.example.toml documents the default as 4
- `jobs = 4` -> also suppressed -> user still gets 8; you can't actually select 4 jobs through topgrade

forwarding based on whether the user set the value at all fixes both cases and survives mise changing its default again:

```rs
// src/config.rs
pub fn mise_jobs(&self) -> Option<u32> {
    self.config_file.mise.as_ref().and_then(|mise| mise.jobs)
}
// src/steps/generic.rs
if let Some(jobs) = ctx.config().mise_jobs() {
    cmd.args(["--jobs", &jobs.to_string()]);
}
```

and drop (default: 4) from the `[mise]` block in `config.example.toml` since the effective default is mise's own. followup PR appreciated after this one gets merged

uwuclxdy (commented):
this is all pre-existing tho so it should be fine regardless

src/steps/generic.rs
teh-hippo: Fixed

src/steps/generic.rs
teh-hippo: Fixed

src/steps/generic.rs
teh-hippo: Makes sense to fix while we're here. I've done it in a separate commit.

uwuclxdy: thanks, then @GideonBear NOSQUASH this maybe?

uwuclxdy:
thanks, lgtm!

config.example.toml
```diff
@@ -334,8 +334,8 @@
 # bump = false
 
 # Number of jobs to run in parallel
-# (default: 4)
-# jobs = 4
+# (defaults to mise's own default)
```
uwuclxdy:
```suggestion
# Number of jobs to run in parallel (mise's default)
```

amend this to the latest commit pls

teh-hippo: Yeah, shouldn't have missed that one - apologies

GideonBear:
ae21b6b8696c0b2aaa1442d3b557180c9a690528 diff during move:
```diff
20,22c20,21
<         // Write the output
<         io::stdout().write_all(&output.stdout)?;
<         io::stderr().write_all(&output.stderr)?;
---
>         std::io::stdout().lock().write_all(&output.stdout)?;
>         std::io::stderr().lock().write_all(&output.stderr)?;
```

Thanks both!

PR #2114, opened 2026-06-23
fix(skills): find lock file via XDG_STATE_HOME

### LINKED ISSUE #2110: skills step doesn't check for lock file in XDG_STATE_HOME alternate location
## Checklist

## Erroneous Behavior

When `XDG_STATE_HOME` is set, `npx skills` writes its global lock file to `$XDG_STATE_HOME/skills/.skill-lock.json` instead of `~/.agents/.skill-lock.json` (see: [`getSkillLockPath()`](https://github.com/vercel-labs/skills/blob/main/src/skill-lock.ts#L67-L74)). Topgrade always checks the hard-coded `~/.agents/.skill-lock.json` path and skips the step with `No ~/.agents/.skill-lock.json found`, even when skills are installed.

## Expected Behavior

The `skills` step should execute even if XDG_STATE_HOME is set and the .skill-lock.json file is located there.

## Steps to reproduce

1. export `XDG_STATE_HOME` (eg to `~/.local/state`)
2. install a skill globally using `npx skills add ...`
3. Run `topgrade --only skills -v` observe the "Skills: SKIPPED: No ~/.agents/.skill-lock.json found"

## Possible Cause (Optional)

[`src/steps/generic.rs#L2220`](https://github.com/topgrade-rs/topgrade/blob/main/src/steps/generic.rs#L2220) hard-codes the lock file path:

```rust
let skill_lock = HOME_DIR.join(".agents").join(".skill-lock.json");
```

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

## Additional Details

- Operation System/Version: Fedora Linux 44 (WSL2)

- Installation: Homebrew

- Topgrade version (`topgrade -V`): 17.6.2

<details>
<pre>
topgrade 17.6.2
</pre>
</details>

uwuclxdy:
## What does this PR do

Closes #2110.

The step now finds the lock the same way `npx skills` does, with `~/.agents/.skill-lock.json` as the fallback.

## Standards checklist

### AI involvement

For issue reproduction, testing that this PR fixes it and review.

### DIFF
--- a/src/steps/generic.rs
+++ b/src/steps/generic.rs
@@ -2225,7 +2225,8 @@ pub fn run_skills(ctx: &ExecutionContext) -> Result<()> {
     {
         Some(state) => Path::new(state).join("skills"),
         None => HOME_DIR.join(".agents"),
-    }.join(".skill-lock.json");
+    }
+    .join(".skill-lock.json");
 
     if !skill_lock.exists() {
         return Err(SkipStep(format!("No skill lock file found at {}", skill_lock.display())).into());
--- a/src/steps/generic.rs
+++ b/src/steps/generic.rs
@@ -2223,9 +2223,9 @@ pub fn run_skills(ctx: &ExecutionContext) -> Result<()> {
         .as_deref()
         .filter(|state| !state.is_empty())
     {
-        Some(state) => Path::new(state).join("skills").join(".skill-lock.json"),
-        None => HOME_DIR.join(".agents").join(".skill-lock.json"),
-    };
+        Some(state) => Path::new(state).join("skills"),
+        None => HOME_DIR.join(".agents"),
+    }.join(".skill-lock.json");
 
     if !skill_lock.exists() {
         return Err(SkipStep(format!("No skill lock file found at {}", skill_lock.display())).into());
--- a/src/steps/generic.rs
+++ b/src/steps/generic.rs
@@ -2217,9 +2217,16 @@ pub fn run_colima(ctx: &ExecutionContext) -> Result<()> {
 pub fn run_skills(ctx: &ExecutionContext) -> Result<()> {
     let npx = require("npx")?;
 
-    let skill_lock = HOME_DIR.join(".agents").join(".skill-lock.json");
+    let skill_lock = match env::var_os("XDG_STATE_HOME")
+        .as_deref()
+        .filter(|state| !state.is_empty())
+    {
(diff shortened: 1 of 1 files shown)


src/steps/generic.rs
```diff
@@ -2217,9 +2217,16 @@ pub fn run_colima(ctx: &ExecutionContext) -> Result<()> {
 pub fn run_skills(ctx: &ExecutionContext) -> Result<()> {
     let npx = require("npx")?;
 
-    let skill_lock = HOME_DIR.join(".agents").join(".skill-lock.json");
+    let skill_lock = match env::var_os("XDG_STATE_HOME")
```
GideonBear, line 2222: Can we use `XDG_DIRS` here? That would be preferred; but I don't know it it's possible (given that XDG_DIRS might give a result even when the env var is unset.)
We could still do that and fall back if it doesn't exist?

If keeping like it currently is, add a comment saying why we can't use XDG_DIRS

GideonBear (changes requested):

src/steps/generic.rs
uwuclxdy: `XDG_DIRS` doesn't work here, explained in the comment. It falls back to `~/.local/state` when `XDG_STATE_HOME` is unset but we need `~/.agents` as fallback to match npx skills. It's also `#[cfg(unix)]` so it'd break for Windows users. Could switch to an XDG_DIRS-with-fallback however I think current implementation is cleaner.

src/steps/generic.rs
```diff
@@ -2217,9 +2217,18 @@ pub fn run_colima(ctx: &ExecutionContext) -> Result<()> {
...
-    let skill_lock = HOME_DIR.join(".agents").join(".skill-lock.json");
+    // Can't use XDG_DIRS since its `state_dir()` falls back to `~/.local/state` instead
+    // of `~/.agents` and it'd also break on non-unix systems
+    let skill_lock = match env::var_os("XDG_STATE_HOME")
+        .as_deref()
+        .filter(|state| !state.is_empty())
+    {
+        Some(state) => Path::new(state).join("skills").join(".skill-lock.json"),
+        None => HOME_DIR.join(".agents").join(".skill-lock.json"),
+    };
```
```suggestion
        Some(state) => Path::new(state).join("skills"),
        None => HOME_DIR.join(".agents"),
    }.join(".skill-lock.json");
```

GideonBear (approved):

PR #2131, opened 2026-07-04
feat(step): add dprint self-update and global config plugins update

### LINKED ISSUE #2129: Support dprint
## Checklist

https://dprint.dev, https://github.com/dprint/dprint

https://dprint.dev/install/

See if the `dprint` executable is in `$PATH`. Plain `require("dprint")?` would suffice I think.

To upgrade dprint's globally installed plugins:
```
dprint config update --global
```
This will error out if there is no global config file, but that's ok. At time of writing, the exit status from `dprint` for this is 11, ref https://dprint.dev/cli/#exit-codes

dprint also has a self-update feature, https://dprint.dev/cli/#upgrade:
```
dprint upgrade
```
However, this is something that I personally would **not** like topgrade to run, as global dprint install/updates are managed by mise in my setup.

I don't think it does.

  option to skip this?

I don't use them myself, but I believe updating [process plugins](https://dprint.dev/plugins/#process-plugins) yields a checksum confirmation prompt. There is a `--yes` option to skip the prompt (for `dprint config update --global`).

## More information

- [x] I am able and willing to implement this step myself

https://github.com/topgrade-rs/topgrade/pull/2131

scop:
## What does this PR do

Adds support for dprint self and global config plugins updating.

More info and closes https://github.com/topgrade-rs/topgrade/issues/2129

## Standards checklist

### AI involvement

AI was used for refactoring and doing all the translations.

## For new steps

  the underlying command


src/steps/generic.rs
```diff
@@ -2457,3 +2457,36 @@ fn refresh_mise_env(ctx: &ExecutionContext, mise: &Path) -> Result<()> {
     debug!("Refreshed process environment from mise");
     Ok(())
 }
+
+pub fn run_dprint(ctx: &ExecutionContext) -> Result<()> {
+    let dprint = require("dprint")?;
+
+    print_separator("dprint");
+
```
scop, line 2508: I did not implement `dprint upgrade` because as noted in https://github.com/topgrade-rs/topgrade/issues/2129, dprint install/upgrade is managed by mise in my setup, and I don't want the dprint self update to be run.

As an idea how this could be made graceful in Topgrade for all tools supporting self updates: if the found tool is located in a path that contains `/mise/installs/`, do not run the tool's self updates, assume it's taken care of by mise.

uwuclxdy:
Rename this step to "dprint plugins" or something since we might want dprint self update in the future or just to make it clearer what the step does. Other than that it looks good!

scop:
Would it be an issue if we now rename to "dprint plugins" but later want both plugins and self-update to happen within the same "dprint" step for consistency with other tools?

IDK, maybe I'll just add the self update here, guarded by an option similarly as composer has it.

scop:
Self-update added in 250122b6668246098370371d3cee740097536d42

uwuclxdy:
> Would it be an issue if we now rename to "dprint plugins" but later want both plugins and self-update to happen within the same "dprint" step for consistency with other tools?

Having the two as separate steps is better IMO as that is the direction we're heading for v18. Also, bun already does this (i think) the way it should. Self update + the update step split is correct and other steps will be split like that in the future as well (right @GideonBear ?).

So let's do "dprint plugins" and "dprint self update" or something like that (2 steps, would make sense for self update to be first probably).

scop:
Ok, separated in 8943149e9dabe57f1af16639108bb3350edd4512.

In this case, self update before plugins is the right thing to do, as plugins will need recompiling on dprint version changes (`dprint config update` does it).

uwuclxdy:
Nice, thanks!

GideonBear:
> Having the two as separate steps is better IMO as that is the direction we're heading for v18. Also, bun already does this (i think) the way it should. Self update + the update step split is correct and other steps will be split like that in the future as well (right @GideonBear ?).

Basically. I don't think it matters a lot which way we do it now.

src/steps/generic.rs
```diff
@@ -2457,3 +2457,48 @@ fn refresh_mise_env(ctx: &ExecutionContext, mise: &Path) -> Result<()> {
...
 }
+
+pub fn run_dprint_self_update(ctx: &ExecutionContext) -> Result<()> {
+    let dprint = require("dprint")?;
+
+    print_separator("dprint self update");
+
+    if ctx.config().dprint_self_update() {
+        ctx.execute(&dprint).arg("upgrade").status_checked()?;
+    }
```
```suggestion
    if !ctx.config().dprint_self_update() {
        return Err(SkipStep("dprint self update is disabled in config"));
    }

    print_separator("dprint self update");
    ctx.execute(&dprint).arg("upgrade").status_checked()?;
```
To avoid printing the separator and then doing nothing, and to return the correct skip message.

src/steps/generic.rs
```diff
@@ -2457,3 +2457,48 @@ fn refresh_mise_env(ctx: &ExecutionContext, mise: &Path) -> Result<()> {
...
+
+    Ok(())
+}
+
+pub fn run_dprint_plugins(ctx: &ExecutionContext) -> Result<()> {
+    let dprint = require("dprint")?;
+
+    print_separator("dprint plugins");
```
GideonBear: Move this to before the `write_all` calls, again to avoid printing the separator when we actually skip

src/steps/generic.rs
```diff
@@ -2457,3 +2457,48 @@ fn refresh_mise_env(ctx: &ExecutionContext, mise: &Path) -> Result<()> {
...
+            .code()
+            .ok_or_eyre("Couldn't get status code (terminated by signal)")?;
+        if status_code == 11 {
+            debug!("dprint global config update not available")
+        } else {
+            std::io::stdout().lock().write_all(&output.stdout)?;
+            std::io::stderr().lock().write_all(&output.stderr)?;
+            if status_code != 0 {
+                return Err(StepFailed.into());
+            }
```
GideonBear: There should be a way to not fail with only a single status code, I think it's called `status_checked_with`

src/steps/generic.rs
```diff
@@ -2457,3 +2457,48 @@ fn refresh_mise_env(ctx: &ExecutionContext, mise: &Path) -> Result<()> {
...
+    }
+
+    let output = cmd.output()?;
+    if let ExecutorOutput::Wet(output) = output {
+        let status_code = output
+            .status
+            .code()
+            .ok_or_eyre("Couldn't get status code (terminated by signal)")?;
+        if status_code == 11 {
+            debug!("dprint global config update not available")
```
```suggestion
            return Err(SkipStep("dprint global config update not available"));
```

GideonBear (changes requested):

src/step.rs
```diff
@@ -333,6 +335,8 @@ impl Step {
                 runner.execute(*self, "dkp-pacman", || linux::run_dkp_pacman_update(ctx))?
             }
             Dotnet => runner.execute(*self, ".NET", || generic::run_dotnet_upgrade(ctx))?,
+            DprintPlugins => runner.execute(*self, "dprint plugins", || generic::run_dprint_plugins(ctx))?,
+            DprintSelfUpdate => runner.execute(*self, "dprint self update", || generic::run_dprint_self_update(ctx))?,
```
uwuclxdy: Sorry for bringing this up again, but could we have `Dprint` (the self update) and `DprintPlugins` instead? only in enum, the step name and separator should keep `"dprint self update"`. This is currently standard for other steps that do self updates as well

uwuclxdy: or actually "dprint" as step name works too; your call.

src/steps/generic.rs
scop: Hm, a two edged sword. Status code 11 could indicate that there is no global config, or that the global config has some errors.

For the first case it's ok and good to skip the separator, but I suppose for the second one I suppose we should let the error output through, and in that case outputting the separator is actually desirable. 

Problem is differentiating between the two properly. Scraping error output is not nice, and replicating the [global config location logic](https://dprint.dev/global-config/#initializing-a-global-configuration) doesn't sound too good either.

But if we keep the status 11 handling as it is now, and don't try improving on the different cases behind it, moving as suggested would be ok for now I think.

src/steps/generic.rs
scop: Hm, right. Then again, after the split, this step has no other uses besides the self update, so does it make sense to have the self update boolean option in the first place? Maybe it'd be better if we remove the config option and instead just don't add this step to default_steps?

src/steps/generic.rs
scop: There's at least `status_checked_with_codes` but I'm not sure how it would be used if we want to output some additional diagnostics for status 11.

We could obviously just treat 11 as success and let dprint output what it wants and remove all of the custom exit status code handling and conditional output suppression, but it would look a bit weird if let's say the config has errors:
```
── 17:10:49 - dprint plugins ───────────────────────────────────────────────────
Error deserializing. Unexpected token on line 6 column 4
    at /home/scop/.config/dprint/dprint.jsonc

── 17:10:49 - Summary ──────────────────────────────────────────────────────────
dprint plugins: OK
```
...or if it's missing (this could maybe be acceptable):
```
── 17:13:21 - dprint plugins ───────────────────────────────────────────────────
Could not find global dprint.json file. Create one with `dprint init --global`

── 17:13:21 - Summary ──────────────────────────────────────────────────────────
dprint plugins: OK
```

Treating 11 as an error is IMO not good because it triggers the failure message and retry prompt, also in the common case where the global config is missing altogether.

src/step.rs
scop: Done in 6644f823dc84c06d01f9af867ad28854eb67b1e3.

src/steps/generic.rs
scop: Removed the config option in a1c03b699b16247808e204d31066d7d1594b337c, simply disabling this step now it's a separate one works for me.

src/steps/generic.rs
scop: Applied + followup in 30c31d43947db04c55217a84c10f973f37370ca8

src/steps/generic.rs
scop: Moved as suggested in 4ca4a0ffc164cd294a0568a1f6899307ca149aa0.

The downside is that the separator is now printed only after the command is run, which could take a bit.

scop: How do we feel about the mentioned downside? I think it feels a bit odd to see the output seemingly stall at the end of the preceding step for some time. The plugin update rarely takes _that_ long, but it's surely easily noticeable when there is an eligible config.

For example mine (which may not be that typical case) has 13 plugins, and an update run even when there are no updates takes about 4,5 seconds.

scop:
Resolved conflicts.

@GideonBear sorry about the ping, but I think I've responded to the requests above, left the threads open for easy location & in case there's further need for discussion. But I'm basically waiting for feedback and don't have anything actively to work on here.

scop:
@GideonBear, @uwuclxdy sorry about the ping again, but just in case it was missed above: I think I've responded to the requests above, left the threads open for easy location & in case there's further need for discussion. But I'm basically waiting for feedback and don't have anything actively to work on here, would like to see this move onwards.

GideonBear:
Hi, I haven't missed it; you're still in my inbox. I will review this when I have the time; I'm pretty busy in this period, and there are issues/PRs with higher priority.

src/steps/generic.rs
```diff
@@ -2500,3 +2500,46 @@ fn refresh_mise_env(ctx: &ExecutionContext, mise: &Path, neutral_cwd: &Path) ->
     debug!("Refreshed process environment from mise");
     Ok(())
 }
+
+pub fn run_dprint_self_update(ctx: &ExecutionContext) -> Result<()> {
+    let dprint = require("dprint")?;
+
+    print_separator("dprint");
+
+    ctx.execute(&dprint).arg("upgrade").status_checked()?;
```
uwuclxdy, line 2509: ref #2129 where you said it yourself that self-update should not run when it's managed by mise. can this be behind `is_descendant_of(&HOME_DIR)` or something (see `run_juliaup` for reference)

src/steps/generic.rs
```diff
@@ -2500,3 +2500,46 @@ fn refresh_mise_env(ctx: &ExecutionContext, mise: &Path, neutral_cwd: &Path) ->
...
+    }
+
+    let output = cmd.output()?;
+    if let ExecutorOutput::Wet(output) = output {
+        let status_code = output
+            .status
+            .code()
+            .ok_or_eyre("Couldn't get status code (terminated by signal)")?;
+        if status_code == 11 {
+            return Err(SkipStep(t!("dprint global config update not available").to_string()).into());
```
uwuclxdy, line 2530: print stderr before skipping

src/steps/generic.rs
```diff
@@ -2500,3 +2500,46 @@ fn refresh_mise_env(ctx: &ExecutionContext, mise: &Path, neutral_cwd: &Path) ->
...
+    Ok(())
+}
+
+pub fn run_dprint_plugins(ctx: &ExecutionContext) -> Result<()> {
+    let dprint = require("dprint")?;
+
+    let mut cmd = ctx.execute(&dprint);
+    cmd.arg("config");
+    cmd.arg("update");
+    cmd.arg("--global");
```
```suggestion
    let mut cmd = ctx.execute(&dprint).args(["config", "update", "--global"]);
```

src/steps/generic.rs
```diff
@@ -2500,3 +2500,46 @@ fn refresh_mise_env(ctx: &ExecutionContext, mise: &Path, neutral_cwd: &Path) ->
...
+
+    if ctx.config().yes(Step::DprintPlugins) {
+        cmd.arg("--yes");
+    }
+
+    let output = cmd.output()?;
+    if let ExecutorOutput::Wet(output) = output {
```
uwuclxdy, line 2524: sorry i must have missed something, but could you still explain why are we checking this? as far as i know, commands that are not `.always()` are skipped automatically in dry run and this makes the step not run in damp mode (which is just normal/wet run but it prints the commands before executing them)

uwuclxdy (changes requested):
looks good but I don't use dprint and can't test it myself, I just know that it exits 11 in some other scenarios like unreachable remotes. I'm not sure if we should actually handle it differently than "no global config" error which also exits 11.

src/steps/generic.rs
scop: In case of no config, stderr would be 
```
Could not find global dprint.json file. Create one with `dprint init --global`
```
...which I feel is unnecessary noise in the presumably most common case where the global plugins update is skipped due to no global config found. The advise to create one is also arguably superfluous -- if there was no global config because the user had no use for one and that's why global plugins were not updated, it doesn't IMO make much sense to create the config just so topgrade can upgrade them (it would not be a skip if the global config exists).

src/steps/generic.rs
scop: There are many things which should not self upgrade if they're managed by mise or some other installer/updater. I'd rather not write a dprint specific implementation. Instead, I'd like to see a more general reusable approach provided by topgrade, akin to what I noted above in https://github.com/topgrade-rs/topgrade/pull/2131#discussion_r3522900190

Being located in home dir somewhere could be one of the conditions (but it's too weak alone), along with detecting install paths of mise, aqua, whatnot.

src/steps/generic.rs
scop: It's just my bad rust-fu, this way we can get hold of output.status.code(). Improvements welcome.

Issue #624, opened 2023-12-03
Support `topgrade.toml` inside `topgrade.d/`


etiennepellegrini:
I think `topgrade` should check `$XDG_CONFIG_DIR/topgrade.d/topgrade.toml` on top of `$XDG_CONFIG_DIR/topgrade.toml`, and maybe _instead_ of `$XDG_CONFIG_DIR/topgrade/topgrade.toml`.
I don't see the point in having either two topgrade dirs, or a dir and a file in my `$XDG_CONFIG_DIR`

I do not fully understand the use case for `topgrade.d/`, but from the comment in the example config file, it's supposed to contain different config files which are superseded by `topgrade.toml`. I therefore imagine they typically are named differently, and it's not a problem to move `topgrade.toml` to `topgrade.d/topgrade.toml`?

Thanks!

niStee:
@etiennepellegrini Stackexchange provide you with this [answer](https://unix.stackexchange.com/questions/4029/what-does-the-d-stand-for-in-directory-names).

And at least on my amd64 win11, topgrade uses topgrade.toml either in $XDG_CONFIG_DIR/topgrade.d/ $XDG_CONFIG_DIR/topgrade/ or $XDG_CONFIG_DIR/topgrade/.

etiennepellegrini:
Yes, I believe the default should be created in `toprade.d`.
Moreover, the README file should be updated to mention `topgrade.d`, and that the main config file is searched for at `topgrade.d/topgrade.toml` as well.

In addition, `tograde.d/topgrade.toml` should maybe be added explicitly to [config.rs](https://github.com/topgrade-rs/topgrade/blob/788e0412f6ad3e78e040de1cffd84af845550e9e/src/config.rs#L472-L474), as are `tograde/topgrade.toml` and `topgrade.toml`

SteveLauC:
This directory was added before I actually maintained this project, my personal view on it is that we don't need it at all cause Topgrade does not have that many configuration entries and a simple `topgrade.toml` file would be totally sufficient.

etiennepellegrini:
That would seem like the best option to me too 👍

SteveLauC:
Well, that's just my personal point, the previous maintainer thought that this feature should be kept, so I cannot remove it. I would recommend that simply don't use it...

bb010g:
As a side note, as long as we have `topgrade.d`, its semantics should be better defined.

[...]

PR #2132, opened 2026-07-04
fix(config): stop creating `topgrade.d/`

uwuclxdy:
## What does this PR do

Topgrade now doesn't automatically create `topgrade.d/` folder anymore. It only checks if it exists and merges those configs like before. Also added a line to the readme about this that was missing before.

ref #624

## Standards checklist

### AI involvement

Only for review and unit test generation.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -772,9 +772,8 @@ impl ConfigFile {
             return Ok(res);
         }
 
-        let entries = fs::read_dir(&dir_to_search).inspect_err(|_| {
-            error!("Unable to read {}", dir_to_search.display());
-        })?;
+        let entries =
+            fs::read_dir(&dir_to_search).wrap_err_with(|| format!("Unable to read {}", dir_to_search.display()))?;
 
         for entry in entries {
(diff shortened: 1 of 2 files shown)


src/config.rs
```diff
@@ -763,16 +763,20 @@ impl ConfigFile {
...
-            debug!("No additional configuration directory exists, creating one");
-            fs::create_dir_all(&dir_to_search)?;
+            debug!("No additional configuration directory exists, skipping");
+            return Ok(res);
         }
 
-        for entry in fs::read_dir(&dir_to_search)? {
+        let entries = fs::read_dir(&dir_to_search).inspect_err(|_| {
+            error!("Unable to read {}", dir_to_search.display());
+        })?;
```
GideonBear: Why not just `.wrap_err`?

GideonBear (changes requested):

src/config.rs
uwuclxdy: no good reason, fixed now

GideonBear (approved):

PR #2134, opened 2026-07-05
feat(step): add sera (KDE Plasma addon updater)

jtbrough:
## What does this PR do
Adds support for `sera`, a CLI updater for KDE Plasma addons (widgets, plugins, plasmoids).

Ref: https://gitlab.com/cscs/sera

## Standards checklist

Dry run:
```
── 11:11:55 - sera ─────────────────────────────────────────────────────────────
Dry running: /home/jordan/.local/bin/sera upgrade

── 11:11:55 - Summary ──────────────────────────────────────────────────────────
sera: OK
```

Full run:
```
── 11:11:59 - sera ─────────────────────────────────────────────────────────────

✔ Fetching data from the API
✔ Checking widgets for updates
✔ Nothing to do


── 11:12:03 - Summary ──────────────────────────────────────────────────────────
sera: OK
```

### AI involvement
none

## For new steps

  the underlying command

### DIFF
--- a/src/step.rs
+++ b/src/step.rs
@@ -156,6 +156,7 @@ pub enum Step {
     Scoop,
     Sdkman,
     SelfUpdate,
+    Sera,
     Sheldon,
     Shell,
     Skills,
@@ -604,6 +605,11 @@ impl Step {
                     }
                 }
             }
+            Sera =>
+            {
+                #[cfg(all(unix, not(any(target_os = "macos", target_os = "android"))))]
+                runner.execute(*self, "sera", || unix::run_sera(ctx))?
+            }
             Sheldon => runner.execute(*self, "sheldon", || generic::run_sheldon(ctx))?,
             Shell => {
                 #[cfg(unix)]
@@ -806,6 +812,7 @@ pub(crate) fn default_steps() -> Vec<Step> {
         Waydroid,
         AutoCpufreq,
         CinnamonSpices,
+        Sera,
         Mandb,
         Pkgfile,
         Yadm,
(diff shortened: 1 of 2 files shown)


uwuclxdy (approved):
Code looks good, I'm only thinking about a possible conflict with the planned built in plasmoid updater. I say let's merge this for now and we'll see when plasmoid-updater gets good enough to adopt it into topgrade? @GideonBear

jtbrough:
Oh, neat, I was not aware of uwuclxdy/plasmoid-updater until now.

Is that planned to be merged into mainline KDE? I suppose one way to look at it is if someone goes to the extent of installing `sera`, that's a fairly strong indicator they intend to use it. If a more integral updater is available, they could either remove `sera` or disable it in `topgrade`. Or worst-case, just get updates checked twice. First one wins!

There are a lot of ways to do a lot of things (sometimes redundantly) in Plasma-land, so this won't be a terrible surprise :).

Happy to support either direction though.

uwuclxdy:
> Is that planned to be merged into mainline KDE?

Uhh no, but it is planned to be merged into topgrade, the same way jetbrains updater is (compiled in as library).

> I suppose one way to look at it is if someone goes to the extent of installing sera, that's a fairly strong indicator they intend to use it. If a more integral updater is available, they could either remove sera or disable it in topgrade.

Yes, actually this gives me an idea; let's add sera as normal step for now, then / when plasmoid-updater gets integrated it takes precedence over sera unless disabled? lmk, just an idea and I'm also not even close to finished with it (I'm rewriting it by hand without AI assistance / vibecoding, although it's perfectly functional as it is right now lol which is also a milestone for v1 and getting it into topgrade).

> There are a lot of ways to do a lot of things (sometimes redundantly) in Plasma-land, so this won't be a terrible surprise :).

Yeah definitely xdd

GideonBear:
> let's add sera as normal step for now, then / when plasmoid-updater gets integrated it takes precedence over sera unless disabled?

Good idea I think!
For now sera definitely deserves a step.

GideonBear (approved):
Thanks!

PR #2136, opened 2026-07-06
fix(antidote): update from interactive zsh when available

### LINKED ISSUE #987: Antidote not being catched when installed via homebrew
## Erroneous Behavior
When `antidote` is installed via a package manager like `brew`, topgrade cannot find it because it has to be sourced into the terminal session first.

## Expected Behavior
Topgrade should still find topgrade 

## Steps to reproduce
1. Install topgrade via `brew`
2. Run `topgrade`

## Possible Cause (Optional)
topgrade cannot find it because it has to be sourced into the terminal session first, for some reason this is not working but I can still access `antidote` directly from the terminal without any issues.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

As alternative workaround I am currently using the following custom command.
```toml
[commands]
"Antidote updates" = "source $(brew --prefix antidote)/share/antidote/antidote.zsh && antidote update"
```

## Additional Details
- Operation System/Version
masOS 15.1.1

- Installation
homebrew

16.0.1

not relevant

gormanity:
## What does this PR do

Fixes #987.

This updates the Antidote step to use interactive zsh when Antidote is already available there, while preserving the existing `$ZDOTDIR/.antidote` sourcing behavior as a fallback.

## Standards checklist

Tested with Antidote installed via Homebrew and loaded from interactive zsh.

Dry run:

```
―― 14:42:57 - antidote ――
Dry running: /bin/zsh -i -c 'antidote update'
―― 14:42:57 - Summary ――
antidote: OK
```

Full wet test:

```
―― 15:15:15 - antidote ――
Updating bundles...
Bundle updates complete.

Self updating is disabled in this build.
Use your OS package manager to update antidote itself.
...
antidote: OK
```
And Rust checks with test ouput:

```
cargo fmt --check
cargo check
cargo clippy
cargo test
  
test result: ok. 24 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

- [ ] If this PR introduces new user-facing messages they are translated (N/A: no user-facing messages added or changed)

### AI involvement

I used Codex to help investigate the root cause, prepare the patch branch, and run validation. I reviewed the implementation, asked for revisions, and am writing/submitting this PR myself.

### DIFF
--- a/src/steps/zsh.rs
+++ b/src/steps/zsh.rs
@@ -1,5 +1,5 @@
 use std::env;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 use color_eyre::eyre::Result;
 use tracing::debug;
@@ -33,17 +33,39 @@ pub fn zshrc() -> PathBuf {
     zdotdir().join(".zshrc")
 }
 
+enum AntidoteUpdate {
+    Shell,
+    Source(PathBuf),
+}
+
 pub fn run_antidote(ctx: &ExecutionContext) -> Result<()> {
     let zsh = require("zsh")?;
-    let mut antidote = zdotdir().join(".antidote").require()?;
-    antidote.push("antidote.zsh");
+    let update = if antidote_available_in_shell(ctx, &zsh) {
+        AntidoteUpdate::Shell
+    } else {
+        let mut antidote = zdotdir().join(".antidote").require()?;
+        antidote.push("antidote.zsh");
+        AntidoteUpdate::Source(antidote)
+    };
 
     print_separator("antidote");
 
+    match update {
+        AntidoteUpdate::Shell => ctx.execute(zsh).args(["-i", "-c", "antidote update"]).status_checked(),
+        AntidoteUpdate::Source(antidote) => ctx
+            .execute(zsh)
+            .arg("-c")
+            .arg(format!("source {} && antidote update", antidote.display()))
+            .status_checked(),
+    }
+}
+
+fn antidote_available_in_shell(ctx: &ExecutionContext, zsh: &Path) -> bool {
     ctx.execute(zsh)
-        .arg("-c")
-        .arg(format!("source {} && antidote update", antidote.display()))
-        .status_checked()
+        .always()
+        .args(["-i", "-c", "antidote home"])
+        .output_checked()
+        .is_ok()
 }
 
 pub fn run_antibody(ctx: &ExecutionContext) -> Result<()> {


GideonBear (approved):
Thanks!

src/steps/zsh.rs
```diff
@@ -33,17 +33,39 @@ pub fn zshrc() -> PathBuf {
...
+        AntidoteUpdate::Shell => ctx.execute(zsh).args(["-i", "-c", "antidote update"]).status_checked(),
+        AntidoteUpdate::Source(antidote) => ctx
+            .execute(zsh)
+            .arg("-c")
+            .arg(format!("source {} && antidote update", antidote.display()))
+            .status_checked(),
+    }
+}
+
+fn antidote_available_in_shell(ctx: &ExecutionContext, zsh: &PathBuf) -> bool {
```
uwuclxdy:
```suggestion
fn antidote_available_in_shell(ctx: &ExecutionContext, zsh: &Path) -> bool {
```

maybe use `&Path` instead if possible

src/steps/zsh.rs
```diff
@@ -33,17 +33,39 @@ pub fn zshrc() -> PathBuf {
...
     ctx.execute(zsh)
-        .arg("-c")
-        .arg(format!("source {} && antidote update", antidote.display()))
-        .status_checked()
+        .always()
+        .args(["-i", "-c", "antidote home >/dev/null"])
```
uwuclxdy:
```suggestion
        .args(["-i", "-c", "antidote home"])
```
redundant since `.output_checked()`

src/steps/zsh.rs
```diff
@@ -33,17 +33,39 @@ pub fn zshrc() -> PathBuf {
...
+    };
 
     print_separator("antidote");
 
+    match update {
+        AntidoteUpdate::Shell => ctx.execute(zsh).args(["-i", "-c", "antidote update"]).status_checked(),
+        AntidoteUpdate::Source(antidote) => ctx
+            .execute(zsh)
+            .arg("-c")
+            .arg(format!("source {} && antidote update", antidote.display()))
```
uwuclxdy, line 58: note for @GideonBear : pre-existing and not an issue because this just means code execution via user's own `.zshrc` + it's done the same way in other zsh steps

uwuclxdy:
spotted by @GideonBear: unquoted `source {}` breaks if paths have spaces. this is true for other zsh steps and also sdkman step (`os/unix.rs`).
We should probably track this in a separate issue/PR.

gormanity:
I saw the same `source {}` quoting concern while working through this. I agree it is real, but because the same pattern appears in other zsh-related steps as well, I think it is better scoped to a separate follow-up issue/PR rather than broadening this Antidote detection fix.

GideonBear (approved):

PR #2141, opened 2026-07-07
fix(shell): quote sourced command paths

gormanity:
## What does this PR do

Addresses https://github.com/topgrade-rs/topgrade/pull/2136#issuecomment-4893272990.

Adds `shell_source_command` util that constructs shell commands with quoted source paths and implements it in command constructors in `src/steps/os/unix.rs` and `src/steps/zsh.rs`. Also adds a unit test for `shell_source_command`.

I kept this as a small util rather than introducing a more abstracted command builder because we only need to safely interpolate the sourced file path given that the commands are static. I also used a helper instead of in-lining the quote call at each site because the same `source <path> && ...` pattern appears in multiple steps, so centralizing it keeps the call sites readable, avoids repeating the quoting logic, and gives the path handling one focused test.

## Standards checklist

Ran the following:

```
cargo fmt --check
cargo check
cargo clippy
cargo test
```

The test suite includes the added unit coverage for source paths with spaces and single quotes.

I also ran an end-to-end smoke test through Topgrade:

- Built local ./target/debug/topgrade
- Created a fake zr binary in a temp PATH
- Created temp ZDOTDIR paths containing a space and a single quote
- Each .zshrc exported a marker
- Ran `./target/debug/topgrade --only shell --run-type wet`

Output:

```
--- 21:22:51 - zr ---
--- 21:22:51 - Summary ---
zr: OK
--- 21:22:51 - zr ---
--- 21:22:51 - Summary ---
zr: OK
topgrade smoke markers:
space-ok
quote-ok
```

I also verified the same smoke test fails on origin/main without this fix:

```
--- baseline case: space dir ---
--- 21:24:53 - zr ---
zsh:source:1: no such file or directory: /tmp/topgrade
--- 21:24:53 - Summary ---
zr: FAILED

--- baseline case: quote's dir ---
--- 21:24:53 - zr ---
zsh:1: unmatched '
--- 21:24:53 - Summary ---
zr: FAILED
```

- [ ] If this PR introduces new user-facing messages they are translated (N/A: no new or changed user-facing messages)

### AI involvement

I prompted Codex to find all instances of unquoted source paths, replace them with the new util, and construct and run the smoke test detailed above. I reviewed the implementation and am submitting the PR myself.

### DIFF
--- a/src/steps/os/unix.rs
+++ b/src/steps/os/unix.rs
@@ -32,7 +32,7 @@ use crate::execution_context::ExecutionContext;
 use crate::executor::Executor;
 use crate::step::Step;
 use crate::terminal::print_separator;
-use crate::utils::{PathExt, require};
+use crate::utils::{PathExt, require, shell_source_command};
 
 fn get_sudo_uid<P: AsRef<Path>>(path: P) -> Option<u32> {
     if let Ok(metadata) = fs::metadata(&path) {
@@ -1059,8 +1059,7 @@ pub fn run_sdkman(ctx: &ExecutionContext) -> Result<()> {
         .map_or_else(|_| HOME_DIR.join(".sdkman"), PathBuf::from)
         .join("bin")
         .join("sdkman-init.sh")
-        .require()
-        .map(|p| format!("{}", &p.display()))?;
+        .require()?;
 
     print_separator("SDKMAN!");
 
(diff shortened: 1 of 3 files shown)


uwuclxdy:
Unsure if we want to do it like this, need @GideonBear's thoughts about it and if I should open a PR from https://github.com/topgrade-rs/topgrade/commit/f77ae67afe8c56766e2eeb3d157ee1fbe972b3a5 instead (positional `$1` variant of this without a helper).
Also, this PR is behind #2136, therefore antidote step isn't fixed here.

gormanity:
Rebased now that #2136 has landed, so the Antidote fallback path is covered as well.

I’m happy to switch to the positional `$1` shape if that is preferred. I agree that direction makes sense if we want to refactor these shell invocations to avoid interpolating paths into command strings at all. This version was intentionally narrower in scope to preserve the existing command shape.

GideonBear:
The zsh ones were already fixed in another PR.
The sdkman ones can be fixed easier by running `bash -i` and dropping the `source` call. The sdkman init file is already guaranteed to be sourced by `.bashrc`.

Could you do that?

PR #2144, opened 2026-07-07
feat(step): add OpenCode upgrade

caiopiccirillo:
## What does this PR do

Add OpenCode upgrade step

## Standards checklist

### AI involvement

No AI involvement

## For new steps

  the underlying command

### DIFF
--- a/src/steps/generic.rs
+++ b/src/steps/generic.rs
@@ -2269,7 +2269,6 @@ pub fn run_skills(ctx: &ExecutionContext) -> Result<()> {
 pub fn run_opencode(ctx: &ExecutionContext) -> Result<()> {
     let opencode = require("opencode")?;
 
-    print_separator("OpenCode");
     let script_install_path = HOME_DIR.join(".opencode").join("bin");
     if !opencode
         .canonicalize()
@@ -2277,6 +2276,7 @@ pub fn run_opencode(ctx: &ExecutionContext) -> Result<()> {
     {
         return Err(SkipStep(t!("OpenCode not installed with the official script").to_string()).into());
     }
+    print_separator("OpenCode");
     ctx.execute(opencode).arg("upgrade").status_checked()
 }
 
--- a/src/steps/generic.rs
+++ b/src/steps/generic.rs
@@ -2266,6 +2266,14 @@ pub fn run_skills(ctx: &ExecutionContext) -> Result<()> {
     command.status_checked()
 }
 
+pub fn run_opencode(ctx: &ExecutionContext) -> Result<()> {
+    let opencode = require("opencode")?;
+
+    print_separator("OpenCode");
+
+    ctx.execute(opencode).arg("upgrade").status_checked()
+}
+
 fn ollama_serve(ctx: &ExecutionContext, ollama: &Path) -> Result<ExecutorChild> {
     ctx.execute(ollama)
         .arg("serve")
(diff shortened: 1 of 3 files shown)


uwuclxdy:
opencode can be installed with npm, bun, brew, paru and with the official download script. the first 4 of these topgrade already supports. 
i've got mine installed through bun and `opencode upgrade` fails because of min release time set to 2 days. there are more possibilities of conflicts like this; so is there a way to detect and run this step only if opencode is installed with the binary installation script? i'm thinking maybe the script puts the binary into a specific location which could be used to identify installation type?
it'd be really annoying to see a step fail every time / having to manually disable it just because opencode isn't installed with the script. 
thanks and lmk

caiopiccirillo:
Thanks for this info! Never crossed my mind to check the other installers.
I just checked their binary install script and it's always in $HOME/.opencode/bin/
Ref.: https://raw.githubusercontent.com/anomalyco/opencode/refs/heads/dev/install 
I'm not sure if it's the same path for other installers, but I can investigate.

uwuclxdy:
> I'm not sure if it's the same path for other installers, but I can investigate.

yes, please do and report back if that path is unique to the script installer (so we can use it as a signal to start this step) or should we do it in another way.

caiopiccirillo:
Sorry for the delay. The path is unique to the script, but there's a pririty of path installation in the script (https://github.com/anomalyco/opencode#installation-directory). And the user can also set the a custom path to install. 
I was thinking on checking if opencode was installed by npm, bun, brew or paru. If not, then we'll perform the `opencode upgrade` command. 
What do you think?

uwuclxdy:
Okay I checked the install script, lets run `opencode upgrade` only when the binary is in `~/.opencode/bin/`.

> And the user can also set the a custom path to install.

Yes and no. Technically a user *can* set a custom directory, but that's only possible if they modify the script. Line 68 of the script: `INSTALL_DIR=$HOME/.opencode/bin`. We should expect the binary to be there; if the user decided to move it somewhere else, that is not the official installation anymore, therefore we should not go look where it is for them.

BTW, take a look at how juliaup step does this in `generic.rs:369`.

src/steps/generic.rs
```diff
@@ -2246,6 +2246,14 @@ pub fn run_skills(ctx: &ExecutionContext) -> Result<()> {
     command.status_checked()
 }
 
+pub fn run_opencode(ctx: &ExecutionContext) -> Result<()> {
+    let opencode = require("opencode")?;
+
```
uwuclxdy, line 2271:
```suggestion

    let script_dir = HOME_DIR.join(".opencode").join("bin");
    if !opencode.canonicalize().is_ok_and(|p| p.is_descendant_of(&script_dir)) {
        return Err(SkipStep(t!("OpenCode not installed with the official script")).into());
    }

```

something like this ig

uwuclxdy (changes requested):

caiopiccirillo:
thanks for the guidance, I pushed the changes as you pointed out. :smile:

uwuclxdy (approved):
thanks, lgtm!

src/steps/generic.rs
```diff
@@ -2266,6 +2266,20 @@ pub fn run_skills(ctx: &ExecutionContext) -> Result<()> {
...
+    print_separator("OpenCode");
```
GideonBear: Move separator printing below the check that might skip the step, to avoid printing the separator when the step is actually skipped

GideonBear (changes requested):

src/steps/generic.rs
caiopiccirillo: done

GideonBear (approved):
Thanks both!

Issue #491, opened 2023-07-11
Do not fail parsing config on invalid keys or steps


xeruf:
## Erroneous Behavior
Topgrade ignores the config when it detects an unknown step or config key.

## Expected Behavior
It warns but uses the remainder of the config.

## Steps to reproduce
Run an older version of topgrade with a new config file (e.g. when using a config file on Debian and Arch through a dotfiles repo)

See https://github.com/r-darwish/topgrade/issues/847

SteveLauC:
This is not something hard to implement, removing `#[serde(deny_unknown_fields)]` from types in `src/config.rs` would get it done.

But this will make users hard to find mistakes in their configuration file...

xeruf:
Just spit out a warning but use the config, as I outlined as expectation above.
How would that hinder users from finding mistakes in their config?

Once again today topgrade ignores all my config and performs wrong steps because some parameter change in an update...

SteveLauC:
> Once again today topgrade ignores all my config and performs wrong steps because some parameter change in an update...

Hi, I am still not convinced this is a good feature to have, what about updating your configuration to the latest one?

And, modifications to the configuration will be seen as breaking changes, after Topgrade 14.0.0, they will be printed on your first run:

```
$ topgrade --version
Topgrade 14.0.0

$ topgrade

── 11:16:41 - Topgrade 14.0.0 Breaking Changes ─────────────────────────────────
1. In 13.0.0, we introduced a new feature, pushing git repos, now this feature
   has been removed as some users are not satisfied with it.

   For configuration entries, the following ones are gone:

  
   [git]
   pull_only_repos = []
   push_only_repos = []
   pull_arguments = ""
   push_arguments = ""
  


Confirmed? (y)es/(N)o
```

so it won't be hard for users to find and correct them:>

xeruf:
I ust topgrade across diferent operating systems, such as arch and debian, which might be published in different versions

I might add a config flag for the newer version on arch, sync my dotfiles to a debian machine, and then topgrade won't work until debian has the newest package

so it's not about an incorrect flag, just a not-yet-implemented one

GideonBear:
@xeruf Good point. Would failing be a better solution than carrying on with default config?

GideonBear:
Re-opened this for more discussion, this does not mean it is accepted, or will not be closed again in the future

PR #2148, opened 2026-07-08
refactor(config): merge `topgrade.d` and `[include]` parse paths

uwuclxdy:
## What does this PR do

This PR merges the `[include]` and `topgrade.d` config parse paths into a helper function. Merge behaviour is preserved, except that a bad file in `topgrade.d/` now reports the error as well. ref #624.

Any bad file in `topgrade.d/` now also causes Topgrade to crash instead of silently ignoring.

For v18 we might want to change how `topgrade.d` works so it becoms a drop-in dir (sorted, last-wins, supplementing `topgrade.toml` instead of overriding it); the same way systemd, `cron.d`, and other `conf.d` work. It's still "first-wins" and also wins over the main config, which is the opposite of how other `.d` do it.

## Standards checklist

### AI involvement

I first wrote a worse implementation myself then had an LLM correct it before commiting and opening this PR.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -795,6 +795,14 @@ impl ConfigFile {
         Ok(res)
     }
 
+    fn merge_config_file(&mut self, path: &Path) -> Result<()> {
+        let contents = fs::read_to_string(path).wrap_err_with(|| format!("Unable to read {}", path.display()))?;
+        let parsed: Self =
+            toml::from_str(&contents).wrap_err_with(|| format!("Failed to deserialize {}", path.display()))?;
+        self.merge(parsed);
+        Ok(())
+    }
+
     /// Read the configuration file.
     ///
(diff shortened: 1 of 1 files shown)


GideonBear:
Half-related, we can also just always crash on invalid config files. I'm not sure why we ever ignored them.
Ref #491

uwuclxdy:
> I'm not sure why we ever ignored them.

https://github.com/uwuclxdy/topgrade/commit/9a3b7e30eef92f5a4e876192ac9cfcaa92bab957

lmk if I should make it crash on any invalid config file

GideonBear:
> lmk if I should make it crash on any invalid config file

I think we can just do that

uwuclxdy:
#491 wants a warn for unknown keys instead of crash like we do it now (this PR). Should we crash for every config error or let unknown keys be warn-only? 

PR is good for review otherwise.

PR #2153, opened 2026-07-09
feat(debian): prefer apt over apt-get

scop:
## What does this PR do

Prefers apt over apt-get. apt has nice colors, progress bar etc.

In practice, it is likely that after this, setups invoking apt-get are very rare. But I suppose they can exist.

## Standards checklist

### AI involvement

None.

## For new steps

  the underlying command

### DIFF
--- a/src/steps/os/linux.rs
+++ b/src/steps/os/linux.rs
@@ -529,6 +529,7 @@ enum AptKind {
     AptFast,
     Mist,
     Nala,
+    Apt,
     AptGet,
 }
 
@@ -541,6 +542,8 @@ fn detect_apt() -> Result<(AptKind, PathBuf)> {
         Ok((Mist, mist))
     } else if Path::new("/usr/bin/nala").exists() {
         Ok((Nala, Path::new("/usr/bin/nala").to_path_buf()))
+    } else if let Some(apt) = which("apt") {
+        Ok((Apt, apt))
     } else {
         Ok((AptGet, require("apt-get")?))
     }


src/steps/os/linux.rs
```diff
@@ -541,6 +542,8 @@ fn detect_apt() -> Result<(AptKind, PathBuf)> {
         Ok((Mist, mist))
     } else if Path::new("/usr/bin/nala").exists() {
         Ok((Nala, Path::new("/usr/bin/nala").to_path_buf()))
+    } else if let Some(apt) = which("apt") {
+        Ok((Apt, apt))
     } else {
         Ok((AptGet, require("apt-get")?))
```
scop, line 548: This means users' sudo/doas/etc configs will likely need adjustments from apt-get to apt. Kind of a breaking change, at least deserves a release note.

GideonBear: Good point, but I've never considered a change in the commands we run as a breaking change. Doesn't need any special treatment IMO.

GideonBear (approved):
Thanks!

matttbe:
Hello,

Just in case, 'apt' devs wouldn't recommend this modification:

> The [apt(8)](https://manpages.debian.org/testing/apt/apt.8.en.html) commandline is designed as an end-user tool and it may change behavior between versions. While it tries not to break backward compatibility this is not guaranteed either if a change seems beneficial for interactive use.
>
> All features of [apt(8)](https://manpages.debian.org/testing/apt/apt.8.en.html) are available in dedicated APT tools like [apt-get(8)](https://manpages.debian.org/testing/apt/apt-get.8.en.html) and [apt-cache(8)](https://manpages.debian.org/testing/apt/apt-cache.8.en.html) as well. [apt(8)](https://manpages.debian.org/testing/apt/apt.8.en.html) just changes the default value of some options (see [apt.conf(5)](https://manpages.debian.org/testing/apt/apt.conf.5.en.html) and specifically the Binary scope). So you should prefer using these commands (potentially with some additional options enabled) in your scripts as they keep backward compatibility as much as possible.

Source: https://manpages.debian.org/testing/apt/apt.8.en.html#SCRIPT_USAGE_AND_DIFFERENCES_FROM_OTHER_APT_TOOLS

In some cases, this warning is displayed:

```
WARNING: apt does not have a stable CLI interface. Use with caution in scripts.
```

I guess switching to `apt` is fine, but there is a (low?) risk it breaks in the future.

scop:
FWIW my guess is that `apt update`, `apt dist-upgrade` and `apt clean`, possibly with `--yes` are very, very unlikely to break, just as unlikely as they are for `apt-get`. I also believe the mentioned warning does not trigger the way topgrade uses it, nor is the usage considered problematic in the first place.

But if topgrade want to err on the side of caution by default, a config option to use `apt` would be welcome.

GideonBear:
> Just in case, 'apt' devs wouldn't recommend this modification:

You're right, I hadn't thought about that yet, thanks for the heads-up!
I personally don't think it's much of a problem, since there is a legitimate reason to use `apt` here. If something breaks we can always fix it quickly.

AJcubes:
Hi! I was using topgrade in my cron jobs, and I have been getting this warning - 'WARNING: apt does not have a stable CLI interface. Use with caution in scripts.' Is there a way that I could somehow specify a flag or edit the config file such that I am able to make topgrade use apt-get instead of apt for myself? Thanks in advance!

GideonBear:
> Hi! I was using topgrade in my cron jobs, and I have been getting this warning - 'WARNING: apt does not have a stable CLI interface. Use with caution in scripts.' Is there a way that I could somehow specify a flag or edit the config file such that I am able to make topgrade use apt-get instead of apt for myself? Thanks in advance!

Hi, thanks for the report! This is currently unconfigurable, but if there are warnings appearing I will revert this change; that wasn't the intention.
Can you show a verbose output (`topgrade --only system --verbose`) where the warning is visible?

scop:
JFYI, that warning can be globally disabled using apt's `Apt::Cmd::Disable-Script-Warning` setting, e.g.
```
echo "Apt::Cmd::Disable-Script-Warning true;" | sudo tee /etc/apt/apt.conf.d/90disablescriptwarning
```

I think this will not happen for majority of users, whereas majority of users will benefit from the better UX of apt over apt-get, so reverting would be unfortunate.

A config option to use apt over apt-get (or more generally, a config option to the path to the binary to use as "the apt") would be much preferable to simply reverting this.

GideonBear:
@AJcubes Does this still happen? Can you show a verbose output (`topgrade --only system --verbose`) where the warning is visible?

PR #2155, opened 2026-07-10
fix(mise): avoid project cwd for upgrades

### LINKED ISSUE #2139: mise step inherits Topgrade cwd and may not upgrade system/user packages
## Checklist

I did not find a direct duplicate. Related prior issues/PRs based on this class of problem:
- #568: `pnpm` fails when Topgrade is run from a project directory
- #1592 / #1596: Aqua update behavior was changed to avoid depending on where Topgrade is run

## Erroneous Behavior

Topgrade’s mise step runs `mise upgrade` in whatever directory `topgrade` was launched from.

Because `mise upgrade` operates on the currently active mise config, running Topgrade from inside a project with a `.mise.toml` file causes the mise step to include that project’s local toolchain alongside the global/user toolchain. If the project defines the same tool as the global config, the project-local version takes precedence. This makes a system-wide updater behave differently depending on the caller’s current directory.

## Expected Behavior

Similar to #1592, Topgrade’s mise step should be cwd-stable, or at least configurable. By default, it should update the system/user-level mise toolchain from a neutral directory rather than inheriting the caller’s cwd.

There may also be a case for allowing users to configure additional project directories for local mise updates, similar to how the Git step uses explicit targets. But I think making the default system/user-level update cwd-stable should be prioritized.

## Steps to reproduce
 
1. Configure a global mise tool with a broad version:
 
```sh
 mise use --global rust@latest
```

2. Create a project-local mise config that pins the same tool to the currently installed version. Replace `1.96.0` with the version currently installed before the available update:

```sh
mkdir -p /tmp/topgrade-mise-cwd-test
cd /tmp/topgrade-mise-cwd-test
mise use --pin rust@1.96.0
```

3. Compare the global view with the project-local view:

```sh
mise outdated rust --cd /
mise outdated rust --cd /tmp/topgrade-mise-cwd-test
```

If a newer Rust version is available, the system/user-level scope reports it, while the project local scope reports no update because the local config masks the system/user-level `rust@latest`.

4. Run Topgrade’s mise step from the project directory:

```sh
cd /tmp/topgrade-mise-cwd-test
topgrade --only mise
```

Topgrade invokes `mise upgrade` from the inherited cwd, so this run uses the project-local effective mise config instead of the system/user-level config.

## Possible Cause (Optional)

Topgrade's mise step does not set a neutral working directory and constructs the `mise upgrade` command without passing mise’s `--cd` option, so the child process inherits Topgrade’s current working directory.

- [x] Yes
- [ ] No

This is expected mise behavior that needs to be worked around for a system updater.

- [ ] Yes
- [x] No

remote host?

- [ ] Yes
- [ ] No

No `[mise]` options are currently set and the disabled-step list does not affect the mise step.

## Additional Details

- Operation System/Version
  
macOS 26.5.2, Apple Silicon

- Installation

Homebrew

topgrade 17.6.2

<details>
<pre>
DEBUG Version: 17.6.2
DEBUG OS: aarch64-apple-darwin
DEBUG Step "mise"
DEBUG Detected "/opt/homebrew/bin/mise" as "mise"

--- mise ---

Topgrade then invokes:
mise plugins update
mise self-update
mise upgrade

A live test on the rust 1.96.1 update confirmed those commands inherit the directory Topgrade was launched from.
</pre>
</details>

gormanity:
## What does this PR do

Fixes #2139.

This updates the Mise step to run from a neutral temporary directory to avoid project-local configs taking precedence over system/user-level configs. A temporary directory keeps the implementation simple and avoids relying on platform-specific root directory behavior.

## Standards checklist

Ran the following tests:

```shell
cargo fmt --check
cargo check
cargo clippy -- --deny warnings
cargo test
cargo build
```

All passed. `cargo` test reported 24 passed.

I also ran an operational smoke test through the built Topgrade binary. The test put a fake mise first in PATH, launched:

```shell
topgrade --only mise --run-type wet
```

from a temp project directory containing a local .mise.toml, and logged the cwd for each fake `mise` invocation. The log showed all four `mise` commands ran from Topgrade’s temp cwd:

```shell
.../.tmpGwS3cn :: plugins update
.../.tmpGwS3cn :: self-update
.../.tmpGwS3cn :: upgrade
.../.tmpGwS3cn :: env --json
```

- [ ] If this PR introduces new user-facing messages they are translated (N/A: No new or changed user-facing messages)

### AI involvement

I used Codex to help investigate the root cause, prepare the patch branch, and run validation, including the smoke test. I reviewed the implementation, asked for revisions, and am writing/submitting this PR myself.

### DIFF
--- a/src/steps/generic.rs
+++ b/src/steps/generic.rs
@@ -2377,13 +2377,20 @@ pub fn run_ollama_pull(ctx: &ExecutionContext) -> Result<()> {
 
 pub fn run_mise(ctx: &ExecutionContext) -> Result<()> {
     let mise = require("mise")?;
+    // Run from a fresh directory so caller project-local mise.toml files do not
+    // affect the mise step.
+    let temp_dir = tempdir()?;
 
     print_separator("mise");
 
-    ctx.execute(&mise).args(["plugins", "update"]).status_checked()?;
+    ctx.execute(&mise)
+        .current_dir(temp_dir.path())
+        .args(["plugins", "update"])
+        .status_checked()?;
 
     let output = ctx
         .execute(&mise)
+        .current_dir(temp_dir.path())
         .args(["self-update"])
         .output_checked_with(|_| Ok(()))?;
     let status_code = output
@@ -2404,6 +2411,7 @@ pub fn run_mise(ctx: &ExecutionContext) -> Result<()> {
     let mut cmd = ctx.execute(&mise);
 
     cmd.arg("upgrade");
+    cmd.current_dir(temp_dir.path());
 
     if ctx.config().mise_interactive() {
         cmd.arg("--interactive");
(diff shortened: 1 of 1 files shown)


GideonBear (approved):
Thanks!

Issue #2158, opened 2026-07-11
while updating claude_code, topgrade suspends shell (can resume & complete after bringing foreground)


charmingnyc:
## Checklist

- [ ✅] I have searched the issue tracker for relevant or duplicate issues.

## Erroneous Behavior

While updating claude_code package/module, topgrade suspends the shell (can resume & complete after bringing foreground, fg)

# topgrade
# fg

## Expected Behavior

Complete without error. On version 17.7, it does appear to be a new error; I don't recall a prior occurrence

## Steps to reproduce

Just running # topgrade
can avoid with topgrade --disable claude_code
when run topgrade --only claude_code results in OK

## Possible Cause (Optional)

Sorry, unsure. no hunches

## Problem persists without calling from topgrade

- [ ] Yes
- [ ✅] No

## Ran through `Remote Execution`

- [ ] Yes
- [ ✅] No

If yes, does the issue still occur when you run topgrade directly in your
remote host?

- [ ] Yes
- [ N/A] No

## Configuration file (Optional)

```toml

```

## Additional Details

- Operating System/Version
ProductName:		macOS
ProductVersion:		26.5.2
BuildVersion:		25F84

- Installation
 Homebrew or cargo probably; upgraded via cargo 

- Topgrade version (`topgrade -V`)
topgrade 17.7.0

## Verbose Output (`topgrade -v`)

```
-- only included Claude Code section 

── 10:26:53 - Claude Code ──────────────────────────────────────────────────────
DEBUG Executing command `/Users/$user/.local/bin/claude update`
Current version: 2.1.207
Checking for updates to latest version...
Claude Code is up to date (2.1.207)
DEBUG Executing command `/Users/$user/.local/bin/claude plugin marketplace update`
[2]  + 50000 suspended (tty output)  topgrade -v

# fg
[2]  - 72176 continued  topgrade --no-retry
✔ Successfully updated 1 marketplace(s)


# .local/bin/claude update                      10:06
Current version: 2.1.207
Checking for updates to latest version...
Claude Code is up to date (2.1.207)
```

NextMerge:
Might be related: I've experienced the same behaviour whenever topgrade presents a failure prompt. It will suspend the job and it needs to be brought back to the foreground.

When it hits the prompt:
```
Retry? (y)es/(N)o/(s)hell/(q)uit    fish: Job 1, 'topgrade' has stopped
```

And after running `fg`:
```
❱ fg
[...]

PR #2182, opened 2026-07-15
fix(zsh): drop redundant .zshrc source

### LINKED ISSUE #2152: zinit step fails when .zshrc guards against re-sourcing; explicit source is redundant under zsh -i
## Checklist

## Erroneous Behavior

The zinit step runs:

```txt
/usr/bin/zsh -i -c 'source /home/robert/.zshrc && zinit self-update && zinit update --all'
```

My `.zshrc` has a guard that rejects being sourced twice (re-sourcing corrupts zinit/p10k state in
an already-initialized shell):

```zsh
if [[ -n "$__ZSHRC_LOADED" ]]; then
    print -P "%F{red}Error:%f source ~/.zshrc is not supported (causes shell corruption)." >&2
    print -P "Use %F{green}reload%f or %F{green}exec zsh%f instead." >&2
    return 1
fi
__ZSHRC_LOADED=1
```

Because topgrade starts the shell with `-i`, zsh already sources `.zshrc` on startup. The explicit
`source ~/.zshrc` in the command body is a second load. My guard catches it and the step fails:

```txt
Error: source ~/.zshrc is not supported (causes shell corruption).
Use reload or exec zsh instead.
zinit failed:
   0: Command failed: `/usr/bin/zsh -i -c 'source /home/robert/.zshrc && zinit self-update && zinit update --all'`
   1: `/usr/bin/zsh` failed: exit status: 1

Location:
   src/steps/zsh.rs:108
```

## Expected Behavior

The zinit step updates zinit and its plugins. Dropping the explicit `source` is enough because `-i`
already loads `.zshrc`:

```console
$ zsh -i -c 'zinit self-update && zinit update --all'   # works
$ zsh -i -c 'zinit version'
zinit v3.15.0-1-gfcd2501f (linux-gnu_x86_64)
```

I realize most users don't guard their `.zshrc` like this, but the double-source affects everyone
running the zinit step. It re-runs plugin ice directives and prompt init in an already-initialized
shell. romkatv commented on this pattern in #646: sourcing `.zshrc` explicitly "has unpredictable
consequences and should never be done." It also looks like the suspected root cause of #238. The
same `source <zshrc> &&` pattern is in `run_zi`, `run_antigen`, and `run_zgenom`.

## Steps to reproduce

1. Add the guard above to the top of `~/.zshrc` (after any non-interactive early return)
2. Install zinit
3. Run `topgrade --only shell`
4. The zinit step exits 1 with the guard's error message

## Possible Cause (Optional)

`run_zinit` in `src/steps/zsh.rs` builds the command as
`source <zshrc> && zinit self-update && zinit update --all` and executes it with `zsh -i -c`. The
`-i` flag already causes zsh to source `.zshrc`, so the explicit `source` loads it a second time.

I'm happy to open a PR removing the explicit `source` from the zsh plugin manager steps if that
approach sounds right to you.

- [x] Yes
- [ ] No

```console
$ zsh -i -c 'source ~/.zshrc && zinit self-update && zinit update --all'
Error: source ~/.zshrc is not supported (causes shell corruption).
Use reload or exec zsh instead.
$ echo $?
1
```

- [ ] Yes
- [x] No

remote host?

- [ ] Yes
- [ ] No

## Additional Details

- Operation System/Version

  Fedora Linux 44, kernel 7.1.3-200.fc44.x86_64

- Installation

  mise (`aqua:topgrade-rs/topgrade`, prebuilt release binary)

  topgrade 17.6.2

<details>
<pre>
DEBUG Step "zinit"
DEBUG Detected "/usr/bin/zsh" as "zsh"
DEBUG Path "/home/robert/.zshrc" exists
DEBUG Path "/home/robert/.local/share/zinit" exists
―― 08:11:28 - zinit ――
DEBUG Executing command `/usr/bin/zsh -i -c 'source /home/robert/.zshrc && zinit self-update && zinit update --all'`
Error: source ~/.zshrc is not supported (causes shell corruption).
Use reload or exec zsh instead.
DEBUG Command failed: Err(
   0: Command failed: `/usr/bin/zsh -i -c 'source /home/robert/.zshrc && zinit self-update && zinit update --all'`
   1: `/usr/bin/zsh` failed: exit status: 1

Location:
   src/steps/zsh.rs:108)
DEBUG Step "zinit" failed:
   0: Command failed: `/usr/bin/zsh -i -c 'source /home/robert/.zshrc && zinit self-update && zinit update --all'`
   1: `/usr/bin/zsh` failed: exit status: 1

Location:
   src/steps/zsh.rs:108
</pre>
</details>

uwuclxdy:
## What does this PR do

Fixes #2152 

This PR removes sourcing of `.zshrc` on startup, because `zsh -i` already does that and the explicit `source` tripped zsh plugin-manager guards. This affects 5 steps, all now run `zsh -i -c '{something}'`. `zplug`, `zim` and `antidote` already did it like that before.

`zr`, `antigen` and `zgenom` steps now do `-i` instead of `-l`. That flag doesn't source `.zshrc`, so this PR would otherwise break them. This change means only that  `.zprofile` and `.zlogin` stop sourcing automatically. `.zshenv` and `.zshrc` still load and plugin managers also define their functions only in `.zshrc`.

Verified on Linux and macOS zsh 5.9: `zsh -i -c 'source zshrc && cmd'` double-sources and the guard exits 1; `zsh -i -c 'cmd'` sources once and exits 0.

## Standards checklist

before
```
―― 00:36:38 - zinit ――
Error: source ~/.zshrc is not supported (causes shell corruption).
Use reload or exec zsh instead.
zinit failed:
   0: Command failed: `/usr/bin/zsh -i -c 'source /home/robert/.zshrc && zinit self-update && zinit update --all'`
   1: `/usr/bin/zsh` failed: exit status: 1

Location:
   src/steps/zsh.rs:130
```

after
```
―― 00:38:19 - zinit ――
[self-update] fetching latest changes from main branch==> Already up to date.
Note: updating also unloaded plugins
The update took 0.38 seconds
```

### AI involvement

Consulted to help me understand the issue, propose a solution and to help with testing. FWIW, I wrote the code and let the LLM review it before PR.

### DIFF
--- a/src/steps/zsh.rs
+++ b/src/steps/zsh.rs
@@ -91,7 +91,6 @@ pub fn run_antibody(ctx: &ExecutionContext) -> Result<()> {
 
 pub fn run_antigen(ctx: &ExecutionContext) -> Result<()> {
     let zsh = require("zsh")?;
-    zshrc().require()?;
     env::var("ADOTDIR")
         .map_or_else(|_| HOME_DIR.join("antigen.zsh"), PathBuf::from)
         .require()?;
@@ -103,7 +102,6 @@ pub fn run_antigen(ctx: &ExecutionContext) -> Result<()> {
 
 pub fn run_zgenom(ctx: &ExecutionContext) -> Result<()> {
     let zsh = require("zsh")?;
-    zshrc().require()?;
     env::var("ZGEN_SOURCE")
         .map_or_else(|_| HOME_DIR.join(".zgenom"), PathBuf::from)
         .require()?;
@@ -115,7 +113,6 @@ pub fn run_zgenom(ctx: &ExecutionContext) -> Result<()> {
 
 pub fn run_zplug(ctx: &ExecutionContext) -> Result<()> {
     let zsh = require("zsh")?;
-    zshrc().require()?;
 
     env::var("ZPLUG_HOME")
         .map_or_else(|_| HOME_DIR.join(".zplug"), PathBuf::from)
@@ -128,7 +125,6 @@ pub fn run_zplug(ctx: &ExecutionContext) -> Result<()> {
 
 pub fn run_zinit(ctx: &ExecutionContext) -> Result<()> {
     let zsh = require("zsh")?;
-    zshrc().require()?;
 
     env::var("ZINIT_HOME")
         .map_or_else(|_| XDG_DIRS.data_dir().join("zinit"), PathBuf::from)
@@ -141,7 +137,6 @@ pub fn run_zinit(ctx: &ExecutionContext) -> Result<()> {
 
 pub fn run_zi(ctx: &ExecutionContext) -> Result<()> {
     let zsh = require("zsh")?;
-    zshrc().require()?;
 
     HOME_DIR.join(".zi").require()?;
 
--- a/src/steps/zsh.rs
+++ b/src/steps/zsh.rs
@@ -21,8 +21,7 @@ pub fn run_zr(ctx: &ExecutionContext) -> Result<()> {
 
     print_separator("zr");
 
-    let cmd = format!("source {} && zr --update", zshrc().display());
-    ctx.execute(zsh).args(["-l", "-c", cmd.as_str()]).status_checked()
+    ctx.execute(zsh).args(["-i", "-c", "zr --update"]).status_checked()
 }
 
 fn zdotdir() -> PathBuf {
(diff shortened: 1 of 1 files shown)


uwuclxdy:
This PR would re-introduce the zsh bug from https://github.com/topgrade-rs/topgrade/issues/2158

I will append `; exit $?` for now, would suggest that we move this somewhere else so that it happens automatically without needing to take care of ending commands on a builtin ourselves, until zsh addresses it.

uwuclxdy:
We could do something like this instead of appending `; exit $?` everywhere manually:

```rust
fn execute_interactive_zsh(ctx: &ExecutionContext, zsh: &Path, command: &str) -> Executor {
    let script = format!("{command}; exit $?");
    let mut exec = ctx.execute(zsh);
    exec.args(["-i", "-c", &script]);
    exec
}
```

Then we call it instead of `execute` + appending `; exit $?` manually. Would allow to track the upstream issue more easily as well, so if zsh fixes this then we just remove the helper and replace callers back with `execute`

GideonBear:
> We could do something like this instead of appending `; exit $?` everywhere manually:
> 
> ```rust
> fn execute_interactive_zsh(ctx: &ExecutionContext, zsh: &Path, command: &str) -> Executor {
>     let script = format!("{command}; exit $?");
>     let mut exec = ctx.execute(zsh);
>     exec.args(["-i", "-c", &script]);
>     exec
> }
> ```
> 
> Then we call it instead of `execute` + appending `; exit $?` manually. Would allow to track the upstream issue more easily as well, so if zsh fixes this then we just remove the helper and replace callers back with `execute`

Sounds good!

uwuclxdy:
comment by claude i think its good tho

GideonBear (approved):

PR #2186, opened 2026-07-15
NOSQUASH: fix(sudo): non-interactive sudo loop refresh and cache credentials upfront

uwuclxdy:
## What does this PR do

This PR changes `sudo -v` -> `sudo -n -v`, which doesn't re-prompt for password if it expired. The reasoning behind this is that if the cache expires for any reason, we should not re-prompt just for the sake of doing that; let the next step that actually needs those credentials re-prompt (if there even is one). We could maybe surface a warning when the cache expires but I'm unsure.

Flags that we use are documented in comments now.

Topgrade now also prompts for sudo password at the start (so the loop is guaranteed to have credentials to refresh) if:

```toml
# pre_sudo = true
sudo_loop = true
```

## Standards checklist

### AI involvement

I let Claude diagnose the issue, describe the solution, write only some failing tests and then review (and fix) my code after I got it to pass the tests. It also split changes in two commits.

### DIFF
--- a/src/sudo.rs
+++ b/src/sudo.rs
@@ -341,6 +341,22 @@ impl Sudo {
         cmd.status_checked().wrap_err("Failed to elevate permissions")
     }
 
+    /// Refresh arguments for the sudo kinds that can cache credentials.
+    fn refresh_args(&self) -> Option<&'static [&'static str]> {
+        match self.kind {
+            // `-v`: on a still-valid timestamp, extends it without touching PAM.
+            SudoKind::Sudo => Some(&["-v"]),
+            // `-w`: refreshes a still-valid token without prompting; only prompts once cold.
+            SudoKind::Please => Some(&["-w"]),
+            _ => None,
+        }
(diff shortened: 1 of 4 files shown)


src/sudo.rs
```diff
@@ -342,6 +342,22 @@ impl Sudo {
         cmd.status_checked().wrap_err("Failed to elevate permissions")
     }
 
+    /// Refresh arguments for the sudo kinds that can cache credentials.
+    fn refresh_args(kind: SudoKind) -> Option<&'static [&'static str]> {
```
GideonBear: Why doesn't this take `&self`? In the two places where it is used, `self.kind` is passed anyway.

GideonBear (changes requested):

src/sudo.rs
uwuclxdy: yep mb, missed this

GideonBear (approved):

PR #1946, opened 2026-04-10
chore(deps): update rust crate merge to 0.2

niStee:
## What does this PR do

Updates the `merge` crate from 0.1 to 0.2 to resolve security advisory [RUSTSEC-2024-0370](https://rustsec.org/advisories/RUSTSEC-2024-0370.html).

fix #1558

`merge` 0.2 removed the blanket `impl Merge for Option<T>`. This PR adds explicit `#[merge(strategy = merge::option::overwrite_none)]` annotations to all bare `Option<T>` fields in config structs, preserving the exact same merge semantics (first-loaded config wins for scalar fields).

A regression test is included to guard against future merge crate upgrades silently changing config precedence.

## Standards checklist

- [x] The PR title is descriptive
- [x] I have read `CONTRIBUTING.md`
- [x] *Optional:* I have tested the code myself, with the relevant tools installed. If yes, add Topgrade's output of the relevant steps.
- [x] If this PR introduces new user-facing messages they are translated

### AI involvement

AI (Claude Code) was used to identify all ~79 bare `Option<T>` fields needing annotation and to draft the regression test. All changes were reviewed and verified by the author.

## For new steps

N/A — no new steps added.

PR #2211, opened 2026-07-21
refactor: migrate from merge to merge2

### LINKED ISSUE #2162: `code that will be rejected by a future version of Rust` / RUSTSEC-2026-0173 (`proc-macro-error2` via `merge`)
`merge` doesn't seem very maintained. `merge2` might be an alternative.

GideonBear:
Ref #1946, partially reverts that since merge2 is based on merge v0.1.
Closes #2162

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -314,16 +314,15 @@ pub fn is_elevated() -> bool {
 }
 
 pub mod merge_strategies {
+    use std::hash::Hash;
     use std::mem;
-
-    use crate::config::Commands;
+    use indexmap::IndexMap;
 
     /// Prepends right to left (both Option<Vec<T>>)
-    pub fn vec_prepend_opt<T: Clone>(left: &mut Option<Vec<T>>, right: &mut Option<Vec<T>>) {
+    pub fn vec_prepend_opt<T>(left: &mut Option<Vec<T>>, right: &mut Option<Vec<T>>) {
         if let Some(left_vec) = left {
             if let Some(right_vec) = right {
-                right_vec.append(left_vec);
-                let _ = left.replace(right_vec.clone());
+                merge2::vec::prepend(left_vec, right_vec);
             }
         } else {
             mem::swap(left, right);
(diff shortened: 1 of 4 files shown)


src/utils.rs
```diff
@@ -314,54 +314,41 @@ pub fn is_elevated() -> bool {
...
     /// Prepends right to left (both Option<Vec<T>>)
-    pub fn vec_prepend_opt<T>(left: &mut Option<Vec<T>>, right: Option<Vec<T>>) {
+    pub fn vec_prepend_opt<T: Clone>(left: &mut Option<Vec<T>>, right: &mut Option<Vec<T>>) {
         if let Some(left_vec) = left {
-            if let Some(mut right_vec) = right {
+            if let Some(right_vec) = right {
                 right_vec.append(left_vec);
-                let _ = left.replace(right_vec);
+                let _ = left.replace(right_vec.clone());
             }
```
uwuclxdy, line 327:
```suggestion
    pub fn vec_prepend_opt<T>(left: &mut Option<Vec<T>>, right: &mut Option<Vec<T>>) {
        if let Some(left_vec) = left {
            if let Some(right_vec) = right {
                merge2::vec::prepend(left_vec, right_vec);
            }
    }
```

`merge2::vec::prepend` already does exactly this, so we can drop the clone and the T: Clone bound with it.

src/utils.rs
```diff
@@ -314,54 +314,41 @@ pub fn is_elevated() -> bool {
...
 
-    pub fn inner_merge_opt<T>(left: &mut Option<T>, right: Option<T>)
-    where
-        T: Merge,
-    {
+    pub fn commands_merge_opt(left: &mut Option<Commands>, right: &mut Option<Commands>) {
         if let Some(left_inner) = left {
             if let Some(right_inner) = right {
-                left_inner.merge(right_inner);
+                left_inner.extend(right_inner.clone());
```
uwuclxdy:
```suggestion
                left_inner.extend(mem::take(right_inner));
```

`mem::take` skips the clone, same as merge2's own hashmap

uwuclxdy (approved):
i see no issues with this thanks! however, claude does so do what you will

PR #2214, opened 2026-07-22
fix(config)!: ignore config's `only` if `--only` is provided

NiekvD-09:
## What does this PR do
Changes the behaviour of enabled steps from including steps in the config `only` as well as steps from the `--only` flag to skipping the config `only` steps if the `--only` flag is provided
Closes #1930

## Standards checklist

### AI involvement

No AI used

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1211,16 +1211,14 @@ impl Config {
         let mut enabled_steps = (!opt.only.is_empty())
             .then(|| opt.only.clone())
             .or_else(|| {
-                config_file.misc.as_ref()
+                config_file
+                    .misc
+                    .as_ref()
                     .and_then(|m| m.only.as_ref())
                     .filter(|v| !v.is_empty())
                     .cloned()
             })
-            .unwrap_or_else(|| {
-                Step::iter()
-                    .filter(|x| !step_is_deprecated(*x))
-                    .collect()
-        });
+            .unwrap_or_else(|| Step::iter().filter(|x| !step_is_deprecated(*x)).collect());
 
         let mut disabled_steps: Vec<Step> = Vec::new();
         disabled_steps.extend(&opt.disable);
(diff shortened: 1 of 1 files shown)


uwuclxdy (approved):
Looks good, thanks! This is a "breaking change" tho and will probably get deferred to v18. I personally wouldn't have an issue with merging.

GideonBear:
Tested and working!
I'll merge this into the v18 branch.

src/config.rs
```diff
@@ -1180,28 +1203,22 @@ impl Config {
...
-        if enabled_steps.is_empty() {
-            // All steps are enabled
-            enabled_steps.extend(Step::iter());
-            // Handle deprecated steps. Disable automatically when not explicitly mentioned in the
-            // config, so that the warning doesn't print.
-            enabled_steps.retain(|x| !step_is_deprecated(*x));
-        }
+        // the --only flag has priority over the config file
+        // if there aren't any only steps specified all steps must be ran
+        // deprecated steps are disabled when not explicitly mentioned in only
```
```suggestion
        // The --only flag has priority over the config file.
        // If there aren't any only steps specified all steps must be ran.
        // Deprecated steps are disabled when not explicitly mentioned in only.
```
I intuitively tried to read this as a single sentence, which it isn't

src/config.rs
```diff
@@ -1180,28 +1203,22 @@ impl Config {
...
+        let mut enabled_steps = (!opt.only.is_empty())
+            .then(|| opt.only.clone())
```
GideonBear, line 1212: `bool.then` is ugly IMO, you can just use an `if` instead:
```suggestion
        let mut enabled_steps = if !opt.only.is_empty() {
            opt.only.clone())
        } else {
            ...
```

GideonBear (changes requested):
Nits, looks good!

src/config.rs
NiekvD-09: IMHO the original is more readable and looks nicer

uwuclxdy: Yes I agree with @NiekvD-09

GideonBear: I still disagree (although less than before your explanation on Discord, thanks), but by majority vote we'll go with your way

GideonBear (approved):
Thanks!

PR #2219, opened 2026-07-23
feat: add Zellij support

### LINKED ISSUE #864: Enable zellij as multiplexer besides tmux
## I want to suggest a new ste
No!

  * Zellij, a tmux-style terminal multiplexer https://github.com/zellij-org/zellij
  * macos, linux, windows (wsl), etc.
  * n/a
  * No commands, but support to open up multiple remote connections, but instead of using tmux, use zellij. 
  * n/a
  option to skip this step?
 * n/a

Topgrade should...
... be able to use zellij besides tmux as a terminal multiplexer. Zellij lives in the same realm of rust tools and would be an awesome addition;

mxxun:
## What does this PR do

This implements zellij support parallel to that of tmux. In particular (following the commits):
- [x] topgrade can now run itself inside a new zellij session
- [x] there's the -z/--zellij flag to control this
- [x] there's a run_in_zellij config option and a --no-zellij flag to override it
- [x] there's a zellij_session_mode config option parallel to tmux_session_mode
- [x] there's a zellij_arguments config option parallel to tmux_arguments
- [x] topgrade runs a fresh zellij session on each run, preserving the previous ones
- [x] ssh_step runs remote topgrade sessions in zellij if asked for, parallel to the same logic for tmux

Closes #864.

## Standards checklist

- [x] I have tested the code myself. topgrade's output is i think irrelevant, since what i'm testing is whether it runs in a multiplexer on various configs. here are the tests that i did and results i expected (and got), for convenience of reproduction.
    - (all tests assume an installed zellij; see https://zellij.dev/documentation/installation)
    - `topgrade --zellij` runs in a new zellij session called `topgrade`, with a single tab in a single pane. (`zellij ls` to list all sessions.) when topgrade finishes, the session output persists.
    - (regression test: absent `--zellij`, topgrade runs where launched)
    - with `run_in_zellij=true` in the config, plain `topgrade` also runs in zellij
    - with `run_in_zellij=true` in the config, `topgrade --no-zellij` runs where launched
    - with `zellij_session_mode = "attach_if_not_in_session"` in the config (default), when run inside a zellij session, topgrade runs in a new background session (and reports that it does so)
    - with `zellij_session_mode = "attach_always"` in the config, when run inside a zellij session, topgrades switches zellij to the newly launched session
    - with `zellij_arguments = "--help"`, topgrade indeed passes `--help` to zellij, which results in zellij's help being printed and no topgrade-inside-zellij being run. (i have not tested zellij_arguments beyond this, since i don't have a use-case for it to hand.)
    - when run multiple times, `topgrade --zellij` uses uniquely named sessions so as to not override old ones
    - (TODO) when run with an ssh step, `topgrade --zellij` runs topgrade on individual remotes in their named tabs/panes in zellij's session. this, i don't have remotes at hand to test with.
    - when run with both --tmux and --zellij, topgrade fails, reporting that it was required to run in two multiplexers at once.
    - for "Topgrade launched in a new zellij session", i reused the tmux message, parametrizing the translation
    - likewise for "Remote Topgrade launched in Zellij"
    - there's a new error message, "Multiple multiplexers specified", untranslated. other fatal configuration failures are also untranslated, but perhaps a new one should be.
    - likewise for "Failed to parse `zellij_arguments`"
    - likewise for "Error running Topgrade in zellij"
    - likewise for "Failed to `execvp(3)` zellij"
    - likewise for "Could not find zellij", though this one can be easily fixed: there's a translation of "Could not find sudo" that can be parametrized with an arbitrary not-found binary

### AI involvement
For the first five commits/subfeatures, I did not use AI/LLMs. The last two (new_unique_window and ssh_step), I pair-programmed with claude code and reviewed myself. All of this is code and comments I would write, or did write.

### DIFF
--- a/src/steps/remote/ssh.rs
+++ b/src/steps/remote/ssh.rs
@@ -29,8 +29,8 @@ pub fn ssh_step(ctx: &ExecutionContext, hostname: &str) -> Result<()> {
     // run remotes in the multiplexer if one is specified
     #[cfg(unix)]
     if !ctx.run_type().dry() {
-        match ctx.config().run_in_multiplexer() {
-            Multiplexer::No => { }
+        match ctx.config().run_in_multiplexer()? {
+            Multiplexer::No => {}
             Multiplexer::Tmux => {
                 prepare_async_ssh_command(&mut args);
                 crate::tmux::run_command(ctx, hostname, &shell_words::join(args))?;
--- a/src/steps/remote/ssh.rs
+++ b/src/steps/remote/ssh.rs
@@ -5,6 +5,9 @@ use crate::{
     command::CommandExt, error::SkipStep, execution_context::ExecutionContext, terminal::print_separator, utils,
 };
 
+#[cfg(unix)]
+use crate::config::Multiplexer;
+
 fn prepare_async_ssh_command(args: &mut Vec<&str>) {
     args.insert(0, "ssh");
     args.push("--keep");
@@ -23,11 +26,18 @@ pub fn ssh_step(ctx: &ExecutionContext, hostname: &str) -> Result<()> {
     let env = format!("TOPGRADE_PREFIX={hostname}");
     args.extend(["env", &env, "$SHELL", "-lc", topgrade]);
 
+    // run remotes in the multiplexer if one is specified
     #[cfg(unix)]
-    if ctx.config().run_in_tmux() && !ctx.run_type().dry() {
-        prepare_async_ssh_command(&mut args);
-        crate::tmux::run_command(ctx, hostname, &shell_words::join(args))?;
-        return Err(SkipStep(String::from(t!("Remote Topgrade launched in Tmux"))).into());
+    if !ctx.run_type().dry() {
+        match ctx.config().run_in_multiplexer() {
+            Multiplexer::No => { }
+            Multiplexer::Tmux => {
+                prepare_async_ssh_command(&mut args);
+                crate::tmux::run_command(ctx, hostname, &shell_words::join(args))?;
+                return Err(SkipStep(String::from(t!("Remote Topgrade launched in Tmux"))).into());
+            }
+            Multiplexer::Zellij => !todo!(),
+        }
     }
 
     if ctx.config().open_remotes_in_new_terminal() && !ctx.run_type().dry() && cfg!(windows) {
--- a/src/steps/remote/ssh.rs
+++ b/src/steps/remote/ssh.rs
@@ -32,11 +32,26 @@ pub fn ssh_step(ctx: &ExecutionContext, hostname: &str) -> Result<()> {
         match ctx.config().run_in_multiplexer()? {
             Multiplexer::No => {}
             Multiplexer::Tmux => {
+                // FIXME: here, we ask tmux to run `ssh ... $SHELL -lc (topgrade) --keep`
+                // and $SHELL does not pass the last --keep to topgrade. we'd need to wire in
+                // format!("'{topgrade} --keep'") as a single arg, conditional on tmux.
                 prepare_async_ssh_command(&mut args);
                 crate::tmux::run_command(ctx, hostname, &shell_words::join(args))?;
-                return Err(SkipStep(String::from(t!("Remote Topgrade launched in Tmux"))).into());
+                return Err(SkipStep(String::from(t!(
+                    "Remote Topgrade launched in {multiplexer}",
+                    multiplexer = "Tmux"
+                )))
+                .into());
             }
-            Multiplexer::Zellij => !todo!(),
+            Multiplexer::Zellij => {
+                // NB: no need to --keep, since zellij keeps panes on exit
+                crate::zellij::run_command(ctx, hostname, "ssh", &args)?;
+                return Err(SkipStep(String::from(t!(
+                    "Remote Topgrade launched in {multiplexer}",
+                    multiplexer = "Zellij"
+                )))
+                .into());
+            }
         }
     }
 
--- a/src/zellij.rs
+++ b/src/zellij.rs
@@ -11,6 +11,7 @@ use crate::config::ZellijConfig;
 use crate::config::ZellijSessionMode;
 use crate::utils::which;
 
+use rust_i18n::t;
 #[cfg(unix)]
 use std::os::unix::process::CommandExt as _;
 
@@ -99,16 +100,21 @@ pub fn run_in_zellij(config: ZellijConfig) -> Result<()> {
 
     let is_inside_zellij = env::var("ZELLIJ").is_ok();
     let err = match config.session_mode {
-        // TODO: zellij AttachIfNotInSession
-        // ZellijSessionMode::AttachIfNotInSession => {
-        //     if is_inside_zellij {
-        //         // Only attach to the newly-created session if we're not currently in a zellij session.
-        //         println!("{}", t!("Topgrade launched in a new zellij session"));
-        //         return Ok(());
-        //     } else {
-        //         zellij.build().args(["attach-session", "-t", &session]).exec()
-        //     }
-        // }
+        ZellijSessionMode::AttachIfNotInSession => {
+            if is_inside_zellij {
+                // Only attach to the newly-created session if we're not currently in a zellij session.
+                println!(
+                    "{}",
+                    t!(
+                        "Topgrade launched in a new {multiplexer} session",
+                        multiplexer = "zellij"
+                    )
+                );
+                return Ok(());
+            } else {
+                zellij.build().args(["attach", &session]).exec()
+            }
+        }
         ZellijSessionMode::AttachAlways => {
             if is_inside_zellij {
                 zellij.build().args(["action", "switch-session", &session]).exec()
--- a/src/zellij.rs
+++ b/src/zellij.rs
@@ -1,0 +1,122 @@
+use std::env;
+use std::path::PathBuf;
+use std::process::Command;
+
+use color_eyre::eyre::Context;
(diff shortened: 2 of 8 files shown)


src/steps/remote/ssh.rs
```diff
@@ -23,11 +26,33 @@
...
     #[cfg(unix)]
-    if ctx.config().run_in_tmux() && !ctx.run_type().dry() {
-        prepare_async_ssh_command(&mut args);
-        crate::tmux::run_command(ctx, hostname, &shell_words::join(args))?;
-        return Err(SkipStep(String::from(t!("Remote Topgrade launched in Tmux"))).into());
+    if !ctx.run_type().dry() {
+        match ctx.config().run_in_multiplexer()? {
+            Multiplexer::No => {}
+            Multiplexer::Tmux => {
+                // FIXME: here, we ask tmux to run `ssh ... $SHELL -lc (topgrade) --keep`
```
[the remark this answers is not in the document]
mxxun: this describes an apparent bug in the logic of running ssh_step in tmux. I decided to document it but not to fix it in this PR.

src/zellij.rs
```diff
@@ -0,0 +1,167 @@
...
+    args: Option<Vec<String>>,
+}
+
+impl Zellij {
+    fn new(args: Vec<String>) -> Self {
+        Self {
+            zellij: which("zellij").expect("Could not find zellij"),
+            args: if args.is_empty() { None } else { Some(args) },
+        }
+    }
```
uwuclxdy:
```suggestion
    fn new(args: Vec<String>) -> Result<Self> {
    	Ok(Self {
        	zellij: which("zellij")
            	.ok_or_else(|| eyre!(t!("Cannot find {binary_name} in PATH", binary_name = "zellij")))?,
        	args: if args.is_empty() { None } else { Some(args) },
    	})
	}
```

src/zellij.rs
```diff
@@ -0,0 +1,167 @@
...
+        Ok(())
+    }
+
+    /// Names of all zellij sessions, including EXITED ones (which still occupy their name).
+    fn session_names(&self) -> Result<HashSet<String>> {
+        let output = self
+            .build()
+            .args(["list-sessions", "--short", "--no-formatting"])
+            // exits with status 1 when there are no sessions, which is fine
+            .output_checked_with_utf8(|_| Ok(()))
```
uwuclxdy:
```suggestion
            .output_checked_with_utf8(|output| {
                if output.status.code() == Some(1) && output.stderr.contains("No active zellij sessions found") {
                    Ok(())
                } else {
                    Err(())
                }
            })
```

src/zellij.rs
```diff
@@ -0,0 +1,167 @@
...
+            .context("Error listing zellij sessions")?;
+        Ok(output.stdout.lines().map(str::to_owned).collect())
+    }
+
+    /// Like [`new_session`] but it appends a digit to the session name (if necessary) to
+    /// avoid duplicate session names.
+    ///
+    /// The session name is returned.
+    fn new_unique_session(&self, session_name: &str, tab_name: &str, command: &str, args: &[&str]) -> Result<String> {
+        let existing = self.session_names().context("Error listing zellij sessions")?;
```
uwuclxdy:
```suggestion
        let existing = self.session_names()?;
```

src/steps/remote/ssh.rs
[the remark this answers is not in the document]
uwuclxdy: could you open an issue about it?

uwuclxdy (commented):
`run_in_multiplexer = "zellij"` was proposed in #864, which might be better than `run_in_zellij = true` for future steps that we might want to run in a multiplexer (maybe for parallel updates or something)

src/steps/remote/ssh.rs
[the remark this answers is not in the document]
mxxun: Sure: #2280

mxxun:
> `run_in_multiplexer = "zellij"` was proposed in #864, which might be better than `run_in_zellij = true` for future steps that we might want to run in a multiplexer (maybe for parallel updates or something)

I considered doing that, and leaned against; I didn't like the prospect of breaking the backward compatibility of either CLI flags or config files.

uwuclxdy:
good point.. then let's keep it like this until the next breaking release (v18) i guess

mxxun:
@uwuclxdy so! Should we go ahead an merge this? Or did I miss something blocking.

uwuclxdy:
hi, sorry it's taking so long. I'll review it today

src/zellij.rs
```diff
@@ -0,0 +1,174 @@
...
+impl Zellij {
+    fn new(args: Vec<String>) -> Result<Self> {
+        Ok(Self {
+            zellij: which("zellij")?
+                .ok_or_else(|| eyre!(t!("Cannot find {binary_name} in PATH", binary_name = "zellij")))?,
+            args: if args.is_empty() { None } else { Some(args) },
+        })
+    }
+
+    #[allow(clippy::disallowed_methods)]
```
uwuclxdy, line 34:
```suggestion
    #[expect(clippy::disallowed_methods)]
```

nit

src/zellij.rs
```diff
@@ -0,0 +1,174 @@
...
+        let mut command = Command::new(&self.zellij);
+        // NB: unlike tmux, zellij seems to nest fine without any env-var wrangling.
+        if let Some(args) = self.args.as_ref() {
+            command.args(args);
+        }
+        command
+    }
+    /// Create a new zellij session with the given name, running `command` with `args` in a tab
+    /// (and pane) named `tab_name`.
+    fn new_session(&self, session_name: &str, tab_name: &str, command: &str, args: &[&str]) -> Result<()> {
```
uwuclxdy, line 45:
```suggestion
    fn new_session(&self, session_name: &str, command: &str, args: &[&str]) -> Result<()> {
	    self.build()
	        .args(["attach", "--create-background", session_name, "--"])
	        .arg(command)
	        .args(args)
	        .output_checked()
	}
```

major: don't reimplement Zellij's `attach --create-background <name> -- <cmd> <args...>` with a hand-built KDL string (at the cost of an unnamed first tab and a raised minimum of zellij 0.45.0, feel free to argue against this as I don't have experience with Zellij)

src/zellij.rs
```diff
@@ -0,0 +1,174 @@
...
+        // rather than make a temp-file, we spawn zellij with a default layout, then replace it with ours.
+
+        // for that, we'll need a layout string approximately of form:
+        // `layout {tab {pane command="env" {args (env args) "topgrade" (topgrade args);};};}`
+        // with all args double-quoted.
+        // see https://zellij.dev/documentation/creating-a-layout.html for reference.
+        let mut args_kdl = String::new();
+        for arg in args {
+            // append double-quoted ` "arg"`, escaping double-quotes inside arg itself
+            args_kdl.push_str(&format!(" \"{}\"", arg.replace("\"", "\\\"")));
```
uwuclxdy, line 61: major: the escaping handles `"` but not `\`, so an argument containing a backslash (topgrade's own argv is replayed through `env::args()`) produces a broken layout string and the relaunch dies. If the builder stays, escape backslashes first, then quotes: `arg.replace('\\', "\\\\").replace('"', "\\\"")`.

src/zellij.rs
```diff
@@ -0,0 +1,174 @@
...
+        }
+        let layout_string = format!(
+            r#"layout {{ tab name="{tab_name}" {{ pane name="{tab_name}" command="{command}" {{ args {args_kdl}; }}; }}; }}"#
+        );
+        self.build()
+            .env("ZELLIJ_SESSION_NAME", session_name)
```
uwuclxdy, line 67: minor: `ZELLIJ_SESSION_NAME` is only consulted when more than one session is active and no `--session` is given; with one active session the action routes to it regardless of the env var, which is correct here only because topgrade's session is the only one or the fallback fires. Pass `--session <name>` before the subcommand here and in `new_tab` (line 76); it's honored in both branches and errors loudly on a mismatch.

src/zellij.rs
```diff
@@ -0,0 +1,174 @@
...
+            .args(["action", "override-layout", "--layout-string", &layout_string])
```
uwuclxdy, line 68: major: explain minimum version (here or in `config.example.toml`) of Zellij required for this; >=0.44.1

src/steps/remote/ssh.rs
```diff
@@ -23,11 +26,33 @@ pub fn ssh_step(ctx: &ExecutionContext, hostname: &str) -> Result<()> {
...
+                crate::tmux::run_command(ctx, hostname, &shell_words::join(args))?;
+                return Err(SkipStep(String::from(t!(
+                    "Remote Topgrade launched in {multiplexer}",
+                    multiplexer = "Tmux"
+                )))
+                .into());
+            }
+            Multiplexer::Zellij => {
+                // NB: no need to --keep, since zellij keeps panes on exit
+                crate::zellij::run_command(ctx, hostname, "ssh", &args)?;
```
uwuclxdy, line 48: major: this path is the one the PR body marks untested, and the remotes land in a new hidden `topgrade-N` background session rather than the one topgrade is running in, while the SkipStep message doesn't say where. Please run `topgrade --zellij` with `remote_topgrades = ["localhost"]` and show where the tab lands, then either reuse the current session (`ZELLIJ_SESSION_NAME` is set in-pane) or name the hidden session in the message.

uwuclxdy (changes requested):
This PR makes `output_checked_with_utf8` used, so the `#[allow(dead_code)]` and the `currently unused` comment above it in `src/command.rs` are stale.

PR #2226, opened 2026-07-24
NOSQUASH: added arg functions and refactored with them

### LINKED ISSUE #2224: Add `.arg_if`/`.args_if`
Idea by @NiekvD-09 from https://github.com/topgrade-rs/topgrade/pull/2216#issuecomment-5051802623
Assigned to @NiekvD-09

`.arg_if(cond, arg)`, not `.arg_if(arg, cond)`, since it should read as `if conf { .arg(arg) }`

### LINKED ISSUE #2225: Refactor steps using `.arg_if`/`.args_if`
We can remove a lot of `let mut command = ctx.execute(...` assignments

NiekvD-09:
Closes #2224
Closes #2225

Added the functions:
`arg_if(cond,arg)`
`arg_if_some(option<t>, fn(t)->arg)`
~`arg_if_else(cond,true_arg,false_arg)`~*
and their `args` counterparts

Also refactored _almost_ all steps where these functions could be used (Sisyphus would be jealous)

*`arg_if_else` has been removed in favour of `arg(if {} else {})`

## Standards checklist

### AI involvement

No AI used

### DIFF
--- a/src/executor.rs
+++ b/src/executor.rs
@@ -79,6 +79,49 @@ impl Executor {
     }
 
     #[allow(dead_code)]
+    /// See `std::process::Command::arg`
+    pub fn arg_if<S: AsRef<OsStr>>(&mut self, cond: bool, arg: S) -> &mut Executor {
+        if cond { self.arg(arg) } else { self }
+    }
+
+    #[allow(dead_code)]
+    /// See `std::process::Command::args`
+    pub fn args_if<I, S>(&mut self, cond: bool, args: I) -> &mut Executor
+    where
+        I: IntoIterator<Item = S>,
+        S: AsRef<OsStr>,
+    {
+        if cond { self.args(args) } else { self }
+    }
+
+    #[allow(dead_code)]
+    /// See `std::process::Command::arg`
+    pub fn arg_if_some<T, F, S>(&mut self, opt: Option<T>, f: F) -> &mut Executor
+    where
+        F: FnOnce(T) -> S,
+        S: AsRef<OsStr>,
+    {
+        if let Some(val) = opt {
+            self.arg(f(val));
+        }
+        self
+    }
+
+    #[allow(dead_code)]
+    /// See `std::process::Command::args`
+    pub fn args_if_some<T, F, I, S>(&mut self, opt: Option<T>, f: F) -> &mut Executor
+    where
+        F: FnOnce(T) -> I,
+        I: IntoIterator<Item = S>,
+        S: AsRef<OsStr>,
+    {
+        if let Some(val) = opt {
+            self.args(f(val));
+        }
+        self
+    }
+
+    #[allow(dead_code)]
     /// See `std::process::Command::current_dir`
     pub fn current_dir<P: AsRef<Path>>(&mut self, dir: P) -> &mut Executor {
         match self {
@@ -142,6 +185,16 @@ impl Executor {
         }
 
         self
+    }
+
+    #[allow(dead_code)]
+    /// See `std::process::Command::env`
+    pub fn env_if<K, V>(&mut self, cond: bool, key: K, val: V) -> &mut Executor
+    where
+        K: AsRef<OsStr>,
+        V: AsRef<OsStr>,
+    {
+        if cond { self.env(key, val) } else { self }
     }
 
     /// See `std::process::Command::spawn`
--- a/src/steps/generic.rs
+++ b/src/steps/generic.rs
@@ -77,18 +77,12 @@ pub fn run_cargo_update(ctx: &ExecutionContext) -> Result<()> {
         return Err(SkipStep(message).into());
     };
 
-    let mut command = ctx.execute(cargo_update);
-    command.args(["install-update", "--all"]);
-    if ctx.config().cargo_update_git() {
-        command.arg("--git");
-    }
-    if ctx.config().cargo_update_quiet() {
-        command.arg("--quiet");
-    }
-    if ctx.config().cargo_update_locked() {
-        command.arg("--locked");
-    }
-    command.status_checked()?;
+    ctx.execute(cargo_update)
+        .args(["install-update", "--all"])
+        .arg_if(ctx.config().cargo_update_git(), "--git")
+        .arg_if(ctx.config().cargo_update_quiet(), "--quiet")
+        .arg_if(ctx.config().cargo_update_locked(), "--locked")
+        .status_checked()?;
 
     if ctx.config().cleanup() {
         let cargo_cache = require("cargo-cache")
@@ -111,14 +105,11 @@ pub fn run_flutter_upgrade(ctx: &ExecutionContext) -> Result<()> {
     let flutter = require("flutter")?;
 
     print_separator("Flutter");
-    let mut command = ctx.execute(flutter);
-    command.arg("upgrade");
 
-    if ctx.config().flutter_force() {
-        command.arg("--force");
-    }
-
-    command.status_checked()
+    ctx.execute(flutter)
+        .arg("upgrade")
+        .arg_if(ctx.config().flutter_force(), "--force")
+        .status_checked()
 }
 
 pub fn run_gem(ctx: &ExecutionContext) -> Result<()> {
@@ -127,16 +118,14 @@ pub fn run_gem(ctx: &ExecutionContext) -> Result<()> {
 
     print_separator("Gems");
 
-    let mut command = ctx.execute(gem);
-    command.arg("update");
-
-    if env::var_os("RBENV_SHELL").is_none() {
-        command.arg("--user-install");
-    } else {
+    if !env::var_os("RBENV_SHELL").is_none() {
         debug!("Detected rbenv. Avoiding --user-install");
     }
 
-    command.status_checked()
+    ctx.execute(gem)
+        .arg("update")
+        .arg_if(env::var_os("RBENV_SHELL").is_none(), "--user-install")
+        .status_checked()
 }
 
 pub fn run_rubygems(ctx: &ExecutionContext) -> Result<()> {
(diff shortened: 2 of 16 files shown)


GideonBear:
I squashed and reworded your commits, take care when pulling back (`git reset --hard origin/argif`)

src/executor.rs
```diff
@@ -78,6 +78,114 @@ impl Executor {
...
+            }
+            Executor::Dry(c) => {
+                c.args.push(arg.as_ref().into());
+            }
+        }
+
+        self
+    }
+
+    #[allow(dead_code)]
```
GideonBear, line 87:
```suggestion
```

src/executor.rs
```diff
@@ -78,6 +78,114 @@ impl Executor {
...
+            Executor::Wet(c) | Executor::Damp(c) => {
+                c.arg(arg);
+            }
+            Executor::Dry(c) => {
+                c.args.push(arg.as_ref().into());
+            }
+        }
+
+        self
+    }
```
GideonBear, line 85: On all of these (incl `env_if`, excl `arg_if_some`); can you just:
```suggestion
    pub fn arg_if<S: AsRef<OsStr>>(&mut self, cond: bool, arg: S) -> &mut Executor {
        if cond {
            self.arg(arg)
        } else {
        	self
        }
    }
```

GideonBear (changes requested):

src/steps/os/linux.rs
```diff
@@ -252,62 +248,43 @@ fn upgrade_redhat(ctx: &ExecutionContext) -> Result<()> {
...
-
-    if let Some(args) = ctx.config().dnf_arguments() {
-        command.args(args.split_whitespace());
-    }
-
-    if ctx.config().yes(Step::System) {
-        command.arg("-y");
-    }
+    sudo.execute(ctx, &dnf)?
+        .arg_if_else(ctx.config().redhat_distro_sync(), "distro-sync", "upgrade")
```
GideonBear: Unsure if I like this (see discord), adding comment for tracking

src/steps/os/linux.rs
```diff
@@ -830,13 +744,11 @@ fn upgrade_nixos(ctx: &ExecutionContext) -> Result<()> {
...
-            command.args(["switch", "--upgrade"]);
-
-            if let Some(args) = ctx.config().nix_arguments() {
-                command.args(args.split_whitespace());
-            }
-            command.status_checked()?;
+            sudo.execute(ctx, "/run/current-system/sw/bin/nixos-rebuild")?
+                .args(["switch", "--upgrade"])
+                .args_if_some(ctx.config().nix_arguments(), |args| {
+                    args.split_whitespace() })
```
GideonBear:
```suggestion
                .args_if_some(ctx.config().nix_arguments(), |args| {
                    args.split_whitespace()
                })
```
Not enforced by the formatter

src/steps/generic.rs
```diff
@@ -1130,17 +1112,13 @@ pub fn run_chezmoi_update(ctx: &ExecutionContext) -> Result<()> {
     )
     .require()?;
 
-    let mut cmd = ctx.execute(&chezmoi);
-
+    // does order matter?
```
GideonBear: Not in this case.
(for next time, add `TODO:` in front)

GideonBear (commented):
Did a quick pass over the second commit.
Thanks! ❤️

GideonBear:
There's quite a few assignments left (`git grep "let mut .* = ctx"`), most of which should also be compactible. Can you do those as well?

src/executor.rs
NiekvD-09: Oh yeah, oopsie. Will change later today

NiekvD-09:
> There's quite a few assignments left (`git grep "let mut .* = ctx"`), most of which should also be compactible. Can you do those as well?

There are a handful left, namely:
`sudo.rs:421` : nvm i can just change this one

`utils.rs:237, zsh.rs:25, powershell:118, linux.rs:551,1140,1157,1189` : these all return an owned Executor, i asked in the discord what the best solution is

`generic.rs:1515,2233` : i think this is the best solution

src/executor.rs
NiekvD-09: Can you close this one?

src/steps/generic.rs
GideonBear:
```suggestion
```

src/steps/powershell.rs
```diff
@@ -115,19 +115,17 @@ impl Powershell {
     where
         S: AsRef<OsStr>,
     {
-        let mut command = ctx.execute(&self.path).always();
+        let mut exec = ctx.execute(&self.path).always();
```
GideonBear: Called `cmd` in some places, `exec` in others. Can we call it `cmd` everywhere?

src/steps/powershell.rs
```diff
@@ -115,19 +115,17 @@ impl Powershell {
...
-        command.args(["-NoProfile", "-Command"]);
-        command.arg(cmd);
-        command.args(args);
+        exec.args(["-NoProfile", "-Command", cmd]).args(args);
 
         // If topgrade was run from pwsh, but we are trying to run powershell, then
         // the inherited PSModulePath breaks module imports
         if !self.is_pwsh {
-            command.env_remove("PSModulePath");
+            exec.env_remove("PSModulePath");
```
GideonBear: You could add `env_remove_if` and `env_remove_if_some`...
Looks like there are only a few uses so your call.

src/steps/generic.rs
```diff
@@ -2253,12 +2230,13 @@ pub fn run_claude_code_plugins(ctx: &ExecutionContext) -> Result<()> {
...
     let mut success = true;
     for plugin in &plugins {
-        let mut cmd = ctx.execute(&claude);
-        cmd.args(["plugin", "update", &plugin.id, "--scope", &plugin.scope]);
+        let mut exec = ctx.execute(&claude);
+        exec.args(["plugin", "update", &plugin.id, "--scope", &plugin.scope]);
+
         if let Some(path) = &plugin.project_path {
-            cmd.current_dir(path);
+            exec.current_dir(path);
```
GideonBear: Same here, I wouldn't be opposed to `current_dir_if` and `current_dir_if_some`.

GideonBear:
It just occurred to me that we could go all the way and change
```rs
pub fn arg<S: AsRef<OsStr>>(&mut self, arg: S) -> &mut Executor {
```
to
```rs
pub fn arg<S: AsRef<OsStr>>(self, arg: S) -> Executor {
```
. This should be able to remove *all* instances of Executor assignments, given that we add all the additional methods (for `env_remove` and `current_dir`) as well.

However then we lose some flexibility by making it completely impossible to mutate them without reassigning. @NiekvD-09 @uwuclxdy thoughts?

[a.patch](https://github.com/user-attachments/files/30387296/a.patch)

GideonBear (changes requested):

src/steps/powershell.rs
NiekvD-09: i changed it to exec everywhere because it is four characters. Why cmd over exec?

src/steps/powershell.rs
NiekvD-09: as with `arg_if_else`, I now think that it is better for these cases to not get a specialized function

src/steps/powershell.rs
GideonBear: It doesn't matter that much, mostly convention. But personally I think "command" makes more sense to readers than "executor".

src/steps/generic.rs
NiekvD-09: also here for clearness and bloat

NiekvD-09:
> It just occurred to me that we could go all the way and change
> 
> ```rust
> pub fn arg<S: AsRef<OsStr>>(&mut self, arg: S) -> &mut Executor {
> ```
> 
> to
> 
> ```rust
> pub fn arg<S: AsRef<OsStr>>(self, arg: S) -> Executor {
> ```
> 
> . This should be able to remove _all_ instances of Executor assignments, given that we add all the additional methods (for `env_remove` and `current_dir`) as well.
> 
> However then we lose some flexibility by making it completely impossible to mutate them without reassigning. @NiekvD-09 @uwuclxdy thoughts?
> 
> [a.patch](https://github.com/user-attachments/files/30387296/a.patch)

At the moment most assignments left are that way because they either return a result with themselves or return themselves in a closure to deduplicate the base command and flags.
What is the status on #1612 ?

GideonBear:
> At the moment most assignments left are that way because they either return a result with themselves or return themselves in a closure to deduplicate the base command and flags.

I know, changing the signature from `&mut Self` to `Self` makes it so that we can directly return the result of e.g. `.arg()`.

> What is the status on https://github.com/topgrade-rs/topgrade/issues/1612 ?

Actually now that you mention it, my idea is even more convenient with #1612.
Before:
```rs
let base_cmd = ctx.execute(thing);
base_cmd.arg_if(ctx.config().yes(Step::Thing));
```
After:
```rs
let base_cmd = ctx.execute(thing)
    .arg_if(ctx.config().yes(Step::Thing));
```

Without changing the signature, #1612 would re-introduce a lot of the split `.arg` you removed here.

GideonBear:
All additional changes have been listed in https://github.com/topgrade-rs/topgrade/issues/1612#issuecomment-5104951737. I will merge this as-is (after minor fixes and squashing), because we want these basic `arg_if`s in any case.

GideonBear (approved):

GideonBear:
Made a few minor fixes/changes:
* `run_custom_command` changed to remove duplication
* `cmd`/`exec` naming (made it so this PR doesn't touch them)
* Some misplaced comments
* I think some more that I forgot

Thanks again!

PR #2228, opened 2026-07-25
feat(vscode): clean obsolete extension directories

### LINKED ISSUE #2028: VS Code-compatible editors: optional clean obsolete extension directories
## Checklist

VS Code-compatible editors record obsolete extension directories in `.obsolete` files under their extension roots:

```text
~/.cursor/extensions/.obsolete
~/.vscode/extensions/.obsolete
~/.vscode-insiders/extensions/.obsolete
```

Example:

```text
obsolete entry: anthropic.claude-code-2.1.142-darwin-arm64
active manifest: ~/.cursor/extensions/anthropic.claude-code-2.1.143-darwin-arm64
```

Would Topgrade accept an opt-in cleanup after successful VS Code-compatible extension updates?

Requested behavior:

1. For each VS Code-compatible editor variant Topgrade updates, find that variant's extension root.
2. Read that root's `.obsolete` file, if present.
3. Delete only existing direct child directories named by that file.
4. Leave `.obsolete` itself intact.
5. Skip or warn if the relevant editor is running.

Topgrade currently appears to have a shared `[vscode]` config for VS Code-compatible editors, not a Cursor-specific config section. Possible new config shape:

```toml
[vscode]
cleanup_obsolete_extensions = true
```

or per variant:

```toml
[cursor]
cleanup_obsolete_extensions = true

[vscode_insiders]
cleanup_obsolete_extensions = true
```

Alternatively, this could be gated behind Topgrade's existing global cleanup flag:

```toml
[misc]
cleanup = true
```

Safety: use `.obsolete` as the editor's source of truth rather than sorting versions, reject path traversal, and only delete direct children of the relevant extension root.

TyceHerrman:
## What does this PR do

Adds opt-in cleanup of obsolete extension directories after successful updates
for VS Code-compatible editors

When `[vscode].cleanup_obsolete_extensions` is enabled, topgrade:

- resolves the extension root for VS Code, VS Code Insiders, VSCodium,
  VSCodium Insiders, Cursor, Windsurf, and Antigravity IDE
- respects the `VSCODE_EXTENSIONS` environment override
- reads directory names from the editor-managed `.obsolete` JSON marker
- removes only existing direct-child directories while leaving `.obsolete`
  intact
- skips cleanup with a warning while the relevant editor is running

Cleanup isn't enabled by default. Dry runs report eligible directories
without deleting them.

The `.obsolete` parsing follows the approach used by
[Mole's `mo clean`](https://github.com/tw93/Mole/commit/e83730449a77ff92a937d1b1be536351eecda45a).
The editor's marker is the source of truth.

Closes #2028.

### Validation

- `cargo fmt --all -- --check`
- `cargo clippy --all-targets --locked -- -D warnings`
- `cargo clippy --all-targets --locked --all-features -- -D warnings`
- `cargo test --locked` — 34 tests passed

### Live integration

Tested with Cursor 3.13.10 on arm64 and an isolated
`VSCODE_EXTENSIONS`.

#### Dry run

```text
―― 08:04:34 - Cursor extensions ――
Dry running: /private/tmp/codex-topgrade-2028-live.B2gUPb/bin/cursor --update-extensions
Would remove obsolete extension directory: /private/tmp/codex-topgrade-2028-live.B2gUPb/extensions/publisher.old-1.0.0
―― 08:04:34 - Summary ――
Cursor extensions: OK
```

The obsolete directory, current directory, and `.obsolete` marker all remained.

#### Cleanup

```text
―― 08:05:17 - Cursor extensions ――
No extension to update
Removed obsolete extension directory: /private/tmp/codex-topgrade-2028-live.B2gUPb/extensions/publisher.old-1.0.0
―― 08:05:18 - Summary ――
Cursor extensions: OK
```

Topgrade removed only `publisher.old-1.0.0`, the extension directory listed in
`.obsolete`. It left `publisher.current-2.0.0` and the `.obsolete` marker file
intact.

#### Active editor guard

```text
―― 08:05:52 - Cursor extensions ――
No extension to update
 WARN Cursor is running; skipping obsolete extension cleanup
Cursor is running; skipping obsolete extension cleanup
―― 08:05:52 - Summary ――
Cursor extensions: OK
```

The obsolete directory remained while a temporary `cursor` process was active.

## Standards checklist

### AI involvement

An AI agent helped review the implementation and PR description. Final
review and edits were performed by me.


src/config.rs
```diff
@@ -2285,6 +2294,26 @@ mod test {
...
+        .expect("toml parse error");
+        let configured = Config {
+            opt: CommandLineArgs::parse_from::<_, String>([]),
+            config_file,
+            allowed_steps: Vec::new(),
+        };
+
+        assert!(configured.vscode_cleanup_obsolete_extensions());
+    }
+
```
```suggestion
```
Please don't add tests for the config file

src/steps/generic.rs
```diff
@@ -598,6 +599,146 @@ impl VSCodeVariant {
...
+            .any(|process| self.matches_process(process.name(), process.exe()))
+    }
+}
+
+fn is_safe_obsolete_extension_name(name: &str) -> bool {
+    if name.is_empty() || name.contains('/') || name.contains('\\') {
+        return false;
+    }
+
+    let mut components = Path::new(name).components();
```
GideonBear: Why not just check the amount of components?

src/steps/generic.rs
```diff
@@ -598,6 +599,146 @@ impl VSCodeVariant {
...
+    matches!(components.next(), Some(Component::Normal(_))) && components.next().is_none()
+}
+
+fn cleanup_obsolete_extensions(extension_root: &Path, dry_run: bool) -> Result<Vec<PathBuf>> {
+    let obsolete_path = extension_root.join(".obsolete");
+    let obsolete = match fs::read_to_string(&obsolete_path) {
+        Ok(obsolete) => obsolete,
+        Err(error) if error.kind() == std::io::ErrorKind::NotFound => return Ok(Vec::new()),
+        Err(error) => {
+            return Err(error).wrap_err_with(|| format!("Failed to read {}", obsolete_path.display()));
```
GideonBear: Always `Err` so doesn't need `_with`

src/steps/generic.rs
```diff
@@ -598,6 +599,146 @@ impl VSCodeVariant {
...
+        }
+    };
+    let obsolete: BTreeMap<String, serde_json::Value> =
+        serde_json::from_str(&obsolete).wrap_err_with(|| format!("Failed to parse {}", obsolete_path.display()))?;
+
+    let mut cleaned = Vec::new();
+    for extension in obsolete.keys() {
```
GideonBear: It looks like the value is a `bool`. Don't we want to check if it's `true` instead of discarding it?

src/steps/generic.rs
```diff
@@ -598,6 +599,146 @@ impl VSCodeVariant {
...
+            );
+            continue;
+        }
+
+        let extension_path = extension_root.join(extension);
+        let metadata = match fs::symlink_metadata(&extension_path) {
+            Ok(metadata) => metadata,
+            Err(error) if error.kind() == std::io::ErrorKind::NotFound => continue,
+            Err(error) => {
+                return Err(error).wrap_err_with(|| format!("Failed to inspect {}", extension_path.display()));
```
GideonBear: Same here, doesn't need `_with`

src/steps/generic.rs
```diff
@@ -598,6 +599,146 @@ impl VSCodeVariant {
...
+            );
+            continue;
+        }
+
+        if !dry_run {
+            fs::remove_dir_all(&extension_path)
+                .wrap_err_with(|| format!("Failed to remove {}", extension_path.display()))?;
+        }
+        cleaned.push(extension_path);
+    }
```
GideonBear: Can you simply move both prints here and remove the return? I don't think the return value is useful.

src/steps/generic.rs
```diff
@@ -679,7 +820,49 @@ fn run_vscode_compatible(variant: VSCodeVariant, ctx: &ExecutionContext) -> Resu
...
+        }
+        Err(error) => {
+            warn!(?error, editor = name, "Failed to clean obsolete extensions");
+            print_warning(t!(
+                "Failed to clean obsolete {editor} extensions: {error}",
+                editor = name,
+                error = format!("{error:#}")
+            ));
+        }
+    }
```
GideonBear: Just `.wrap_err` and propagate the error (removing the return value makes it so a simple `cleanup_obsolete_extensions(...).wrap_err(...)?` is enough)

src/steps/generic.rs
```diff
@@ -714,6 +897,176 @@ pub fn run_antigravity_extensions_update(ctx: &ExecutionContext) -> Result<()> {
...
+            (VSCodeVariant::Codium, ".vscode-oss/extensions"),
+            (VSCodeVariant::CodiumInsiders, ".vscodium-insiders/extensions"),
+            (VSCodeVariant::Cursor, ".cursor/extensions"),
+            (VSCodeVariant::Windsurf, ".windsurf/extensions"),
+        ];
+
+        for (variant, expected) in variants {
+            assert_eq!(variant.extensions_root(home, None), home.join(expected));
+        }
+    }
```
GideonBear: Basically reiterates the function itself, please remove

GideonBear (changes requested):

src/steps/generic.rs
TyceHerrman: A component-only count check would also accept `.` and `..`: Rust reports both as one component (`CurDir` and `ParentDir`, respectively). The current check intentionally requires exactly one `Component::Normal`. I could add a comment there explaining why this check requires exactly one normal path component, or do something else if you still don't like this approach

GideonBear: I get why you want to check the component type, but can't you replace `name.is_empty() || name.contains('/') || name.contains('\\')` with `components.size() == 1` (after collecting into a vec)?

GideonBear (approved):
Thanks!

PR #2229, opened 2026-07-25
Draft: proof of concept for overide step path

### LINKED ISSUE #2200: Option to override step path
Example:
```toml
[hx]
path = "not_hx"
# Or:
path = "/usr/bin/hx"
```
We simply have a default value of `hx` and let the user override it. `which` should be able to handle absolute paths just fine.

This could help when a user has conflicting binaries installed, and
1. has renamed one of the binaries
2. uses the full path / has aliased the full path

(thx @uwuclxdy)

NiekvD-09:
Closes #2200
@GideonBear I came up with this implementation, initially i didnt want to change all step functions and just edit `require()`. That could handle alternative paths, but doesn't solve the case where multiple steps have the same command name.
My proposal is:
```
pub fn run_rustup(ctx: &ExecutionContext) -> Result<()> {
    let rustup = require(ctx.config().path_name(Binaries::Rustup))?;

    print_separator("rustup");

    ctx.execute(rustup)
        .arg("update")
        .args(ctx.config().rustup_channels())
        .status_checked()
}
```

`path_name` initially returns the default name for a binary that gets passed onto which.
If there are multiple commands with the same name, the default can be set to the relative path of the command.
If the user decides to change it, that can be done in the config with
```
[misc]
alt_paths = [{name="rustc",path="/usr/bin/bear"},{name="spicetify",path="/mnt/data/spice"}]
```

uhh also i apparently messed up with git again, only look at the last commit 

feedback please

## Standards checklist

### AI involvement

No AI used

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -21,7 +21,7 @@ use strum::IntoEnumIterator;
 use tracing::{debug, error};
 
 use crate::execution_context::RunType;
-use crate::step::{DEPRECATED_STEPS, Step};
+use crate::step::{Binaries, DEPRECATED_STEPS, Step};
 use crate::sudo::SudoKind;
 use crate::terminal::print_warning;
 use crate::utils::string_prepend_str;
@@ -61,6 +61,12 @@ pub struct Containers {
     runtime: Option<ContainerRuntime>,
     system_prune: Option<bool>,
     use_sudo: Option<bool>,
+}
+
+#[derive(Deserialize, Default, Debug, Merge)]
+pub struct AltPath {
+    name: Option<String>,
+    path: Option<String>,
 }
 
 #[derive(Deserialize, Default, Debug, Merge)]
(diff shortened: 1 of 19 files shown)


NiekvD-09:
also, this:
```
[hx]
path = "not_hx"
# Or:
path = "/usr/bin/hx"
```
wouldn't work*, because some steps invoke multiple commands

*_as mentioned before i am not that smart so please don't hesitate to tell me how incredibly wrong i am about this_

GideonBear:
Thanks for the PR, but the issue was not labeled with "needs pr" for a reason (namely that it, or its easy implementation, is dependent on some other v18 stuff).
Next time, please ask before PRing for an issue that is not labeled "needs pr" (or has an open PR, or is assigned to someone else), I don't want you to waste your time :)

GideonBear:
We want per-step configuration instead of central, which is possible with the v18 config.
> because some steps invoke multiple commands

This is currently true, but won't be in v18.

Your implementation and idea are good, but unfortunately I'll have to close this.

PR #2234, opened 2026-07-27
refactor: add a generic `require_one_path` function

GideonBear:


src/steps/os/macos.rs
uwuclxdy: remove unused PathBuf import

src/utils.rs
```diff
@@ -265,6 +265,33 @@ pub fn require_one<T: AsRef<OsStr> + Debug>(binary_names: impl IntoIterator<Item
...
+            return Ok(path.to_path_buf());
+        } else {
+            failed_paths.push(path_s);
+        }
+    }
+
+    Err(SkipStep(format!(
+        "{}",
+        t!(
+            "None of {paths} exist",
```
uwuclxdy, line 284: there's no translation for this

uwuclxdy (approved):
oh i like this thanks! but i couldn't find any other places where this could get used as it is

Issue #2165, opened 2026-07-12
Walk through, minimize, and comment, all panics


GideonBear:
* [x] `.unwrap()`
* [x] `.expect(`
* [x] `panic!`
* [x] `unreachable!` (currently used for bad output from steps)
* [x] `assert*` (outside of tests)
* [x] [...?]

We should use `return Err(eyre!(output_changed_message!(...)))` or `return Err(eyre!("..."))` instead.

GideonBear:
List by @uwuclxdy:

missing comments:
- [x] terminal.rs:306 lock
- [x] terminal.rs:310 lock
- [x] terminal.rs:315 lock
- [x] terminal.rs:320 lock
- [x] terminal.rs:325 lock
- [x] terminal.rs:329 lock
- [x] terminal.rs:334 lock
- [x] terminal.rs:338 lock
- [x] terminal.rs:342 lock
- [x] terminal.rs:348 lock
- [x] terminal.rs:354 lock
- [x] terminal.rs:358 lock
- [x] terminal.rs:362 lock
- [x] execution_context.rs:120 lock
- [x] execution_context.rs:124 lock
- [x] main.rs:51
- [x] main.rs:53
- [x] main.rs:56
- [x] config.rs:749 regex
- [x] generic.rs:1684 regex
- [x] generic.rs:1704 regex
- [x] vagrant.rs:223 regex
- [x] generic.rs:1714 some stuff that I don't understand
- [x] generic.rs:1718 some stuff that I don't understand
- [x] breaking_changes.rs:29
- [x] sudo.rs:226 somewhat self-explaining
- [x] main.rs:347 you added this :P I initially thought this was a panic in the thread so hard to propagate the error, but it's not (fixed)
- [x] generic.rs:1729 some stuff that I don't understand

should be Err:
- [x] utils.rs:222
- [x] tmux.rs:83 valid use of unreachable
- [x] config.rs:806
- [x] Valid uses of unreachable
  - [x] sudo.rs:339
  - [x] sudo.rs:438
  - [x] sudo.rs:478
  - [x] sudo.rs:500
  - [x] sudo.rs:519
  - [x] sudo.rs:542
- [x] os/unix.rs:1035 valid use of unreachable
- [x] vagrant.rs:79
- [x] vagrant.rs:80
- [x] vagrant.rs:115 valid use of unreachable, although unidiomatic
- [x] vagrant.rs:138 valid use of unreachable, although unidiomatic
[...]

PR #2239, opened 2026-07-29
refactor(sudo): split null sudo off into separate enum variant to remove unreachable match arms

GideonBear:
While doing #2165 I found these `unreachable!()` and `.unwrap()` calls that are technically valid but could be removed by not making `Null` an actual `SudoKind`.
Ping @atoerien, maybe you already considered this. I don't know if this is a good idea, but I wanted to at least try it.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -144,7 +144,7 @@ fn run() -> Result<()> {
 
     let sudo = match config.sudo_command() {
         Some(kind) => Sudo::new(kind),
-        None if elevated => Ok(Sudo::new_already_root()),
+        None if elevated => Ok(Sudo::AlreadyRoot),
         None => Sudo::detect(),
     };
     debug!("Sudo: {:?}", sudo);
--- a/src/main.rs
+++ b/src/main.rs
@@ -24,7 +24,7 @@ use self::config::{CommandLineArgs, Config};
 use self::error::StepFailed;
 use self::runner::StepResult;
 use self::steps::{remote::*, *};
-use self::sudo::{Sudo, SudoCreateError, SudoKind};
+use self::sudo::{Sudo, SudoCreateError};
 use self::terminal::*;
 use self::utils::{install_color_eyre, install_tracing, is_elevated, set_wsl_use_windows_path, update_tracing};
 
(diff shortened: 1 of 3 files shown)


GideonBear:
Renamed to `AlreadyRoot` which we don't have to keep (can always keep calling it null sudo internally and use `already root` in the error msg)

src/main.rs
```diff
@@ -144,7 +144,7 @@ fn run() -> Result<()> {
 
     let sudo = match config.sudo_command() {
         Some(kind) => Sudo::new(kind),
-        None if elevated => Sudo::new(SudoKind::Null),
+        None if elevated => Ok(Sudo::new_already_root()),
```
uwuclxdy:
```suggestion
        None if elevated => Ok(Sudo::AlreadyRoot),
```

src/sudo.rs
```diff
@@ -257,13 +258,8 @@ impl Sudo {
...
-    /// Gets the path to the `sudo` binary. Do not use this to execute `sudo` directly - either use
-    /// [`Sudo::elevate`], or if you need to specify arguments to `sudo`, use [`Sudo::elevate_opts`].
-    /// This way, sudo options can be specified generically and the actual arguments customized
-    /// depending on the sudo kind.
-    #[allow(unused)]
-    pub fn path(&self) -> Option<&Path> {
-        self.path.as_deref()
+    pub fn new_already_root() -> Self {
+        Self::AlreadyRoot
     }
```
uwuclxdy, line 263:
```suggestion
```

this has no callers anymore

src/sudo.rs
```diff
@@ -27,9 +26,16 @@ use crate::terminal::print_separator;
...
 #[derive(Clone, Debug)]
-pub struct Sudo {
+pub enum Sudo {
+    Real(InternalSudo),
+    /// A "no-op" sudo, used when topgrade itself is running as root
+    AlreadyRoot,
+}
+
+#[derive(Clone, Debug)]
+struct InternalSudo {
```
uwuclxdy, line 36:
```suggestion
pub struct InternalSudo {
```

`error: type sudo::InternalSudo is more private than the item sudo::Sudo::Real::0`

src/sudo.rs
```diff
@@ -391,59 +312,153 @@ impl Sudo {
...
-                // TODO: emulate running in a login shell with su/runuser
-                return Err(UnsupportedSudo {
-                    sudo_kind: self.kind,
-                    option: "login_shell",
+        match self {
+            // already root sudo is very different, do separately
+            Self::AlreadyRoot => {
+                if opts.login_shell {
+                    // TODO: emulate running in a login shell with su/runuser
+                    return Err(UnsupportedSudo::new_already_root("login_shell").into());
```
uwuclxdy, line 320: "already root does not support the login_shell option" reads a bit weird

PR #2241, opened 2026-07-29
docs: remove unnecessary dry-run question from PR template

GideonBear:
@uwuclxdy I think this is all that needs to change in the PR template, right?


uwuclxdy:
yea i think so

uwuclxdy (approved):

PR #2304, opened 2026-09-07
fix(claude): skip plugins without a configured marketplace

jpweytjens:
## What does this PR do
The current "Claude Code Plugins" step fails on any plugin loaded from `~/.claude/skills/` or `<cwd>/.claude/skills/`, which `claude plugin list --json` reports with the id suffix `@skills-dir`. This PR skips those plugins, since [Claude Code itself](https://code.claude.com/docs/en/plugins-reference#skills-directory-plugins) says they cannot be updated and edits there take effect after a reload.

## Standards checklist

- [X] *Optional:* The PR title is a descriptive commit message (this will appear in release notes, leave unchecked if you want a maintainer to improve it)

```
Before, with topgrade 17.9.0:

Checking for updates for plugin "tinymist-lsp@skills-dir" at user scope…
✘ Failed to update plugin "tinymist-lsp@skills-dir": This plugin is loaded from ~/.claude/skills/ with no marketplace backing — it cannot be updated. Delete the directory to remove it; `claude plugin disable` to turn it off; edits there take effect after /reload-plugins.
ERROR Updating plugin tinymist-lsp@skills-dir failed: Command failed: `/Users/johannes/.local/bin/claude plugin update tinymist-lsp@skills-dir --scope user`
―― 13:32:20 - Summary ――
Claude Code Plugins: FAILED

After, with this branch and -v:

DEBUG Skipping plugin tinymist-lsp@skills-dir: loaded from ~/.claude/skills, no marketplace
―― 14:24:52 - Summary ――
Claude Code Plugins: OK
```

### AI involvement
AI found the cause and drafted the patch. I reviewed and tested it, and wrote this description.

### DIFF
--- a/src/steps/generic.rs
+++ b/src/steps/generic.rs
@@ -2549,6 +2549,16 @@ pub fn run_claude_code_plugins(ctx: &ExecutionContext) -> Result<()> {
 
     let mut success = true;
     for plugin in &plugins {
+        // A plugin loaded straight from `~/.claude/skills/` has no marketplace behind it,
+        // so `claude plugin update` refuses it and nothing can be updated anyway.
+        if plugin.id.ends_with("@skills-dir") {
+            debug!(
+                "Skipping plugin {}: loaded from ~/.claude/skills, no marketplace",
+                plugin.id
+            );
+            continue;
+        }
+
         let mut cmd = ctx.execute(&claude);
         cmd.args(["plugin", "update", &plugin.id, "--scope", &plugin.scope]);
 
--- a/src/steps/generic.rs
+++ b/src/steps/generic.rs
@@ -8,7 +8,7 @@ use regex::bytes::Regex;
 use rust_i18n::t;
 use semver::Version;
 use serde::Deserialize;
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, HashSet};
 use std::env;
 use std::ffi::{OsStr, OsString};
 #[cfg(unix)]
@@ -2528,6 +2528,11 @@ pub fn run_claude_code_plugins(ctx: &ExecutionContext) -> Result<()> {
         project_path: Option<PathBuf>,
     }
 
+    #[derive(Deserialize)]
+    struct ClaudeMarketplace {
+        name: String,
+    }
+
     let claude = require("claude")?;
 
     print_separator("Claude Code Plugins");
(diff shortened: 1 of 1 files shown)


uwuclxdy (commented):
Thanks for the PR, this is a good thing to implement but probably not in this way. It only covers cases where the "marketplace" is set to `skills-dir` (which is not a valid remote to pull from and fails, so it should get skipped correctly). Could you find a way to detect if any of the plugins have no marketplace backing and skip them as well? If `skills-dir` is the only such case where a plugin has no marketplace backing, then this is good to merge imo.

jpweytjens:
There's no exhaustive list of suffixes that aren't marketplace. The best I could find is this [sdk.d.js](https://cdn.jsdelivr.net/npm/@anthropic-ai/claude-agent-sdk@0.3.263/sdk.d.ts). `skills-dir` seems to be only one without marketplace. It seems that the shape of the skills marketplaces has already been changed once by Anthropic, so not sure how reliable this is.

If you'd prefer something more general, we could do something like

```
struct ClaudeMarketplace {
    name: String,
}

let output = ctx
    .execute(&claude)
    .args(["plugin", "marketplace", "list", "--json"])
    .output_checked_utf8()?;
let marketplaces: Vec<ClaudeMarketplace> = serde_json::from_str(&output.stdout)
    .wrap_err_with(|| output_changed_message!("claude plugin marketplace list --json", "..."))?;
let known: HashSet<&str> = marketplaces.iter().map(|m| m.name.as_str()).collect();

for plugin in &plugins {
    let marketplace = plugin.id.rsplit_once('@').map(|(_, m)| m);
    if !marketplace.is_some_and(|m| known.contains(m)) {
        debug!("Skipping plugin {}: no configured marketplace behind it", plugin.id);
        continue;
    }
    ...
}
```

uwuclxdy:
thanks for checking, the more general solution you proposed is what i had in mind yes. 
could you update the pr? then its good to merge

jpweytjens:
Updated the PR with the general approach: the step now reads `claude plugin marketplace list --json` and skips any plugin whose `@<marketplace>` suffix is not a configured marketplace.

Tested locally with `topgrade --only claude_code_plugins -v`:

```
DEBUG Skipping plugin tinymist-lsp@skills-dir: no configured marketplace behind it
―― 18:56:15 - Summary ――
Claude Code Plugins: OK
```

All marketplace-backed plugins still updated as before.

src/steps/generic.rs
```diff
@@ -2536,6 +2541,18 @@ pub fn run_claude_code_plugins(ctx: &ExecutionContext) -> Result<()> {
...
+        .execute(&claude)
+        .args(["plugin", "marketplace", "list", "--json"])
+        .output_checked_utf8()?;
+    let marketplaces: Vec<ClaudeMarketplace> = serde_json::from_str(&output.stdout).wrap_err_with(|| {
+        output_changed_message!(
+            "claude plugin marketplace list --json",
+            "json output is invalid or does not match expected structure"
+        )
+    })?;
+    let marketplaces: HashSet<&str> = marketplaces.iter().map(|m| m.name.as_str()).collect();
```
```suggestion
    let marketplaces: HashSet<String> = marketplaces.into_iter().map(|m| m.name).collect();
```

uwuclxdy (approved):
Looks good, thanks! @GideonBear this can be merged as far as I'm concerned

GideonBear (approved):
Thanks both!

PR #2333, opened 2026-09-13
NOSQUASH: refactor: change return type of `which` to `Result<Option<PathBuf>>`

GideonBear:
Does some weird stuff...


src/sudo.rs
```diff
@@ -171,7 +171,9 @@ impl Sudo {
             return Ok(Self { path: None, kind });
         }
 
-        match kind.which() {
+        // TODO: This throws away the error as `which` did previously.
+        //  If possible, propagate this error.
+        match kind.which().ok().flatten() {
```
uwuclxdy:
```suggestion
        match kind.which().inspect_err(|e| tracing::error!("Detecting sudo failed: {e}")).ok().flatten() {
```

src/steps/os/linux.rs
```diff
@@ -783,19 +783,20 @@ fn upgrade_kde_linux(ctx: &ExecutionContext) -> Result<()> {
...
     };
-    ctx.execute(&dnf)
+    Ok(ctx
+        .execute(&dnf)
         .always()
         .arg("--version")
         .output_checked_utf8()
         .map(|output| !output.stdout.contains("dnf5"))
-        .unwrap_or(false)
+        .unwrap_or(false))
```
uwuclxdy: this drops `dnf --version` errors and the needrestart step silently proceeds. remove `.unwrap_or(false)` to turn it into a step error.

src/execution_context.rs
```diff
@@ -129,14 +129,21 @@ impl<'a> ExecutionContext<'a> {
         self.distribution
     }
 
-    pub fn powershell(&self) -> &Result<Powershell, SkipStep> {
+    pub fn powershell(&self) -> &Result<Powershell> {
```
uwuclxdy: the widened `Err` from `Powershell::new` reaches `src/steps/git.rs:61` as `if let Ok(...)`, so genuine detection errors become a silent skip. routing that caller through `require_powershell()` with a `SkipStep` downcast keeps the skip case and surfaces the rest.

src/steps/powershell.rs
```diff
@@ -23,9 +23,9 @@ pub struct Powershell {
 impl Powershell {
     /// Detects a usable PowerShell. The `Err` carries the reason so the step can
     /// surface it as its skip message instead of a generic "not installed"
```
uwuclxdy: `new` can now also carry a `which` detection error, and those surface as step failures, not skip messages.

src/utils.rs
```diff
@@ -149,51 +160,44 @@ fn which_native_in_wsl<T: AsRef<OsStr> + Debug>(binary_name: T) -> Option<PathBu
...
         None => {
             debug!("Cannot find {:?}", &binary_name);
-            None
+            Ok(None)
         }
     }
 }
 
-pub fn which<T: AsRef<OsStr> + Debug>(binary_name: T) -> Option<PathBuf> {
+pub fn which<T: AsRef<OsStr> + Debug>(binary_name: T) -> Result<Option<PathBuf>> {
```
uwuclxdy, line 180:
```suggestion
/// `Ok(None)` = not found, `Err` = lookup failed
pub fn which<T: AsRef<OsStr> + Debug>(binary_name: T) -> Result<Option<PathBuf>> {
```

src/utils.rs
```diff
@@ -240,21 +244,21 @@ pub fn require_flatpak(ctx: &ExecutionContext, name: &str) -> Result<Executor> {
 }
 
 #[allow(unused)]
-pub fn which_one<T: AsRef<OsStr> + Debug>(binary_names: impl IntoIterator<Item = T>) -> Option<PathBuf> {
+pub fn which_one<T: AsRef<OsStr> + Debug>(binary_names: impl IntoIterator<Item = T>) -> Result<Option<PathBuf>> {
     for bin in binary_names {
-        if let Some(path) = which(&bin) {
-            return Some(path);
+        if let Some(path) = which(&bin)? {
```
uwuclxdy, line 249:
```suggestion
        if let Some(path) = which(&bin).wrap_err_with(|| format!("Detecting {bin:?} failed"))? {
```

a `which` failure propagated through `which_one` reports as just "cannot canonicalize path", since the `which_crate` error doesn't name the binary. same for `require_one` below.

src/sudo.rs
GideonBear: oh that's what you meant, good idea

src/execution_context.rs
GideonBear: `require_powershell` doesn't help, that also just returns a `eyre::Result`.
I made a new method.

src/utils.rs
GideonBear: I've just copied the wrapping that `require` already did to `which`. Better to put in `which` than in `which_one`, right?

uwuclxdy (approved):

src/utils.rs
```diff
@@ -127,17 +141,14 @@ fn is_windows_mount_path(path: &Path, prefixes: &[PathBuf]) -> bool {
...
         Err(which_crate::Error::CannotFindBinaryPath) => {
             debug!("Cannot find {:?}", &binary_name);
-            return None;
-        }
-        Err(e) => {
-            error!("Detecting {:?} failed: {}", &binary_name, e);
-            return None;
+            return Ok(None);
         }
+        Err(e) => return Err(e.into()),
```
GideonBear: same msg as below

PR #2341, opened 2026-09-14
refactor: replace unnecessary `OnceLock` with `LazyLock`

GideonBear:
@uwuclxdy your claude should've used a `LazyLock`


uwuclxdy (approved):
sorry @GideonBear, i had my claude in pace mode

Image

